PR #7, opened 2018-08-28
Mocha tests + DurableOrchestrationContext, DurableOrchestrationClient API parity.

kashimiz:
* Moved /test/sample to /samples
* Configured TypeScript linting and VS Code debugging
* Added unit tests for:
  - TimerTask
  - RetryOptions
  - OrchestrationClient
* Added integration tests for:
  - Orchestrator (all methods)
* Added functionality:
  - error handling/propagation (#18)
  - instanceId (#8), isReplaying (#9), parentInstanceId (#14)
  - callActivityWithRetry(name, retryOptions, input) (#15)
  - callSubOrchestrator(name, input, instanceId) (#10)
  - callSubOrchestratorWithRetry(name, retryOptions, input, instanceId) (#16)
  - continueAsNew(input) (#11)
  - setCustomStatus(status) (#17)
  - DurableOrchestrationClient API (all of it)
* Fixed bugs:
  - Azure/azure-functions-durable-extension#516
  - #23


README.md
```diff
@@ -89,23 +89,24 @@ The shim strives to hew closely to the C# DurableOrchestrationContext API, while
...
 * `CreateTimer(Date fireAt)`
 * `GetInput()`
 * `WaitForExternalEvent(String name)`
 * `OrchestrationClient` binding to `string`
 
 **Not Yet Implemented**
-* `InstanceId`
-* `IsReplaying`
 * `CallActivityWithRetryAsync(String, RetryOptions, Object)`
 * `CallSubOrchestratorAsync(String, Object)`
```
cgillum: It might be too late for this feedback, but why are we suffixing these API calls with `Async`? Is there any precedent for that in JavaScript? I only know it as a .NET convention for APIs that return a Task.

src/orchestrator.ts
```diff
@@ -18,7 +20,10 @@ export class Orchestrator {
 
         // Assign methods to context
         context.df = {};
+        context.df.instanceId = context.bindings.context.instanceId;
+        context.df.isReplaying = context.bindings.context.isReplaying;
         context.df.callActivityAsync = this.callActivityAsync.bind(this, state);
+        context.df.callSubOrchestratorAsync = this.callSubOrchestratorAsync.bind(this, state);
```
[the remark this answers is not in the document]
kashimiz: Yes; that's getting updated as well.

src/historyevent.ts
```diff
@@ -2,14 +2,19 @@ import { HistoryEventType } from "./classes";
 
 export class HistoryEvent {
     constructor(
```
christopheranderson: Same comment here on constructors having lots of arguments - makes reading code really hard if there is more than a few arguments.

src/orchestrator.ts
```diff
@@ -1,12 +1,16 @@
 import * as debug from "debug";
-import { ActionType, CallActivityAction, CreateTimerAction, HistoryEvent, HistoryEventType,
-    IAction, Task, TaskSet, TimerTask, WaitForExternalEventAction } from "./classes";
+import { ActionType, CallActivityAction, CallActivityWithRetryAction, CallSubOrchestratorAction,
```
hartra344: nit: ActionType seems to be unused, dont import it

src/historyevent.ts
```diff
@@ -2,14 +2,19 @@ import { HistoryEventType } from "./classes";
 
 export class HistoryEvent {
```
hartra344: I see a pattern of making wrapper classes with an any object inside of it. Might be better to make these wrappers Generics so HistoryEvent<T> with input having type T.

src/timertask.ts
```diff
@@ -3,12 +3,17 @@ import { CreateTimerAction, Task } from "./classes";
...
     constructor(
         public isCompleted: boolean,
-        public isCanceled: boolean,
+        public isFaulted: boolean,
         public action: CreateTimerAction,
         public result?: any,
         public timestamp?: Date,
         public id?: number,
-    ) { super(isCompleted, action, result, timestamp, id); }
+    ) { super(isCompleted, isFaulted, action, result, timestamp, id); }
```
hartra344: nit: move super to a new line

hartra344:
On first linter pass(not necessarily understanding the functionality yet). I agree with @christopheranderson on pretty much everything he said. Only major thing I'll add is on the large use of 'any'. You're losing out on a lot of typescript functionality that way. I would suggest building on generics so who ever is using the class/function can tell you what type input is, this will allow some better type safety for you but a lot more for whoever is consuming this

src/orchestrator.ts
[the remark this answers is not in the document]
kashimiz: There isn't a precedent, of which I was not aware. Removing these suffixes.

src/callsuborchestratoraction.ts
```diff
@@ -0,0 +1,11 @@
+import { ActionType, IAction } from "./classes";
+
+export class CallSubOrchestratorAction implements IAction {
+    public actionType: ActionType = ActionType.CallSubOrchestrator;
+
+    constructor(
+        public functionName: string,
```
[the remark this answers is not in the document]
kashimiz: Good point. Once these properties are set, they should remain set.

package.json
```diff
@@ -39,18 +37,20 @@
     "winston": "~2.3.1"
   },
   "devDependencies": {
-    "@types/chai": "3.5.0",
+    "@types/chai": "*",
```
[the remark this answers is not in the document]
kashimiz: Is this a general rule of thumb using the chai dependency as an example, or advice specific to chai?

src/retryoptions.ts
```diff
@@ -0,0 +1,14 @@
+export class RetryOptions {
+    public backoffCoefficient: number;
+    public maxRetryIntervalInMilliseconds: number;
+    public retryTimeoutInMilliseconds: number;
+
+    constructor(
+        public firstRetryIntervalInMilliseconds: number,
```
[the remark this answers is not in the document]
kashimiz: Yes; this is the JS counterpart to [RetryOptions](https://github.com/Azure/azure-functions-durable-extension/blob/c6d3ce6dd427f2c1b9659d0ef6776922da062f44/src/WebJobs.Extensions.DurableTask/RetryOptions.cs#L35) in Azure/azure-functions-durable-extension, whose properties are all public and mutable.

src/orchestrator.ts
```diff
@@ -53,48 +65,204 @@ export class Orchestrator {
...
             } catch (error) {
-                this.error(context, error);
+                this.error(context, actions, error);
                 return;
             }
         }
     }
 
-    private callActivityAsync(state: HistoryEvent[], name: string, input: any = "__activity__default") {
+    private callActivity(state: HistoryEvent[], name: string, input?: any) {
```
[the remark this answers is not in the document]
kashimiz: I think this is a case for any. At this point the input hasn't been serialized yet, so it could legally be a(n):

- boolean
- string
- number
- object
- array
- tuple
- enum
- undefined (I don't know why you would, but you could)
- null (we allow passing null to CallActivityAsync in the C# API)

src/historyevent.ts
kashimiz: This is true, and it bothered me while I was writing it (but I shrugged it off.) I should learn to be less tolerant of coding annoyances. 🤔

src/timertask.ts
```diff
@@ -3,12 +3,17 @@ import { CreateTimerAction, Task } from "./classes";
...
+
+    get isCanceled(): boolean {
```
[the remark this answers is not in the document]
kashimiz: Good call.

src/orchestrator.ts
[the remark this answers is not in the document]
christopheranderson: If you're targeting 3.0 TypeScript, you can use `Unknown` for these types, which forces type sniffing down stream.

src/task.ts
```diff
@@ -3,9 +3,11 @@ import { IAction } from "./classes";
 export class Task {
     constructor(
         public isCompleted: boolean,
```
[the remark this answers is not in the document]
kashimiz: Very true. I've refactored to reflect this.

package.json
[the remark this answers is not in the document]
kashimiz: Dependencies pinned to patch versions (~). Moved dependencies that were samples-specific out of the main package.json.

src/actions/callactivitywithretryaction.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export class CallActivityWithRetryAction implements IAction {
+    public readonly actionType: ActionType = ActionType.CallActivityWithRetry;
+
+    constructor(
+        public readonly functionName: string,
+        public readonly retryOptions: RetryOptions,
+        public readonly input?: unknown,
+    ) {
+        if (!retryOptions) {
```
christopheranderson: why do you only validate retryOptions?

src/actions/callsuborchestratorwithretryaction.ts
```diff
@@ -0,0 +1,16 @@
...
+    public readonly actionType: ActionType = ActionType.CallSubOrchestratorWithRetry;
+
+    constructor(
+        public readonly functionName: string,
+        public readonly retryOptions: RetryOptions,
+        public readonly input: unknown,
+        public readonly instanceId?: string,
+    ) {
+        if (!retryOptions) {
+            throw new TypeError(`retryOptions: expected type RetryOptions but got ${typeof retryOptions}`);
```
christopheranderson: ditto validation question here

src/actions/callsuborchestratoraction.ts
```diff
@@ -0,0 +1,11 @@
+import { ActionType, IAction } from "../classes";
+
+export class CallSubOrchestratorAction implements IAction {
+    public readonly actionType: ActionType = ActionType.CallSubOrchestrator;
+
+    constructor(
+        public readonly functionName: string,
+        public readonly instanceId: string,
+        public readonly input?: unknown,
+    ) { }
```
christopheranderson: Is this a public type to customers? If so, you should make sure you have a unit test to make sure you don't break the public surface area.

christopheranderson: Apply this same question to all the these dto type classes

src/history/historyeventoptions.ts
```diff
@@ -0,0 +1,17 @@
+export class HistoryEventOptions {
+    public details?: string;
```
christopheranderson: are all of these properties really optional?

src/orchestrationclient.ts
```diff
@@ -326,25 +323,27 @@ export class OrchestrationClient {
     }
 
     private getClientData(): OrchestrationClientInputData {
-        const dataKey = this.context.bindings ? Object.keys(this.context.bindings).filter((val: string) => {
-            return Object.keys(new OrchestrationClientInputData(undefined, undefined, undefined)).every((key) => {
-                return this.context.bindings[val].hasOwnProperty(key);
-            });
-        })[0] : undefined;
+        const dataKey = (this.context as IFunctionContext).bindings
```
christopheranderson: this is a gnarly bit of code. I'd recommend putting in a method and writing some unit tests

src/orchestrationclient.ts
```diff
@@ -355,7 +354,11 @@ export class OrchestrationClient {
         return payload;
     }
 
-    private hasValidRequestUrl(request: any): boolean {
-        return request && (request.http && request.http.url || request.url);
+    private hasValidRequestUrl(request: IRequest): boolean {
+        if (request && request.url) {
```
christopheranderson: why not simplify this to `return request && request.url`?

kashimiz: TypeScript complains that the return type is `string` instead of `boolean` in that format. It accepts the slightly more verbose `request !== undefined && request.url !== undefined` at least.

src/actions/callactivitywithretryaction.ts
kashimiz: Good point; we could validate the functionName too as it's passed in from the user code.

src/actions/callsuborchestratorwithretryaction.ts
kashimiz: Ditto response above.

src/actions/callsuborchestratoraction.ts
kashimiz: This is not a public type to customers; they should never interact with it directly.

src/history/historyeventoptions.ts
kashimiz: On different derivatives of HistoryEvent, yes. It's a.. bit of a property bag. I could stand to refactor this out into some interfaces.

src/orchestrationclient.ts
kashimiz: Done.

README.md
```diff
@@ -1,39 +1,68 @@
 # Durable Functions for Node.js
 
-This library provides a shim to write your [Durable Functions](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-overview) orchestrator functions in [Node.js](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node), using Durable's out-of-proc execution protocol. **JS orchestrators have the same [code constraints](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-checkpointing-and-replay#orchestrator-code-constraints) as C# orchestrators.**
+The `durable-functions` npm package allows you to write [Durable Functions](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-overview) in [Node.js](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node). Durable Functions is an extension of [Azure Functions](https://docs.microsoft.com/en-us/azure/azure-functions/functions-overview) that lets you write stateful functions and workflows in a serverless environment. The extension manages state, checkpoints, and restarts for you. Durable Functions' advantages include:
```
mhoeger: "in Node.js" feels a bit strange to me. "for Node.js"? Or "in JavaScript"

README.md
```diff
@@ -1,39 +1,68 @@
...
+You can follow the [Visual Studio Code quickstart](https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode) to get started with a function chaining example, or follow the general checklist below:
 
-Run this command from the root folder of your functions app:
-```
+1. Install prerequisites:
+    - [Azure Functions Core Tools version 2.x](https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local#install-the-azure-functions-core-tools)
+    - [Azure Storage Emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator) (Windows) or an actual Azure storage account (Mac or Linux)
+    - Node.js 8.6.0 or later
+
+2. [Create an Azure Functions app.](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-function-vs-code) Visual Studio Code is recommended.
```
mhoeger: Could be a good place to promote the VS Code Azure Functions plugin / extension too!

README.md
```diff
@@ -1,39 +1,68 @@
...
+
+3. Install the Durable Functions extension
+
+Run this command from the root folder of your Azure Functions app:
+```bash
 func extensions install -p Microsoft.Azure.WebJobs.Extensions.DurableTask -v 1.7.0
 ```
 
-2. Install the package
+4. Install the `durable-functions` npm package
```
mhoeger: Will this still work if installed globally, or does it need to be installed locally and deployed with your function?

README.md
kashimiz: For Node.js it is!

README.md
```diff
@@ -1,39 +1,68 @@
...
 
 ```bash
 npm install durable-functions
 ```
 
-3. Add the shim library and generator to your code:
+5. Write an activity function:
+```javascript
+module.exports = function(context) {
```
mhoeger: Could we make this example async?

README.md
```diff
@@ -43,9 +72,11 @@ module.exports = async function (context, input) {
 };
 ```
 
+**Note:** Client functions are started by a trigger binding available in the Azure Functions 2.x major version. [Read more about trigger bindings and 2.x-supported bindings.](https://docs.microsoft.com/en-us/azure/azure-functions/functions-triggers-bindings#overview)
+
 ## Samples
```
mhoeger: nit: above 3 steps are numbered "5"

kashimiz: Fixed.

README.md
```diff
@@ -1,39 +1,68 @@
...
+    context.done(null, `Hello ${context.bindings.name}!`);
+};
+```
+
+5. Write an orchestrator function:
 
 ```javascript
 const df = require('durable-functions');
-    // ... your code here
```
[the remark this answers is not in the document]
kashimiz: 👍

README.md
```diff
@@ -1,39 +1,68 @@
...
 ```
 
-4. Write your orchestration logic :
-```javascript
-yield context.df.callActivity("foo", "bar");
-```
+**Note:** Orchestrator functions must follow certain [code constraints.](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-checkpointing-and-replay#orchestrator-code-constraints)
 
-5. Write your [orchestration starter](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-instance-management#starting-instances):
+5. Write your client function:
```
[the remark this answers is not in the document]
kashimiz: Done!

README.md
kashimiz: Good call.

README.md
```diff
@@ -1,57 +1,84 @@
 # Durable Functions for Node.js
 
-This library provides a shim to write your [Durable Functions](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-overview) orchestrator functions in [Node.js](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node), using Durable's out-of-proc execution protocol. **JS orchestrators have the same [code constraints](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-checkpointing-and-replay#orchestrator-code-constraints) as C# orchestrators.**
+The `durable-functions` npm package allows you to write [Durable Functions](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-overview) for [Node.js](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node). Durable Functions is an extension of [Azure Functions](https://docs.microsoft.com/en-us/azure/azure-functions/functions-overview) that lets you write stateful functions and workflows in a serverless environment. The extension manages state, checkpoints, and restarts for you. Durable Functions' advantages include:
```
cgillum: I wonder if we should say "in JavaScript" instead of "for Node.js".

README.md
```diff
@@ -1,57 +1,84 @@
...
+* Define workflows in code. No JSON schemas or designers are needed.
+* Call other functions synchronously and asynchronously. Output from called functions can be saved to local variables.
+* Automatically checkpoint progress whenever the function schedules async work. Local state is never lost if the process recycles or the VM reboots.
 
-Not all functionality has been implemented yet and there may be significant changes to both this library and the protocol.
+You can find more information at the following links:
+
+* [Azure Functions overview](https://docs.microsoft.com/en-us/azure/azure-functions/functions-overview)
+* [Azure Functions JavaScript developers guide](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node)
+* [Durable Functions overview](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-overview)
```
cgillum: I would put the Durable Functions overview at the top of this list. We definitely want them going through that before jumping into the developer guide so they can learn about critical behaviors, like replay, etc.

README.md
```diff
@@ -1,57 +1,84 @@
...
+
+A durable function, or _orchestration_, is a solution made up of different types of Azure Functions:
```
cgillum: It sounds like you're saying a "durable function" is equivalent to "a solution made up of different types of Azure Functions", which is not really correct. In fact, I don't think we have any concept of a "durable function" in singular form.  Maybe we could say something like: "Durable Functions allows you to define an _orchestration_, which is a solution...".

README.md
```diff
@@ -1,57 +1,84 @@
...
+
+* **Activity:** the functions and tasks being orchestrated by your workflow.
+* **Orchestrator:** a function that describes the way and order actions are executed in code.
```
cgillum: This is the first time we've introduced the term "action", which may be confusing. Maybe replace "actions" with "activities"?

README.md
```diff
@@ -1,57 +1,84 @@
...
 
 ## Getting Started
 
-1. Install Durable Functions
+You can follow the [Visual Studio Code quickstart](https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode) to get started with a function chaining example, or follow the general checklist below:
 
-Run this command from the root folder of your functions app:
-```
-func extensions install -p Microsoft.Azure.WebJobs.Extensions.DurableTask -v 1.6.2
+1. Install prerequisites:
```
cgillum: I feel there is too much redundancy with the content on docs.microsoft.com, which will make it hard to maintain and has the potential to confuse readers. Ideally this repo is a place where _contributors_ can come and learn, but app developers should go either to the official docs or to the extension repo.

So I recommend we reduce the scope of this README as follows:

1. Brief explanation of what this repo is for and how it relates to azure-functions-durable-extension
2. Brief explanation of how JS is special (quick notes on the use of generators, etc.)
3. Show a few code snippets
4. Direct link to the JS quick-start on docs.microsoft.com

src/actions/callsuborchestratoraction.ts
```diff
@@ -0,0 +1,20 @@
...
+        public readonly instanceId?: string,
+        public readonly input?: unknown,
+    ) {
+        if (!functionName || typeof functionName !== "string") {
+            throw new TypeError(`functionName: Expected non-empty string but got ${typeof functionName}`);
+        }
+
+        if (instanceId && typeof instanceId !== "string") {
+            throw new TypeError(`isntanceId: Expected non-empty string but got ${typeof instanceId}`);
+        }
```
cgillum: It seems there are a lot of redundant checks. Could these be factored out into common functions?

src/actions/waitforexternaleventaction.ts
```diff
@@ -0,0 +1,14 @@
+import { ActionType, IAction } from "../classes";
+
+/** @hidden */
+export class WaitForExternalEventAction implements IAction {
+    public readonly actionType: ActionType = ActionType.WaitForExternalEvent;
+
+    constructor(
+        public readonly externalEventName: string,
+    ) {
+        if (!externalEventName || typeof externalEventName !== "string") {
```
cgillum: For my own curiosity since we seem to do this same thing in many different places: why do we need to check the type of `externalEventName` if we've already declared it as a `string` a few lines above?

src/durableorchestrationclient.ts
```diff
@@ -0,0 +1,443 @@
...
+        require_tld: false,
+        require_protocol: true,
+        require_valid_protocol: true,
+    };
+
+    /**
+     * @param context The context object of the Azure function whose body
+     * calls this constructor.
+     */
+    constructor(private context: unknown) {
```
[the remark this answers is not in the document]
kashimiz: This is a TypeScript shorthand that declares a class property and assigns whatever is passed to the constructor argument in one statement. `context` is a private property used in the [getClient](https://github.com/Azure/azure-functions-durable-js/blob/3151cfd85b67e1a29521788ee6d24423fe471458/src/durableorchestrationclient.ts#L412) method.

src/durableorchestrationclient.ts
```diff
@@ -0,0 +1,443 @@
...
+ *
+ * module.exports = df.orchestrator(function*(context) {
+ *     const client = df.getClient(context);
+ *     const instanceId = await client.startNew(req.params.functionName, undefined, req.body);
+ *
+ *     return client.createCheckStatusResponse(instanceId);
+ * });
+ * ```
+ */
+export function getClient(context: unknown): DurableOrchestrationClient {
```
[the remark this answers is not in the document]
kashimiz: Mostly to enable unit tests like [this one,](https://github.com/Azure/azure-functions-durable-js/blob/3151cfd85b67e1a29521788ee6d24423fe471458/test/unit/orchestrationclient-spec.ts#L68) which test wrongly-shaped objects getting passed in but otherwise throw compilation errors (even with explicit type assertions.)

TypeScript type-checking only happens at compile-time; by design types are stripped out during transpilation, and no type-checking happens at runtime, so defining `context's type as `unknown` versus `IFunctionContext` has (barring some corner case or nuance I don't yet understand) no effect on how the code runs, but does allow for testing possible scenarios.

src/durableorchestrationcontext.ts
```diff
@@ -0,0 +1,177 @@
...
+     * Schedules an activity function named `name` for execution.
+     *
+     * @param name The name of the activity function to call.
+     * @param input The JSON-serializable input to pass to the activity
+     * function.
+     * @returns A Durable Task that completes when the called activity
+     * function completes or fails.
+     */
+    public callActivity(name: string, input?: unknown): Task {
+        throw new Error("Not implemented");
```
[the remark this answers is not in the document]
kashimiz: Changed; agreed, that is more descriptive. The explanation for this is a weird one:

The `DurableOrchestrationContext` type is used to specify the shape of the [`df` property on `IFunctionContext`](https://github.com/Azure/azure-functions-durable-js/blob/3151cfd85b67e1a29521788ee6d24423fe471458/src/ifunctioncontext.ts#L7), which was previously `any` (as in -thing, which is not good practice.) It exists solely to allow for more explicit type definition within the code and be a place to hang documentation from. (Originally it was an interface, but typedoc doesn't render function parameter information well on interfaces or abstract classes.)

Its constructor is never called and never should be, hence the exception-throwing (and since the type isn't exported, would take some digging to do so.) Instead, the `df.orchestrator` wrapper function attaches an [anonymous object shaped like `DurableOrchestrationContext`](https://github.com/Azure/azure-functions-durable-js/blob/3151cfd85b67e1a29521788ee6d24423fe471458/src/orchestrator.ts#L38) whose method properties are bound to the actual implementations to the `IFunctionContext` at runtime. The TypeScript compiler is satisfied and the documentation generates nicely.

src/durableorchestrationclient.ts
```diff
@@ -0,0 +1,443 @@
...
+        if (!orchestratorFunctionName) {
+            throw new Error("orchestratorFunctionName must be a valid string.");
+        }
+
+        let url = this.clientData.creationUrls.createNewInstancePostUri;
+        url = url
+            .replace(this.functionNamePlaceholder, orchestratorFunctionName)
+            .replace(this.instanceIdPlaceholder, (instanceId ? `/${instanceId}` : ""));
+
+        const res = await this.webhookClient.post(new URL(url), input);
```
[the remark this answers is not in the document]
kashimiz: It does not. I'll file an issue to add that capability.

README.md
kashimiz: Noted, agreed, and changed.

src/actions/callsuborchestratoraction.ts
kashimiz: This could definitely be done as an optimization; I can put them in Utils.

src/actions/waitforexternaleventaction.ts
kashimiz: TypeScript's types are a compile-time only thing and no type-checking is done at runtime (see a later comment below), and `externalEventName` is a user-supplied piece of information. If this check was removed, it would be possible for a user to write something like

`yield context.df.waitForExternalEvent({ name: "aBadIdea", willWork: false });`

in their orchestrator, which could lead to the orchestration shim or extension schema-parsing failing and throwing a less informative error further down the stack, or the `waitForExternalEvent` method looking for a history event whose event name matched the anonymous object in question, never finding it, and causing the orchestrator to hang without an easily visible cause.

Short version: TypeScript's types are for the person writing the TypeScript, not the code running against the transpiled JS.

cgillum:
Merging for now so that we can get the README updates live. Let's open a separate PR for remaining feedback.

src/durableorchestrationclient.ts
```diff
@@ -0,0 +1,443 @@
...
+            .replace(this.reasonPlaceholder, reason);
+
+        const res = await this.webhookClient.post(new URL(url));
+        switch (res.status) {
+            case 202:
+                return;
+            case 404:
+                throw new Error(`No instance with ID '${instanceId}' found.`);
+            case 410:
+                throw new Error(`The rewind operation is only supported on failed orchestration instances.`);
```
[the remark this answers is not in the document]
kashimiz: Coming back to this post-release. I agree that it would be best to flow any messages from the extension, but the HTTP responses do not currently contain error messages. We'll have to tweak things in the extension to enable this.

src/durableorchestrationclient.ts
```diff
@@ -0,0 +1,443 @@
...
+            default:
+                throw new Error(`Webhook returned unrecognized status code ${res.status}: ${res.body}`);
```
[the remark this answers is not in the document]
kashimiz: Coming back to this post-release. I agree that it would be best to flow any messages from the extension, but the HTTP responses do not currently contain error messages. We'll have to tweak things in the extension to enable this.

src/ifunctioncontext.ts
```diff
@@ -0,0 +1,9 @@
+import { DurableOrchestrationContext } from "./classes";
+
+export interface IFunctionContext {
```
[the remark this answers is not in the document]
kashimiz: Added.

Issue #28, opened 2018-12-08
startNew throws EPROTO error locally


kashimiz:
When running locally, making calls to `DurableOrchestrationClient.startNew` to start an orchestrator function results in an EPROTO error being thrown and the orchestrator not starting.

This is due to Azure/azure-functions-host#2024, which causes the DurableOrchestrationClient class to use the wrong webhook URLs. The workaround is to set local environment variable `WEBSITE_HOSTNAME` to `localhost:<port>`, ex. `localhost:7071`.

sckelly:
Hey Kashimiz, thanks for flagging this.
I'm trying to set up the starter project as found here:
https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode

I'm getting the EPROTO error that you have mentioned here I believe, it relates to SSL wrong version number? (Just want to be sure I'm not mismatching issues). HttpStart won't kick off.
I've tried setting the Local env variable as you've mentioned in your workaround, but still no joy.

Is there any more info on this?
Thanks and all the best.

kashimiz:
Thanks for reporting you ran into this error, @sckelly. It's interesting that you're still running into that error.. Could you confirm a couple of things to rule out some possible causes?

1) What version of the Durable Functions extension do you have installed? Should be 1.7.0 in your app's `extensions.csproj` file.
2) Did you restart your editor after setting the local environment variable?
3) Are you able to start your function app and call HttpStart? (Just execute it, not necessarily return a good response.) If so, can you:
   - set a breakpoint in your HttpStart function just after your function declaration
   - call HttpStart
   - let me know the value of the following variable: `context.bindings.starter.creationUrls.createNewInstancePostUri`

sckelly:
I ended up fixing the error with your work-around mentioned above! I wasn't paying attention properly so that's my bad. I think everything is up to date: locally I can run OrchestratorFunction to get the output I want.

However when deploying to Azure I'm running into an error where I am told URL is not defined..
I know here really isn't the place for that. I shall do some googling and see what I can see!

Thanks again for the help and flagging this, it's good to know what's going on! :smile_cat:

sckelly:
Though trying to fix the deployment now, I get another issue in regards to URL not defined in the DurableOrchestrationClient.. (this may be something simple I'm missing like an env param)

kashimiz:
[...]

Issue #34, opened 2018-12-14
getStatusBy does not filter by runtime statuses


kashimiz:
Calls webhook with query parameter `runtimeStats` instead of `runtimeStatus`.

kashimiz:
This has been fixed in the durable-functions 1.1.3 release.

PR #39, opened 2019-01-10
Post-GA Fixup Part I

kashimiz:
General post-release improvements. This PR fixes issues #28 and #34, refactors portions of the code for better quality, and adds some missing type-checking.


test/testobjects/testconstants.ts
```diff
@@ -0,0 +1,25 @@
+export class TestConstants {
+    public static readonly connectionPlaceholder: string = "CONNECTION";
+    public static readonly taskHubPlaceholder: string = "TASK-HUB";
+    public static readonly hostPlaceholder: string = "HOST-PLACEHOLDER";
+
+    public static readonly testCode: string = "code=6PDWS6KrNPAOv/hw7cTDgiabhQwKTQQEnEZptmIUawunGPpOFvzDRQ==";
```
cgillum: Does this `code` have to look like a real secret?  I worry about GitHub code scanners flagging this.

test/unit/orchestrationclient-spec.ts
```diff
@@ -218,74 +186,109 @@ describe("Orchestration Client", () => {
...
+                new DurableOrchestrationStatus(
+                defaultOrchestrationName,
+                defaultInstanceId,
+                null,
+                null,
+                null,
+                null,
+                OrchestrationRuntimeStatus.Pending,
+                null,
+                null),
```
cgillum: These parameters look like they need to be indented another level.

test/testobjects/testconstants.ts
kashimiz: Ah, good point. It doesn't; will replace.

src/constants.ts
```diff
@@ -1,4 +1,21 @@
...
     public static readonly OrchestrationClientNoBindingFoundMessage: string = "An orchestration client function must have an orchestrationClient input binding. Check your function.json definition."; // tslint:disable-line max-line-length
+    public static readonly OrchestrationContextNoBindingFoundMessage: string = "Could not finding an orchestrationClient binding on context."; // tslint:disable-line max-line-length
+    public static readonly PlaceholderMessage: string = "This is a placeholder.";
+    public static readonly RewindNonFailedInstanceMessage: string = "The rewind operation is only supported on failed orchestration instances."; // tslint:disable-line max-line-length
+    public static readonly TimeoutLessThanRetryTimeoutMessage: string = "Total timeout {0} (ms) should be bigger than retry timeout {1} (ms)"; // tslint:disable-line max-line-length
+
+    public static readonly DefaultLocalHost: string = "localhost:7071";
+    public static readonly DefaultLocalOrigin: string = `http://${Constants.DefaultLocalHost}`;
```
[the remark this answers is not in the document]
kashimiz: That's a valid worry, and one that niggled at me. At the time I made this change I was concerned with having to update tests when/if the error messages change and maintaining message consistency.

test/unit/orchestrationclient-spec.ts
kashimiz: That they do. Thanks for the catch.

src/durableorchestrationclient.ts
```diff
@@ -222,33 +290,40 @@ export class DurableOrchestrationClient {
...
+                    throw new Error(Constants.InstanceNotFoundMessage
+                        .replace("{0}", instanceId));
+                case 400:
+                    throw new Error(Constants.InvalidRequestContentFormatMessage);
+                default:
+                    throw new Error(Constants.BadWebhookStatusMessage
+                        .replace("{0}", response.status.toString()));
+            }
+        } catch (error) {
+            throw new Error(error.message);
```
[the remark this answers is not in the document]
kashimiz: This was following the [guidance provided by the axios module's documentation](https://www.npmjs.com/package/axios#handling-errors). The `error` object isn't an exception/JavaScript error itself, but also includes references to the original HTTP request and other details. I'll add a clarifying note.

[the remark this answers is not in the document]
hartra344: I'm also curious here, you're using axios and by throwing just the error.message you're losing a lot of information about the error. Again I'm not entirely sure what's happening here when the error is thrown(if it haults execution or catches up the stack) but if the lost information isn't needed then I do agree with Chris on documenting what is happening here.

src/constants.ts
```diff
@@ -1,4 +1,21 @@
 /** @hidden */
 export class Constants {
+    public static readonly BadWebhookStatusMessage: string = "Webhook returned unrecognized status code {0}";
+    public static readonly CancelCompletedTaskMessage: string = "Cannot cancel a completed task.";
+    public static readonly InstanceNotFoundMessage: string = "No instance with ID '{0}' found.";
+    public static readonly InvalidFirstRetryIntervalValueMessage: string = "firstRetryIntervalInMilliseconds value must be greater than 0."; // tslint:disable-line max-line-length
+    public static readonly InvalidRequestContentFormatMessage: string = "Only application/json request content is supported"; // tslint:disable-line max-line-length
+    public static readonly InvalidStringMessage: string = "{0} must be a valid string.";
```
hartra344: NIT: since you're not using string.Format the C# way, there might be some better ways to  do this. If you want to use replace it might be better to name the replacement so it's easier to read below(rather than {0} use {stringName} or something) another good option would be to make it a lambda function so you get property names below so this would be 
```suggestion
    public static InvalidStringMessage = (stringName: string) => `{stringName} must be a valid string.`;
```

and usage would just be 'InvalidStringMessage(stringName)' rather than a chain of replace

kashimiz: I like this suggestion. It does leave the issue Chris mentioned in one of his comments, namely;

> My worry about this pattern is that it's not obvious from source code whether […] the error message actually matches the error condition in the code

I'll stick with straight strings for now.

PR #43, opened 2019-01-22
Default POST headers Content-Type to application/json

### LINKED ISSUE #42: Orchestrator values are not passed if not objects
If I have a simple string the orchestrator return no input values
```js
client.startNew(orchestrator, undefined, "bla")
```
```js
// input will be null;
```

I imagine this can be related with the change to use Axios now.
And in order to make my code work I needed to change the code in start new from
```js
const response = await axios.post(requestUrl, JSON.stringify(input));
```
to
```js
const response = await axios.post(requestUrl, input);
```

kashimiz:
Fixes #42:

- The POST requests from axios default their `Content-Type` header to `x-www-form-urlencoded`, which causes problems on our extension's end. I set a global header (added to each POST request) for `Content-Type` to be `application/json`, since we don't plan to send anything else.
- While I was updating tests I caught a second bug. The `axios` library, which is for making HTTP requests, will throw Errors in certain cases. I wanted to pare down the information we bubble up in the event one happens, hence the try-catch, but that was also catching the errors I meant to throw based on returned status codes. Using return `Promise.reject` should mean those errors don't get caught in the catch block, which was eating their messages.

### DIFF
--- a/samples/extensions.csproj
+++ b/samples/extensions.csproj
@@ -5,7 +5,7 @@
 	<DefaultItemExcludes>**</DefaultItemExcludes>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="1.7.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="1.7.1" />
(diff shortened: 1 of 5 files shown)


Nepomuceno:
I would add the fact that it is not recommended to use the axios default client but to use the axios.create instead of axios.default the default an interfere with other requests ans lead to unexpected behaviors.

PR #51, opened 2019-02-28
fix: orchestrator shouldn't swallow thrown errors fixes #48

### LINKED ISSUE #48: Errored Orchestration function is not reported as Failed in monitor/app insights
Just calling this simple Orchestration function that always throws an Error, I'd expect it to be reported as failed in Monitor or App Insights. But that's not the case - 
```

    throw Error("Failed purposefully");
});
```
I trigger this via a Queue bound Client
```

module.exports = async function (context, myQueueItem) {
    const instanceId = await client.startNew("FailPurposefully", undefined, myQueueItem);
};
```
When I check the execution in Monitor, both are reported as `success=True`. Here is the output of running it locally. Both reported as `Succeeded`
```Started orchestration with ID = '8d269cb140ff46fe977fd6158fe2a0e4'.
Executed 'Functions.QueueTrigger' (Succeeded, Id=a4ba77d1-c789-4474-9ec4-86b40e1dd909)
Executing 'Functions.FailPurposefully' (Reason='', Id=484bfb15-c17f-422a-ac4f-b9bd391b74b6)
8d269cb140ff46fe977fd6158fe2a0e4: Function 'FailPurposefully (Orchestrator)' started. IsReplay: False. Input: (196 bytes). State: Started. HubName: DurableFunctionsHub. AppName: . SlotName: . ExtensionVersion: 1.7.1. SequenceNumber: 3.
Executed 'Functions.FailPurposefully' (Succeeded, Id=484bfb15-c17f-422a-ac4f-b9bd391b74b6)
```

Just to make sure that I am not missing anything, I checked the source of [orchestrator.ts](https://github.com/Azure/azure-functions-durable-js/blob/dev/src/orchestrator.ts#L111) and looks like the `orchestrator` always passes `null` as error, essentially swallowing it.

How do we get the Durable functions flagged when Errored?

amclin:
By not passing null, the orchestrator swallows any errors that are intentionally thrown. Fixes #48 

This apparently exposes a bunch of errors in the unit tests that were being silently swallowed by the orchestrator. So unit tests are failing and will need to be addressed.

Here's an example function that can be used to test this:
```
const df = require('durable-functions')

module.exports = df.orchestrator(function * (context) {
  throw new Error(`This should throw and show up in the log.`)
})
```

### DIFF
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -109,7 +109,7 @@ export class Orchestrator {
         } catch (error) {
             log(`Error: ${error}`);
             context.done(
-                null,
+                error,
                 new OrchestratorState({
                     isDone: false,
                     actions,


kashimiz:
I'm sorry if took me so long to take a look at this. Thank you for this PR, @amclin. This will go out in an upcoming patch release.

PR #73, opened 2019-05-06
1.2.0 release payload

kashimiz:


cgillum:
BTW, the v1.8.1 release of the extension is out in case you want to target that one in your samples.

PR #87, opened 2019-06-05
Update issue templates

kashimiz:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,57 @@
...
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is. Please make an effort to fill in all the sections below; the information will help us investigate your issue.
+
+**Investigative information**
+
+- Durable Functions extension version:
+- Function App version (1.0 or 2.0):
+- Programming language used:
```
ConnorMcMahon: This seems redundant, unless any languages other than `Node` are used. Maybe change this to `Node version used`?

kashimiz: I left that in because users might be writing in JavaScript or TypeScript, but I agree that's not immediately evident. Node version is also a good field.

PR #97, opened 2019-07-25
Durable HTTP for JavaScript

cgillum:
This is the initial implementation for JavaScript support of the [Durable HTTP](https://github.com/Azure/azure-functions-durable-extension/issues/469) feature. I was interested in learning how to contribute to Durable JS, and figured this would be a good opportunity.

Things that still need to be figured out:
- [x] How to incorporate managed identities
- [x] How to incorporate `DurableHttpResponse`
- [ ] ~~What is an appropriate version number for the NPM package since this depends on Durable 2.0?~~ We'll probably just add this as a minor release and document that it requires v2 of the extension.

The first iteration was validated using the following JS orchestrator function:

```js

    const req = new df.DurableHttpRequest('GET', 'https://httpstat.us/200', null);
    const res = yield context.df.callHttp(req);
    return res;
});
```

FYI @mhoeger, @bachuv, @ConnorMcMahon, we can use this as a starting point for the hackathon project.


src/orchestrator.ts
```diff
@@ -307,6 +308,38 @@ export class Orchestrator {
         );
     }
 
+    private callHttp(state: HistoryEvent[], req: DurableHttpRequest): Task {
```
mhoeger: @cgillum - only change i might make is to this api shape? I think for node people may be more familiar with directly inputting the req object inputs

ex:
```
callHttp(state: HistoryEvent[], uri: string, method: string, content?: string | object) {
    if (content && typeof content !== 'string') {
        content = JSON.stringify(content);
    }
    const req = new DurableHttpRequest(method, uri, content);
    ...
}
```

cgillum: Yes, I love this suggestion! I didn't know how to do this in TypeScript, but I'll make this change now that you've shown it to me.

src/orchestrator.ts
```diff
@@ -308,16 +309,22 @@ export class Orchestrator {
...
             content = JSON.stringify(content);
         }
 
-        const req = new DurableHttpRequest(method, uri, content as string);
+        const req = new DurableHttpRequest(method, uri, content as string, headers, tokenSource);
         const newAction = new CallHttpAction(req);
 
         // callHttp is internally implemented as a well-known activity function
-        const httpScheduled = this.findTaskScheduled(state, "Durable:Http:Async:Activity:Function");
+        const httpScheduled = this.findTaskScheduled(state, "BuiltIn::HttpActivity");
```
ConnorMcMahon: Did we change this in the extension as well?

cgillum: Yes, it's part of this PR: https://github.com/Azure/azure-functions-durable-extension/pull/953

src/durablehttprequest.ts
```diff
@@ -0,0 +1,31 @@
...
+export class DurableHttpRequest {
+
+    /**
+     * Creates a new instance of DurableHttpRequest with the
+     * specified parameters.
+     *
+     * @param method The HTTP request method.
+     * @param uri The HTTP request URL.
+     * @param content The HTTP request content.
+     * @param tokenSource The type of OAuth token to add to the request.
```
ConnorMcMahon: NIT: Not sure I like the language "type of" to refer to token source, but I don't have a great alternative.

src/tokensource.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param resource The Azure Active Directory resource identifier of the web API being invoked.
+ * @example Gets a `ManagedIdentityTokenSource` object.
+ * ```javascript
+ * const df = require("durable-functions");
+ *
+ * module.exports = df.orchestrator(function*(context) {
+ *     return yield context.df.callHttp(
+ *         "GET",
+ *         "https://management.azure.com/subscriptions?api-version=2019-06-01",
+ *         null,
```
ConnorMcMahon: I believe this sample needs another `null` value for headers.

Also, as a NIT, I have heard [some recommendations](https://basarat.gitbooks.io/typescript/docs/javascript/null-undefined.html) to use `undefined` over `null`. Apparently the TypeScript team doesn't use `null` at all. However, `null` is all over Node, so this is likely fine.

src/durablehttprequest.ts
cgillum: Yeah, I'm not really sure how to describe this parameter succinctly. Since neither of us have good ideas, I'll just replace "type" with "source" since I can see how "type of" could be misleading.

src/tokensource.ts
cgillum: Good catch on the missing parameter.

I'll use `undefined` for both parameters instead of `null`. We seem to use `undefined` more frequently in this codebase than we do `null`, though it's not totally consistent.

Issue #54, opened 2019-03-07
Sub Orchestrator only running the first sub orchestrator


KieranDaviesV:
Hello,

I'm trying to run a few sub orchestrators from my main orchestrator, only my first orchestration works.

Main orchestrator

`const df = require("durable-functions");

module.exports = df.orchestrator(function* (context) {
    const input = context.df.getInput();
    const data = yield context.df.callActivity("dataStore", input.treeId);
    const nodes = data.nodes;
    const hubs = nodes.filter((node) => {
        return node.nodetype === "hub"
    });
    const sites = nodes.filter((node) => {
        // console.log(nodes);
        return node.nodetype === "site"
    });
    const results1 = yield context.df.callSubOrchestrator("CreateHubSite", { hub: hubs, table: data.table });

    const results2 = yield context.df.callSubOrchestrator("CreateSite", { sites: sites, table: data.table });

    context.log("results", results1);

});`
First sub orchestrator

`
const df = require("durable-functions");

module.exports = df.orchestrator(function* (context) {
    context.log("Starting hub site");
    const inputs = context.df.getInput();
    const hubArray = [];
    inputs.hub.forEach((hub, index) => {
        hubArray.push(context.df.callActivity("hubs", { hub: hub, table: inputs.table[0] }));
    });
    let results = yield context.df.Task.all(hubArray);

    return results;

});
`

Anything after results 1 doesn't run, not even the context.logs. I've looked through the examples and i cant see anything wrong ( I have been banging my head against this all day). Im not sure if this is a bug or I'm doing something wrong.

Thanks

kashimiz:
Hi @KieranDaviesV,

Thanks for reporting this issue. I've been able to reproduce it, and it seems to stem from not having explicitly-named sub-orchestrators, which confuses the module's history replay.

Flagging this as a bug because you should be able to call sub-orchestrators without specifying an instanceId, but in the meantime, have you tried changing your `callSubOrchestrator` calls so that `instanceId` isn't undefined?

KieranDaviesV:
Hey @kashimiz 

I've tried doing that and it seems to still be doing the same, anything after result one doesn't work. 

Im using context.df.instanceId

To get out the instance id, is this the right way?

hanvyj:
I just came to make this exact same issue.

The sub orchestrator completes, then the parent orchestrator instantly completes without performing any more logic.

I've tried modifying the call to pass in a child ID just like the tests do:

```
[...]

PR #108, opened 2019-10-23
Fix creating sub orchestrations without specifying instance ids

ConnorMcMahon:
Currently, only the first sub orchestration created without specifying
an instance id will ever be marked as completed. This fix aims to
address that. Addresses issue #54.

The new approach should also more aggressively find non-deterministic replays, and added some tests to emulate these errors.

### DIFF
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -254,7 +254,7 @@ export class Orchestrator {
 
     private callSubOrchestrator(state: HistoryEvent[], name: string, input?: unknown, instanceId?: string): Task {
         if (!name) {
-            throw new Error("A suborchestration function name must be provided when attempting to create a suborchestration");
+            throw new Error("A sub-orchestration function name must be provided when attempting to create a suborchestration");
         }
 
         const newAction = new CallSubOrchestratorAction(name, instanceId, input);
@@ -302,7 +302,7 @@ export class Orchestrator {
         instanceId?: string)
         : Task {
         if (!name) {
-            throw new Error("A suborchestration function name must be provided when attempting to create a suborchestration");
+            throw new Error("A sub-orchestration function name must be provided when attempting to create a suborchestration");
         }
 
         const newAction = new CallSubOrchestratorWithRetryAction(name, retryOptions, input, instanceId);
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -14,10 +14,6 @@ import { TokenSource } from "./tokensource";
 /** @hidden */
 const log = debug("orchestrator");
 
-interface SubOrchestrationCounter {
-    [key: string]: number | undefined;
-}
-
 /** @hidden */
 export class Orchestrator {
     private currentUtcDateTime: Date;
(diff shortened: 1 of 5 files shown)


src/orchestrator.ts
```diff
@@ -595,16 +595,29 @@ export class Orchestrator {
...
                     && (val as SubOrchestrationInstanceCreatedEvent).Name === name
-                    && (val as SubOrchestrationInstanceCreatedEvent).InstanceId === instanceId
                     && !val.IsProcessed;
-            })[0]
-            : undefined;
-        return returnValue as SubOrchestrationInstanceCreatedEvent;
+            });
+
+            if (matches.length > 0) {
+                // Grab the first unprocessed sub orchestration with the same name.
```
cgillum: We've already filtered the list of matches by name. Should the comment say "...with the same instance ID"?

src/orchestrator.ts
```diff
@@ -595,16 +595,29 @@ export class Orchestrator {
         state: HistoryEvent[],
         name: string,
         instanceId: string)
-        : SubOrchestrationInstanceCreatedEvent {
-        const returnValue = name
-            ? state.filter((val: HistoryEvent) => {
+        : SubOrchestrationInstanceCreatedEvent | undefined {
+
+        if (name) {
```
cgillum: I misread this logic originally. You can improve code readability and reduce complexity if you do the error checking in the `if` block and then leave out the `else` block. Something like this:

```javascript
if (!name) {
    throw new Error("...");
}

const matches = state.filter(...
```

src/orchestrator.ts
ConnorMcMahon: Substantially simplified the logic here.

src/orchestrator.ts
ConnorMcMahon: Null checking made this irrelevant :)

src/orchestrator.ts
```diff
@@ -251,12 +253,16 @@ export class Orchestrator {
     }
 
     private callSubOrchestrator(state: HistoryEvent[], name: string, input?: unknown, instanceId?: string): Task {
+        if (!name) {
+            throw new Error("A suborchestration function name must be provided when attempting to create a suborchestration");
```
cgillum: nit: in the documentation we use sub-orchestration with the hyphen. We should probably be consistent with this spelling in our error messages as well.

src/orchestrator.ts
```diff
@@ -601,58 +611,68 @@ export class Orchestrator {
     private findSubOrchestrationInstanceCreated(
         state: HistoryEvent[],
         name: string,
-        instanceId: string)
+        instanceId: string | undefined)
         : SubOrchestrationInstanceCreatedEvent | undefined {
+        const matches = state.filter((val: HistoryEvent) => {
+            return val.EventType === HistoryEventType.SubOrchestrationInstanceCreated
+                && !val.IsProcessed;
```
cgillum: So the basic idea of this iteration is that you don't need to compare names since you can simply rely on the order of these events, right?

ConnorMcMahon: That is the core idea.

I do some extra validation to error out when I detect that the order seems to be incorrect based on the fact that the history doesn't match what we are trying to execute.

PR #110, opened 2019-10-24
Turn on strict null checks

ConnorMcMahon:
Not required for any bug-fixes or features, but hopefully should help prevent null/undefined oriented bugs in the future.

### DIFF
--- a/src/history/eventraisedevent.ts
+++ b/src/history/eventraisedevent.ts
@@ -21,6 +21,6 @@ export class EventRaisedEvent extends HistoryEvent {
             this.Name = options.name;
         }
 
-        this.Input = options.input !== null ? options.input : undefined;
+        this.Input = options.input;
     }
 }
--- a/src/history/eventraisedevent.ts
+++ b/src/history/eventraisedevent.ts
@@ -3,7 +3,7 @@ import { HistoryEvent, HistoryEventOptions, HistoryEventType } from "../classes"
 /** @hidden */
 export class EventRaisedEvent extends HistoryEvent {
     public Name: string;
-    public Input: string;
+    public Input: string | undefined;
 
     constructor(
         options: HistoryEventOptions,
@@ -15,7 +15,12 @@ export class EventRaisedEvent extends HistoryEvent {
             options.timestamp,
         );
 
-        this.Name = options.name;
-        this.Input = options.input;
+        if (options.name === undefined) {
+            throw new Error("EventRaisedEvent needs a name provided.");
+        } else {
+            this.Name = options.name;
+        }
+
+        this.Input = options.input !== null ? options.input : undefined;
     }
 }
--- a/src/history/historyeventoptions.ts
+++ b/src/history/historyeventoptions.ts
@@ -2,7 +2,7 @@
 export class HistoryEventOptions {
     public details?: string;
     public fireAt?: Date;
-    public input?: string | null;
+    public input?: string;
     public instanceId?: string;
     public name?: string;
     public reason?: string;
--- a/src/history/historyeventoptions.ts
+++ b/src/history/historyeventoptions.ts
@@ -2,7 +2,7 @@
 export class HistoryEventOptions {
     public details?: string;
     public fireAt?: Date;
-    public input?: string;
+    public input?: string | null;
     public instanceId?: string;
     public name?: string;
     public reason?: string;
--- a/src/history/suborchestrationinstancefailedevent.ts
+++ b/src/history/suborchestrationinstancefailedevent.ts
@@ -16,6 +16,18 @@ export class SubOrchestrationInstanceFailedEvent extends HistoryEvent {
             options.timestamp,
         );
 
+        if (options.taskScheduledId === undefined) {
+            throw new Error("SubOrchestrationInstanceFailedEvent needs a task scheduled id provided.");
+        }
+
+        if (options.reason === undefined) {
+            throw new Error("SubOrchestrationInstanceFailedEvent needs a reason provided.");
+        }
+
+        if (options.details === undefined) {
+            throw new Error("SubOrchestrationInstanceFailedEvent needs details provided.");
+        }
+
         this.TaskScheduledId = options.taskScheduledId;
         this.Reason = options.reason;
         this.Details = options.details;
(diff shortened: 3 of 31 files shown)


src/history/eventraisedevent.ts
```diff
@@ -15,7 +15,12 @@ export class EventRaisedEvent extends HistoryEvent {
...
 
-        this.Name = options.name;
-        this.Input = options.input;
+        if (options.name === undefined) {
+            throw new Error("EventRaisedEvent needs a name provided.");
+        } else {
+            this.Name = options.name;
+        }
+
+        this.Input = options.input !== null ? options.input : undefined;
```
cgillum: Just so I understand, this basically coerces any `null` values into `undefined`?

src/history/historyeventoptions.ts
```diff
@@ -2,7 +2,7 @@
 export class HistoryEventOptions {
     public details?: string;
     public fireAt?: Date;
-    public input?: string;
+    public input?: string | null;
```
cgillum: Elsewhere you use `undefined` instead of `null`. Why is this one different?

src/history/suborchestrationinstancefailedevent.ts
```diff
@@ -16,6 +16,18 @@ export class SubOrchestrationInstanceFailedEvent extends HistoryEvent {
             options.timestamp,
         );
 
+        if (options.taskScheduledId === undefined) {
+            throw new Error("SubOrchestrationInstanceFailedEvent needs a task scheduled id provided.");
+        }
+
+        if (options.reason === undefined) {
+            throw new Error("SubOrchestrationInstanceFailedEvent needs a reason provided.");
```
cgillum: Do we require a reason? I don't expect we needed that in .NET.

src/history/historyeventoptions.ts
ConnorMcMahon: Good catch. The general recommended pattern in TypeScript is to try to stick to `undefined`, as that limits the number of checks you need to make.

src/history/eventraisedevent.ts
ConnorMcMahon: That is exactly what this is doing.

src/history/suborchestrationinstancefailedevent.ts
ConnorMcMahon: Good call.

src/history/taskfailedevent.ts
```diff
@@ -16,6 +16,18 @@ export class TaskFailedEvent extends HistoryEvent {
...
+        if (options.taskScheduledId === undefined) {
+            throw new Error("TaskFailedEvent needs a task scheduled id provided.");
+        }
+
+        if (options.reason === undefined) {
+            throw new Error("TaskFailedEvent needs a reason provided.");
+        }
+
+        if (options.details === undefined) {
+            throw new Error("TaskFailedEvent needs details provided.");
```
cgillum: I wonder if it's really worth throwing if `reason` or `details` are undefined. As far as I can tell, these are not required to be populated in DurableTask.Core.

src/orchestrator.ts
```diff
@@ -36,15 +36,14 @@ export class Orchestrator {
 
         const state: HistoryEvent[] = orchestrationBinding.history;
         const input: unknown = orchestrationBinding.input;
-        const instanceId: string = orchestrationBinding.instanceId;
+        const instanceId: string | undefined = orchestrationBinding.instanceId;
```
cgillum: In what cases do we expect an orchestration to have an `undefined` instance ID?

src/orchestrator.ts
```diff
@@ -78,7 +77,7 @@ export class Orchestrator {
         // Setup
         const gen = this.fn(context);
         const actions: IAction[][] = [];
-        let partialResult: Task | TaskSet;
+        let partialResult: Task | TaskSet | undefined;
```
cgillum: In what cases do we expect `partialResult` to be `undefined`? On line 86 it seems we assign it as either a `Task` or `TaskSet`. I apologize if this is a dumb question since I don't yet fully understand how the type system works in TypeScript.

src/orchestrator.ts
```diff
@@ -105,7 +104,10 @@ export class Orchestrator {
                 }
 
                 if ((partialResult instanceof Task || partialResult instanceof TaskSet) && partialResult.isFaulted) {
-                    g = gen.throw(partialResult.exception);
+                    if (!gen.throw) {
+                        throw new Error("Cannot properly throw the execption returned by customer code");
```
cgillum: What does this error mean? Why would `gen.throw` not be defined? Maybe we could add a comment explaining this?

test/testobjects/testentitybatches.ts
```diff
@@ -52,7 +52,7 @@ export class TestEntityBatches {
 
     public static GetCounterBatch(operations: CounterOperation[], existingState: number | undefined): EntityInputsAndOutputs {
         const id = new df.EntityId("stringstore", "stringstorekey");
-        let currentState: number;
+        let currentState: number | null = null;
```
cgillum: What does the `null = null` expression mean?

src/orchestrator.ts
ConnorMcMahon: So TypeScript defines gen.throw as an optional method. In the wonderful world of JavaScript, it is completely possible to have methods be undefined, so we have to check for it to make the null-checker happy. I'll have to look more into what scenarios `gen.throw` would actually be undefined, so that I can write a more descriptive error message.

src/orchestrator.ts
ConnorMcMahon: Good point. I might be able to restrict this further.

test/testobjects/testentitybatches.ts
ConnorMcMahon: This would be more clear as `let currentState: (number | null) = null;`. The first null is a part of the type signature. That being said, I should probably use `undefined` to follow the rest of the codebase.

src/orchestrator.ts
ConnorMcMahon: It had to do with the reference to partialResult being used before being explicitly set (line 86). I have edited the code so this is no longer required.

src/orchestrator.ts
ConnorMcMahon: So TypeScript defines gen.throw as an optional method. In the wonderful world of JavaScript, it is completely possible to have methods be undefined, so we have to check for it to make the null-checker happy. I'll have to look more into what scenarios `gen.throw` would actually be undefined, so that I can write a more descriptive error message.

Issue #81, opened 2019-05-16
continueAsNew not working in 1.8.1


waiwaizp:
Please confirm if the continueAsNew can be working in your environment, in my lab 1.8.0 and 1.8.1 both not working, 1.7.1 working well.
Thanks

Eikc:
Guessing it's linked to my comments in #57

waiwaizp:
@mpaul31 my method as below, thanks

       private static async Task<ServiceRequest> StartServiceRequestAsync(
            DurableOrchestrationContextBase context, ILogger logger)
        {
            ServiceRequest serviceRequest = null;

            try
            {
                var inputData = context.GetInput<InputData>();
                var serviceRequestId = context.InstanceId;
                var typeName = QueryableObjectBase.GetQueryableObjectTypeName(typeof(ServiceRequest));

                serviceRequest = await context.CallActivityWithRetryAsync<ServiceRequest>(
                    GetDocumentActivity.GetDocumentByIdActivityName,
                    OrchestratorSettings.OrchestrationRetryOptions,
                    (inputData, serviceRequestId, typeName)).ConfigureAwait(true);

                // some business code

                await context.CallActivityWithRetryAsync<bool>(
                    UpdateDocumentActivity.UpdateDocumentActivityName,
                    OrchestratorSettings.OrchestrationRetryOptions,
                    (inputData, serviceRequest)).ConfigureAwait(true);

                if (!string.IsNullOrWhiteSpace(lastStep.OnFunctionCompleteActivity))
                {
                    await context.CallActivityWithRetryAsync(
                        lastStep.OnFunctionCompleteActivity,
                        OrchestratorSettings.OrchestrationRetryOptions,
                        lastStep).ConfigureAwait(true);
                }

                if (nextStep.State != ActionChainItemState.Pending)
                    return serviceRequest;

                context.ContinueAsNew(inputData);
                
                logger.LogAuditEvent(context, EventName, ("NewValue", lastStep));
                return serviceRequest;
            }
            catch (Exception e)
            {
                throw;
            }
            finally
            {
                await RunStepCompletionTasksSafetyAsync(context, serviceRequest, logger).ConfigureAwait(true);
            }
        }

waiwaizp:
Hi There, Any updates? Thanks

jeffhollan:
Sebastian's fix here may have resolved.  @kashimiz can check it out on backlog

waiwaizp:
[...]

Issue #57, opened 2019-03-25
JS context.df.continueAsNew not working in 1.8.0


deman4ik:
**Describe the bug**
contunueAsNew stopped working after updating to 1.8.0
in 1.7.1 it works fine

**Investigative information**

- Durable Functions extension version: 1.8.0
- Function App version (1.0 or 2.0): 2.0
- Programming language used: JS
- durable-functions npm module version: 1.1.4

**To Reproduce**
https://github.com/deman4ik/funcs-durable-js

- func extensions install
- npm install
- func start

1. POST http://localhost:7071/api/HttpTrigger with any not null data in body
2. Check HubHistory

**Expected behavior**
Orchestrator started with random not null input
Checks Input
continueAsNew with new state (false)
Orchestrator started with "false" input
Checks Input
Orchestrator finished with result "false"

**Actual behavior**
Orchestrator started with random not null input
Checks Input
Orchestrator finished with result from input

kashimiz:
Thanks for reporting this issue, @deman4ik , and for including a repro! It looks like the npm library is returning the expected execution payload, but the extension isn't consuming it properly. I'll investigate this one further.

deman4ik:
Any ETA? We need some features from 1.8.0

kashimiz:
My apologies for the delay in response @deman4ik. It looks like this regression is a result of this [bugfix](https://github.com/Azure/azure-functions-durable-extension/commit/e43843d713de9884f5e08d5cda9fbfd9a62795b0). I don't have an ETA on a fix just yet but will keep this thread updated.

Eikc:
Is there any workaround for making the durable functions work with continueAsNew until there is a fix ?

deman4ik:
@Eikc We just downgraded to 1.7.1 Durable Functions extension version

Eikc:
@deman4ik Do you still use the 1.1.4 version of the durable-functions package ? Or does that need to be downgraded as well?

deman4ik:
@Eikc yep still 1.1.4 npm package

Eikc:
Hey @kashimiz 

I hate to ask, but is there any news on this issue? or anyway that I or others can help with this great package? I really need to get up on Durable 1.8.1 and this is sadly blocking me from using it..

Anyway, I would love to help. I just need a pointer in the right direction getting started :)

kashimiz:
[...]

PR #114, opened 2019-10-25
Rework generator result handling logic

ConnorMcMahon:
Fixes continueAsNew() for 1.8.0 and later, as well as handles some edge
cases, like customers yielding non-Task objects and returning unyielded
Tasks. Addresses #57 and #81.

### DIFF
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -86,6 +86,29 @@ export class Orchestrator {
             let g = gen.next(undefined);
 
             while (true) {
+
+                if (!(g.value instanceof Task || g.value instanceof TaskSet)) {
+                    if (!g.done) {
+                        // The orchestrator must have yielded a non-Task related type,
+                        // so just return execution flow with what they yielded back.
+                        g = gen.next(g.value);
+                        continue;
+                    } else {
+                        log("Iterator is done");
+                        // The customer returned an absolute type.
+                        context.done(
+                            null,
+                            new OrchestratorState({
+                                isDone: true,
+                                output: g.value,
+                                actions,
+                                customStatus: this.customStatus,
+                            }),
+                        );
+                        return;
+                    }
+                }
+
                 partialResult = g.value as Task | TaskSet;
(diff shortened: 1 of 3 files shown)


src/orchestrator.ts
```diff
@@ -86,14 +86,51 @@ export class Orchestrator {
             let g = gen.next(undefined);
 
             while (true) {
+
+                if (!(g.value instanceof Task || g.value instanceof TaskSet)) {
+                    if (!g.done) {
+                        // The orchestrator must have yielded a non-Task related type,
+                        // so just return execution flow with what they yielded back.
+                        g = gen.next(g.value);
```
cgillum: So this case handles a user error? For example, they yielded a `5` instead of the result of `context.df.callActivity(...)`?

src/orchestrator.ts
```diff
@@ -86,14 +86,51 @@ export class Orchestrator {
...
+                        continue;
+                    } else {
+                        log("Iterator is done");
```
cgillum: Where does this log statement go? Does it show up in the user logs?

src/orchestrator.ts
```diff
@@ -106,21 +143,24 @@ export class Orchestrator {
...
-                if ((partialResult instanceof Task || partialResult instanceof TaskSet) && partialResult.isFaulted) {
+                if (partialResult.isFaulted) {
                     if (!gen.throw) {
                         throw new Error("Cannot properly throw the execption returned by customer code");
                     }
                     g = gen.throw!(partialResult.exception);
                     continue;
                 }
 
+                // Handles the case where an orchestration completes with a returns a return value of
```
cgillum: Typo?

src/orchestrator.ts
ConnorMcMahon: Correct. There is no reason we can't just gracefully return the flow back to their code in this case.

src/orchestrator.ts
ConnorMcMahon: Not sure. I copied it from the other place where we mark the iterator as completed.

Issue #123, opened 2019-11-20
Contribution Guide


priyaananthasankar:
This issue tracks the need for a contributor's guide similar to the one written for https://github.com/Azure/azure-functions-durable-extension aimed towards MacOS and Windows.

davidmrdavid:
This issue is done now! Thanks @priyaananthasankar

PR #124, opened 2019-11-20
Contributor's Guide for Durable JS

priyaananthasankar:
https://github.com/Azure/azure-functions-durable-js/issues/123

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,6 +72,18 @@ The following extensions should be installed if using Visual Studio Code for deb
 - Create a Durable Functions Orchestrator for FunctionChaining pattern using [starter templates](https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode)
   Note: In this starter template, ignore the line that says: "On a Mac or Linux computer, you must set the AzureWebJobsStorage property to the connection string of an existing Azure storage account". We will be setting up the AzureWebJobsStorage property to `UseDevelopmentStorage=true`
 
+- Then, in your sample code, instead of requiring the `npm`-hosted version of the code, directly refer to your local changes. There are many ways of going about this, but a simple solution is changing a reference to `require("durable-functions")` to `require("<your-local-path-to-my-this-repo>")`.
+
+- Finally, start your VSCode editor, click Debug -> Start Debugging. This will internally start `func host start` through core tools and provides the orchestrator client URL.
+After doing this, you should be able to add breakpoints to test your changes for Durable JavaScript!
+
+
+### Debugging end-to-end with the Durable Extension
+
+In some advanced scenarios, you may want to inspect how your repo interacts with the underlying Durable Extension. In these settings, we recommend using Visual Studio.
+
+Two changes are necessary in your JavaScript sample code to make this work.
+
 - In host.json, remove the extensionsBundle portion to enable specific version debugging. Provide a hub name (else remove the entire extensions portion to default to DurableFunctionsHub) Here's how the host.json should look like:
 
 ```
@@ -85,7 +97,7 @@ The following extensions should be installed if using Visual Studio Code for deb
 }
 ```
 
-- `func extensions install` (this will install an extensions.csproj that contains the version of DurableTask as seen below). We recommend using version `2.2.0` or greater.
+- Create an `extensions.csproj` file containing the version of the DurableTask extension that you wish to use. We recommend using the latest version on [nuget.org](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.DurableTask/). To do this, take the sample `extensions.csproj` shown below, update the version of the `Microsoft.Azure.WebJobs.Extensions.DurableTask` dependency, and save it at the root of your sample.
 
 ```xml <ItemGroup>
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.2.0" />
@@ -93,26 +105,18 @@ The following extensions should be installed if using Visual Studio Code for deb
   </ItemGroup>
 ```
 
-- Then, in your sample code, instead of requiring the `npm`-hosted version of the code, directly refer to your local changes. There are many ways of going about this,
-but a simple solution is changing a reference to `require("durable-functions")` to `require("<your-local-path-to-my-this-repo>")`.
-
-- Finally, start your VSCode editor, click Debug -> Start Debugging. This will internally start `func host start` through core tools and provides the orchestrator client URL.
-After doing this, you should be able to add breakpoints to test your changes for Durable JavaScript!
+- Finally, open a terminal and run `func extensions install`. This will install the dependencies listed in your `extensions.csproj` file.
 
 
-### Debugging end-to-end with the Durable Extension
-
-In some advanced scenarious, you may want to inspect how your repo interacts with the underlying Durable Extension. In these settings, we
-recommend using Visual Studio.
-
+Now you are ready to start debugging Durable Functions end-to-end. Follow the instructions below:
 1. Open the Azure Storage Explorer and connect to the local storage emulator or the storage account you are using.
 2. In the VSCode editor for durable-js click Debug -> Start Debugging. This will internally start `func host start` through core tools and provides the orchestrator client URL
 3. In the Visual Studio editor for DurableTask, click Debug -> .NET Core Attach Process and search for `func host start` process and attach to it.
 4. Add a breakpoint in both editors and continue debugging.
 
-## Testing changes locally (Windows)
+## Testing with emulated local storage (Windows)
 
-Follow all the steps above, use the Azure Storage Emulator for windows to simulate the storage account, and optionally use Visual Studio to debug the .NET Durable Extension.
+When debugging your changes, you may want to use your local storage instead of an Azure account. To do this, follow all the steps above, use the Azure Storage Emulator for windows to simulate the storage account, and optionally use Visual Studio to debug the .NET Durable Extension.
 
 ## Getting help
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@
 
 The general flow for making a change to the script host is:
 1. 🍴 Fork the repo (add the fork via `git remote add me <clone url here>`
-2. 🌳 Create a branch for your change (generally use dev) (`git checkout -b my-change`)
+2. 🌳 Create a branch for your change (generally branch from dev) (`git checkout -b my-change`)
 3. 🛠 Make your change
 4. ✔️ Test your changes
 5. ⬆️ Push your changes to your fork (`git push me my-change`)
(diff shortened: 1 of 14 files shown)


priyaananthasankar:
cc: @ConnorMcMahon : Can you please review this?

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+
+The following extensions should be installed if using Visual Studio Code for debugging:
+
+- C# for Visual Studio Code (powered by OmniSharp)
+- Azure Functions Extensions for Visual Studio Code v0.19.1 and above.
+
+### Setting up durable-js debugging
+
+- Create a Durable Functions Orchestrator for FunctionChaining pattern using [starter templates](https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode)
+  Note: In this starter template, ignore the line that says: "On a Mac or Linux computer, you must set the AzureWebJobsStorage property to the connection string of an existing Azure storage account". We will be setting up the AzureWebJobsStorage property to `UseDevelopmentStorage=true`
```
ConnorMcMahon: If you found that Azurite is sufficient for local dev, it may be a good idea for us to update this line in our template to give that option.

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+### Debugging end-to-end
+
+1. Open the Azure Storage Explorer and connect to the local storage emulator or the storage account you are using.
+2. In the VSCode editor for durable-js click Debug -> Start Debugging. This will internally start `func host start` through core tools and provides the orchestrator client URL
+3. In the VSCode editor for DurableTask, click Debug -> .NET Core Attach Process and search for `func host start` process and attach to it.
+4. Add a breakpoint in both editors and continue debugging.
+
+## Testing changes locally (Windows)
+
+Follow all the steps above in Visual Studio and use Azure Storage Emulator for Windows to simulate the storage account.
```
ConnorMcMahon: I would clarify that the only steps above to do in Visual Studio is the Durable Extension .NET debugging. Even then, Windows JS developers may still feel more comfortable using VS Code, so maybe reword with something like the below:

Follow all the steps above, use the Azure Storage Emulator for windows to simulate the storage account, and optionally use Visual Studio to debug the .NET Durable Extension.

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+    - VS Code (or) Visual Studio
+- Tools
+    - [Azurite V2](https://github.com/Azure/Azurite/tree/legacy-master) (for MacOS) (or) [Azure Storage Emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator) (or) Storage account in Azure
+    - [Azure Functions Core Tools](https://github.com/Azure/azure-functions-core-tools) v2.7.x and above.
+
+## Change flow
+
+The general flow for making a change to the script host is:
+1. 🍴 Fork the repo (add the fork via `git remote add me <clone url here>`
+2. 🌳 Create a branch for your change (generally use dev) (`git checkout -b my-change`)
```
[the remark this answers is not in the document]
priyaananthasankar: Done. https://github.com/Azure/azure-functions-durable-js/pull/124/commits/7f085c2fb23d722fb07afce41cf6f653b739a2ee

CONTRIBUTING.md
priyaananthasankar: Azurite V2 (which is not the version officially recommended by Azurite) works. V3 does not support some storage mechanisms. Does it make sense to add to starter templates now or wait until Azurite stabilizes?

CONTRIBUTING.md
priyaananthasankar: Done https://github.com/Azure/azure-functions-durable-js/pull/124/commits/7f085c2fb23d722fb07afce41cf6f653b739a2ee

priyaananthasankar:
@scgbear : The primary instructions are to make changes in the JS library itself.

priyaananthasankar:
@ConnorMcMahon : Sorry Got delayed reworking this, but here it is. Do let me know if it looks good to be merged.

CONTRIBUTING.md
ConnorMcMahon: I see. It looks like there are no plans to support Tables in V3, due to Cosmos DB having a separate emulator for Azure Tables (which unfortunately, is NOT cross platform). You can see the [discussion here](https://github.com/Azure/Azurite/issues/253). With that in mind, maybe Azurite V2 is the best local dev option?

priyaananthasankar: @ConnorMcMahon : Yes given those limitations v2 is the best (as of now)

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Pre-reqs
+
+- OS
+    - MacOS (or) Windows10
+- Language Runtimes
+    - .NET Core 2.0
+    - Python 3.6.x
+- Editor
+    - VS Code (or) Visual Studio
```
cgillum: Should we add links to the language runtime installation downloads and to the editors?

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+      "hubName": "{hubName}"
+    }
+  }
+}
+```
+
+- `func extensions install` (this will install an extensions.csproj that contains the version of DurableTask as seen below)
+
+```xml <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="1.8.2" />
```
cgillum: I suggest we point people to 2.0.0 or greater.

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+    <PackageReference Include="Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator" Version="1.1.0" />
+  </ItemGroup>
+```
+
+### Setting up Durable Extension .NET debugging
+
+- Download the source code of DurableTask extension from [here](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.DurableTask)
```
cgillum: Is it necessary to download the source code? We use SourceLink today and Visual Studio knows how to automatically download source code at debug time (does VS Code know how to do this?). I would hope that downloading the extension project would only be needed if they need to modify the extension code (super-advanced scenario).

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+            "request": "attach",
+            "processId": "${command:pickProcess}",
+            "requireExactSource": false,
+            "justMyCode": false
+ }
+ ```
+
+### Debugging end-to-end
+
+1. Open the Azure Storage Explorer and connect to the local storage emulator or the storage account you are using.
```
cgillum: Should we add Azure Storage Explorer (with a download link) to the prereqs list above?

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+2. In the VSCode editor for durable-js click Debug -> Start Debugging. This will internally start `func host start` through core tools and provides the orchestrator client URL
+3. In the VSCode editor for DurableTask, click Debug -> .NET Core Attach Process and search for `func host start` process and attach to it.
```
cgillum: Could this be considered a more advanced scenario, for people who are trying to debug both the extension and the JS library?

CONTRIBUTING.md
```diff
@@ -0,0 +1,125 @@
...
+4. Add a breakpoint in both editors and continue debugging.
+
+## Testing changes locally (Windows)
+
+Follow all the steps above, use the Azure Storage Emulator for windows to simulate the storage account, and optionally use Visual Studio to debug the .NET Durable Extension.
```
cgillum: In my experience, there is a mechanism for pointing your dev environment to get the sources from durable-functions npm package from you local enlistment rather than from npm.org. Should we add those instructions here?

CONTRIBUTING.md
davidmrdavid: Good idea, done!

CONTRIBUTING.md
davidmrdavid: There's already a link to it in the "Tools" section. I have updated the link to point to the download URL though

CONTRIBUTING.md
davidmrdavid: I have updated the example `.csproj` and pointed out that we recommend sticking to versions equal or higher to `2.2.0`

CONTRIBUTING.md
davidmrdavid: You are right, I have addressed it as such in my latest commit :)

CONTRIBUTING.md
davidmrdavid: I'm reworking this section to make it clear that, if users want to debug the .NET side of things, they should use Visual Studio

CONTRIBUTING.md
davidmrdavid: Added!

CONTRIBUTING.md
```diff
@@ -0,0 +1,119 @@
...
+  "version": "2.0",
+  "extensions": {
+    "durableTask": {
+      "hubName": "{hubName}"
+    }
+  }
+}
+```
+
+- `func extensions install` (this will install an extensions.csproj that contains the version of DurableTask as seen below). We recommend using version `2.2.0` or greater.
```
ConnorMcMahon, line 88: Instead of 2.2.0 or greater, why don't we just say "the latest version on nuget.org" and point to the package link.

CONTRIBUTING.md
```diff
@@ -0,0 +1,119 @@
...
+
+- C# for Visual Studio Code (powered by OmniSharp)
+- Azure Functions Extensions for Visual Studio Code v0.19.1 and above.
+
+### Setting up Durable JavaScript using sample code
+
+- Create a Durable Functions Orchestrator for FunctionChaining pattern using [starter templates](https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode)
+  Note: In this starter template, ignore the line that says: "On a Mac or Linux computer, you must set the AzureWebJobsStorage property to the connection string of an existing Azure storage account". We will be setting up the AzureWebJobsStorage property to `UseDevelopmentStorage=true`
+
+- In host.json, remove the extensionsBundle portion to enable specific version debugging. Provide a hub name (else remove the entire extensions portion to default to DurableFunctionsHub) Here's how the host.json should look like:
```
ConnorMcMahon, line 75: This whole section about extension development could probably move down into the `Debugging end-to-end with the Durable Extension` section, as it technically isn't required if you are just doing development on our npm package.

CONTRIBUTING.md
ConnorMcMahon: I think we should add that new chunk to this section, as opposed to the `Setting up Durable JavaScript using sample code` section.

CONTRIBUTING.md
davidmrdavid: Addressed in next commit.

CONTRIBUTING.md
davidmrdavid: Moved it :)

CONTRIBUTING.md
davidmrdavid: resolved offline

PR #127, opened 2019-11-28
[Fix] Changed callHttp signature to match exposed type

andreujuanc:
Hi!, 

So I found that the data passed to the callHttp was wrong, and figure there were two different paths on passing parameters to functions, one is a comma-separated arguments, and the other being an object containing both of them.

Obviously I would love to unify this , but it might require braking changes for already exposed types.

## _callActivity_ 
We can see that a comma-separated parameters was used:
#### Described in Context Type:    
`callActivity(name: string, input?: unknown)`
#### Implemented in Orchestrator:  
`callActivity(state: HistoryEvent[], name: string, input?: unknown)`

## _callHttp_ 
Exposed type does not match the implemented signature of the orchestrator's method.
#### Described in Context Type:    
`callHttp(req: DurableHttpRequest)`
#### Implemented in Orchestrator:  
`callHttp(state: HistoryEvent[], method: string, uri: string, content?: string | object, headers?: { [key: string]: string }, tokenSource?: TokenSource)`

I would have prefer to fix this in a way that keep consistency high, BUT, I'd rather not break code.
Tho, no one might be using this method because it's not working at all, so we can consider to undo this PR and remove the `DurableHttpRequest` from the exposed type(DurableOrchestrationContext) 

Best,
JC

### DIFF
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -394,18 +394,12 @@ export class Orchestrator {
         );
     }
 
-    private callHttp(
-        state: HistoryEvent[],
-        method: string,
-        uri: string,
-        content?: string | object,
-        headers?: { [key: string]: string },
-        tokenSource?: TokenSource) {
+    private callHttp(state: HistoryEvent[], req: DurableHttpRequest) {
+        let { content } = req;
         if (content && typeof content !== "string") {
             content = JSON.stringify(content);
         }
 
-        const req = new DurableHttpRequest(method, uri, content as string, headers, tokenSource);
         const newAction = new CallHttpAction(req);
 
         // callHttp is internally implemented as a well-known activity function
--- a/test/testobjects/TestOrchestrations.ts
+++ b/test/testobjects/TestOrchestrations.ts
@@ -1,4 +1,5 @@
 import * as df from "../../src";
+import { DurableOrchestrationContext } from "../../src/classes";
 
 export class TestOrchestrations {
     public static AnyAOrB: any = df.orchestrator(function*(context: any) {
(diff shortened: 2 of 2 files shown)


andreujuanc:
BTW, this issue would have been caught if the unit tests were not using `any`.

andreujuanc:
Let me know if a rebase / squash is required for this PR, and I will do it. But I'll keep it like this just in case we want to make more changes to this.

Cheers

anthonychu:
Just realized I created #128 which is the alternative solution that you suggested. I think the original intent was to take multiple parameters, as suggested by this [comment](https://github.com/Azure/azure-functions-durable-js/blob/94f1fb34035b251e8b7981186a0231bcd79760db/src/tokensource.ts#L9-L15).

We could (should?) support both signatures. We probably need to keep `callHttp(method: string, uri: string, content?: string | object, headers?: { [key: string]: string }, tokenSource?: TokenSource)` because people might be using it successfully now and it'd be a breaking change to remove it. `callHttp(req: DurableHttpRequest)` would be nice to have but it doesn't work today.

@ConnorMcMahon @cgillum thoughts?

andreujuanc:
It's so interesting to encounter different approaches to solve the same problem.

I agree that the intent must have been to have multiple parameters, liked I described on `callActivity`. But since that would be a breaking change, I decided to fix things "down the road" and not to disturb users.

Personally, for TS, I prefer the object signature a bit more, since it looks like named parameters :stuck_out_tongue: 

But yes, it'd be cool to support both of them, but maybe it's a bit overkill.

Anyway, I would suggest to slowly refactor all signatures to be more coherent.

ConnorMcMahon:
@andreujuanc 

Thanks for your contribution. Since we have already merged a fix, I am going to close this PR.

anthonychu:
Thanks @andreujuanc. There might still be an opportunity to take this contribution if we want to start supporting multiple signatures: #143

PR #128, opened 2019-11-29
Update DurableOrchestrationContext.callHttp signature

### LINKED ISSUE #119: callHttp fails to make HTTP call
**Describe the bug**
`context.df.callHttp` doesn't work.

**Investigative information**

- Durable Functions extension version: 2.0.0
- durable-functions npm module version: 1.3.1
- Language (JavaScript/TypeScript) and version: JS
- Node.js version: 10.16.3

**To Reproduce**

1. Clone this repo: https://github.com/anthonychu/test-durable-js
1. ```bash
    npm i
    func extensions install
    cp local.settings.SAMPLE.json local.settings.json
    # Add storage connection string to local.settings.json
    func start
    ```
1. Call `http://localhost:7071/api/orchestrators/DurableFunctionsOrchestratorJS` to start the orchestrator that makes an http call.
1. Get the `statusQueryGetUri` in the response and use it to check the status. It should show an error:
    ```json
    {
        "name":"DurableFunctionsOrchestratorJS","instanceId":"d38d33aea86342c28a27b571def7e720",
        "runtimeStatus":"Failed",
        "input":null,
        "customStatus":null,
        "output":"Unexpected token while deserializing object: PropertyName. Path 'actions[0][0].actionType', line 12, position 21.",
        "createdTime":"2019-11-17T06:54:36Z",
        "lastUpdatedTime":"2019-11-17T06:54:37Z"
    }
    ```

@ConnorMcMahon @cgillum

anthonychu:
Fixes #119

### DIFF
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -1,5 +1,4 @@
-import { DurableHttpRequest, DurableLock, EntityId, ITaskMethods, LockState,
-     RetryOptions, Task, TimerTask } from "./classes";
+import { EntityId, ITaskMethods, RetryOptions, Task, TimerTask } from "./classes";
 import { TokenSource } from "./tokensource";
 
 /**
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -1,5 +1,6 @@
 import { DurableHttpRequest, DurableLock, EntityId, ITaskMethods, LockState,
      RetryOptions, Task, TimerTask } from "./classes";
+import { TokenSource } from "./tokensource";
 
 /**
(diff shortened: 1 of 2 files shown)


mhoeger:
@anthonychu - can you also include `"strictBindCallApply": true` to tsconfig.json? This would catch this issue in the future. Thanks!

anthonychu:
@mhoeger @cgillum Looks like we need TypeScript 3.2 for the `strictBindCallApply` flag. Currently on ~3.1.6.

cgillum:
Ah, bummer. Let's wait to add this then. I don't want to take on the risk of upgrading the TypeScript compiler quite yet (I'm told they don't follow SemVer, which means there could be breaks).

PR #129, opened 2019-12-03
Fix error message object to string conversion

### LINKED ISSUE #53: Unhelpful error message when orchestration not found.
I was following the sample [on the docs site](https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode) but received the following error when hitting the URL for `httpStart`:

```
System.Private.CoreLib: Exception while executing function: Functions.HttpStart. System.Private.CoreLib: Result: Failure
Exception: Error: [object Object]
Stack: Error: [object Object]
    at DurableOrchestrationClient.<anonymous> (C:\_Projects\tmp03\node_modules\durable-functions\lib\src\durableorchestrationclient.js:247:43)
    at Generator.next (<anonymous>)
    at fulfilled (C:\_Projects\tmp03\node_modules\durable-functions\lib\src\durableorchestrationclient.js:4:58)
    at process._tickCallback (internal/process/next_tick.js:68:7).
```

I then cloned the project repo (b1512bf7b365b4ff72046934cc9c008c042aff08, current `dev` HEAD) to try the samples and received the same error. I've attached the log file from the function run: [logs.txt](https://github.com/Azure/azure-functions-durable-js/files/2929384/logs.txt). Not quite sure where to go next in diagnosis of the problem, but happy to help.

## Environment

* OS: Windows 10
* `node -v`: 10.15.2
* `func -v`: 2.4.419

anthonychu:
Fixes #53

### DIFF
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -536,7 +536,7 @@ export class DurableOrchestrationClient {
             if (response.data && response.status <= 202) {
                 return (response.data as HttpManagementPayload).id;
             } else  {
-                return Promise.reject(new Error(response.data as string));
+                return Promise.reject(new Error(JSON.stringify(response.data)));
             }
         } catch (message) {
             throw new Error(message.error);


anthonychu:
I ran into [one more](https://github.com/Azure/azure-functions-durable-js/blob/94f1fb34035b251e8b7981186a0231bcd79760db/src/durableorchestrationclient.ts#L542) last week when testing other stuff, but in that case I think it printed `Error` instead of `[object object]`. I changed that particular instance to just throw the string and it printed the error.

Feels like exceptions might be dealt with differently in different places and maybe a better place to address these issues is where they are getting serialized, deserialized, or output.

ConnorMcMahon:
@mhoeger Do you have any recommendations on consistent ways to emit error messages in JS from the language worker, so they show up in customer logs?

cgillum:
I'm going to go ahead and take this fix for now, but I agree we should do an audit to see whether there are other problems with how we surface errors in this SDK.

mhoeger:
@ConnorMcMahon - sorry about the delay! Yeah that's interesting - the problem with trying to handle this issue where the exception is about to be serialized is that at that point, the useful data has likely already been lost. When a `new Error({ with: "myObject"});` is created, the thing that is thrown is an `Error` object with properties `message` and `stack`, where `message` is `[object Object]`. 

If someone did something like `throw { my: "object"};`, then special handling on the Node.js worker would be helpful (but I don't think that's too common).

One thing that might help is using a custom `Error` class that can takes objects as input:
```
class DurableError extends Error {
    constructor(error: string | object) {
        if (typeof error == "object") {
            error = JSON.stringify(error);
        }
        super(error);
    }
}
```

I think what is safest though is to look for things that are cast using "as". This is a TypeScript-only construct, means that it only affects how the typescript => javascript transpiler yells at you for thinking you've done something wrong. It's deceptive because unlike c#, TypeScript code that casts doesn't actually alter the data.

anthonychu:
Thanks @mhoeger. I've created a new issue to track the other instances where errors aren't printed correctly. #142

PR #136, opened 2020-01-07
Fix tokenSource format issue in callHttp API

cgillum:
There was an issue with the `callHttp` API where it didn't correctly encode token source values in a way that was understood by the Durable extension. If a user tried to use managed identities, they would see an error similar to the following:

> Could not create an instance of type Microsoft.Azure.WebJobs.Extensions.DurableTask.ITokenSource. Type is an interface or abstract class and cannot be instantiated. Path 'resource', line 11, position 25."

The problem was that the JSON returned by Durable JS that represented a `callHttp` action was missing the `kind` field that tells the durable extension parser how to interpret the token source data. The implementation of this field made incorrect assumptions about how TypeScript works. This field has been fixed.

The unit test for this was unfortunately flawed. It verified that `callHttp` produced the right TypeScript output but it failed to verify whether that TypeScript object correctly represented the JSON expected by the durable extension. I've fixed this test as well as part of the PR.

**Pending tasks**
- [x] Test end-to-end on Functions 2.0 host
- [x] Add sample function to demonstrate `callHttp` usage

FYI @mhoeger this is the issue you ran into!

### DIFF
--- a/samples/ListAzureSubscriptions/index.js
+++ b/samples/ListAzureSubscriptions/index.js
@@ -1,0 +1,13 @@
+const df = require("durable-functions");
+
+module.exports = df.orchestrator(function*(context) {
+    // More information on the use of managed identities in the callHttp API:
+    // https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-http-features#managed-identities
+    var res = yield context.df.callHttp(
+        "GET",
+        "https://management.azure.com/subscriptions?api-version=2019-06-01",
+        undefined,
+        undefined,
+        new df.ManagedIdentityTokenSource("https://management.core.windows.net"));
+    return res;
+});
--- a/src/tokensource.ts
+++ b/src/tokensource.ts
@@ -17,7 +17,7 @@
  */
 export class ManagedIdentityTokenSource {
     /** @hidden */
-    public kind: "AzureManagedIdentity";
+    public readonly kind: string = "AzureManagedIdentity";
 
     /**
      * Returns a `ManagedIdentityTokenSource` object.
--- a/samples/ListAzureSubscriptions/function.json
+++ b/samples/ListAzureSubscriptions/function.json
@@ -1,0 +1,9 @@
+{
+  "bindings": [
+  {
+    "name": "context",
+    "type": "orchestrationTrigger",
+    "direction": "in"
+    }
+  ]
+}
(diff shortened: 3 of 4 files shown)


src/tokensource.ts
```diff
@@ -17,7 +17,7 @@
  */
 export class ManagedIdentityTokenSource {
     /** @hidden */
-    public kind: "AzureManagedIdentity";
+    public readonly kind: string = "AzureManagedIdentity";
```
ConnorMcMahon: Hmm, was this field not being serialized before? If so, what about this change fixed that.

mhoeger: It's a bit strange that TypeScript lets you do this, but `kind: "AzureManagedIdentity;"` is declaring "kind" as an undefined variable that will only be able to be assigned to the string "AzureManagedIdentity". So `kind === undefined` in this case

cgillum: This was my attempt at using [Discriminated Unions](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) in TypeScript, which Connor I think you recommended me to look into. I was surprised to see that the field was not included in the transpiled JavaScript at all.

cgillum: But yes, I verified end-to-end that this change fixes the issue.

cgillum:
Thanks for the review. In the latest iteration of this PR I've added a sample that shows how to fetch the list of all Azure Subscriptions available to the current identity.

samples/ListAzureSubscriptions/index.js
```diff
@@ -0,0 +1,13 @@
...
+module.exports = df.orchestrator(function*(context) {
+    // More information on the use of managed identities in the callHttp API:
+    // https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-http-features#managed-identities
+    var res = yield context.df.callHttp(
+        "GET",
+        "https://management.azure.com/subscriptions?api-version=2019-06-01",
+        undefined,
+        undefined,
+        new df.ManagedIdentityTokenSource("https://management.core.windows.net"));
+    return res;
```
ConnorMcMahon: Do you actually want to return the entire response here, or just the body?

cgillum: Ah, yeah, it's probably cleaner to just return the response content. But for now I think I'll leave it as-is. I guess if nothing else it allows people to see the full ARM response, which might be interesting.

PR #145, opened 2020-01-14
Fix nondeterminism on Activity/Suborchestration failures

### LINKED ISSUE #125: Exception in activity function fails orchestrator with non-deterministic error
**Describe the bug**
An exception in an activity results in a non-deterministic error.

**Investigative information**

- Durable Functions extension version: 2.0.0
- durable-functions npm module version: 1.3.1
- Language (JavaScript/TypeScript) and version: JS
- Node.js version: 10.16.3

**To Reproduce**

1. Clone this repo: https://github.com/anthonychu/test-durable-js
1. ```bash
    npm i
    func extensions install
    cp local.settings.SAMPLE.json local.settings.json
    # Add storage connection string to local.settings.json
    func start
    ```
1. Call `http://localhost:7071/api/orchestrators/OrchestratorWithBadActivity` to start the orchestrator that calls an activity that throws an exception.
1. Get the `statusQueryGetUri` in the response and use it to check the status. I think it should show the actual error, but instead it shows a non-deterministic error:
    ```json
    {
      "name": "OrchestratorWithBadActivity",
      "instanceId": "4db326c0a2d34fdbad47cf134e6ddf1c",
      "runtimeStatus": "Failed",
      "input": null,
      "customStatus": null,
      "output": "Non-Deterministic workflow detected: TaskScheduledEvent: 0 TaskScheduled BadActivity",
      "createdTime": "2019-11-20T22:27:37Z",
      "lastUpdatedTime": "2019-11-20T22:27:38Z"
    }
    ```

@ConnorMcMahon @cgillum

ConnorMcMahon:
When we throw errors in the JS layer, we need some way to embed the orchestration replay state in the errors so that the C# extension can replay behavior. This is necessary so that the Durable Task does not throw an non-determinism exception.

This also requires a fix on the C# extension to be able to read this embedded data from the RPC exception. Here is the PR for the fix in Durable 1.x: https://github.com/Azure/azure-functions-durable-extension/pull/1171. 

Resolve #125


antempus:
@ConnorMcMahon Awesome 👏 

> To avoid this, we mark the JS execution as successful and put the burden on the C# shim for out-of-proc languages to surface these
failures.

Not knowing how C# shim works under the hood yet, could we have a situation where the shim sees a failure and mistakenly treats it as a successful execution to continue processing?

A cursory review of the changes seems to indicate that's not a possibility, but just a curiosity.

ConnorMcMahon:
@antempus 

At least in the latest version of both [Durable 1.x](https://github.com/Azure/azure-functions-durable-extension/blob/v1.8.3/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs) and [Durable 2.x](https://github.com/Azure/azure-functions-durable-extension/blob/v2.1.1/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs#L70), that should not be the case. 

Given that the code to correctly handle this has been in the C# out-of-proc logic from the beginning,  this was a misunderstanding in how the JS-to-C# layer transfers data in the function failure case. I have now fixed the Durable JS tests so that they accurately reflect the data contract that C# expects.

That being said, it would be a good idea to add some tests to the C# side of things to make sure it does not regress it's side of the contract.

package-lock.json
```diff
@@ -24,7 +24,7 @@
       "integrity": "sha512-ZAR2bPHOl4Xg6eklUGpsdiIJ4+J1SNag1DHHrG/73Uz/nVwXqjgUtRPLoS+aVyieN9cSbc0E4LsU984tWcDyNg==",
       "dev": true,
       "requires": {
-        "@sinonjs/samsam": "^2 || ^3"
+        "@sinonjs/samsam": "2.1.3"
```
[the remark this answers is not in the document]
anthonychu: I think npm install tries to install the latest dependencies that satisfies the versions in package.json, so that causes the lock file to change. AFAIK this is normal.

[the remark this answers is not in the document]
cgillum: In that case, why were the previous changes made (the one with `^2 || ^3"`, etc.)? Or should those types of changes only be made in package.json?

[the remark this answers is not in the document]
ConnorMcMahon: I recently upgraded from npm 5 to npm 6, so it is possible that they changed the default behavior.

src/orchestrator.ts
```diff
@@ -186,17 +187,36 @@ export class Orchestrator {
...
-                    customStatus: this.customStatus,
-                }),
-            );
-            return;
+            if (error instanceof DurableFailure) {
+                // We will return the function execution as a success. It is up to the C# shim
+                // to correctly process the error only after reprocessing all of the actions to
+                // prevent a non-determinism error.
+                context.done(
+                    undefined,
```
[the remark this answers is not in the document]
ConnorMcMahon: Unfortunately, I do believe this would result in a monitoring difference. I would have to double check with @mhoeger about the implications of this, and if there is any other workaround.

Right now, the problem is that failing the function execution means we get an RPC exception, with no useful information to perform the replay on the Durable Extension side of things. 

One potential workaround is to re-execute the orchestration function one more time from the C# extension if we detect a failure. This time we could indicate that the execution should fail when it encounters a `DurableFailure`. Note that this would also result in a monitoring difference between C# and JS, but at least customers would see an orchestration failure in App Insights. It also has a slight cost implication, due to the extra function execution.

package-lock.json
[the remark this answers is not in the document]
anthonychu: I'm on npm 6.13.4. I think different npm versions generate the file slightly differently and that it's okay. But in CI we should be running `npm ci` so it installs deps using the lock file in that commit.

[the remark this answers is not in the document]
ConnorMcMahon: So I ran npm install again and it undid all of the changes... I'm not exactly sure what happened here.

src/orchestrator.ts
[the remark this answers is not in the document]
ConnorMcMahon: So I think I found a workaround. The RPC exception actually does have access to the Error returned (at least the message and the stacktrace). I could theoretically embed the `OrchestratorState` object in the error message, and then the C# shim could parse out the object from the exception. The downside is that this would require customers to upgrade both their DurableJS version as well as their extension version to get the fixed behavior.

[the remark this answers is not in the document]
ConnorMcMahon: I should have addressed this in my latest iteration.

src/outofprocerrorwrapper.ts
```diff
@@ -0,0 +1,8 @@
+import { OrchestratorState } from "./classes";
+
+export class OutOfProcErrorWrapper extends Error {
+    constructor(state: OrchestratorState) {
+        const message = JSON.stringify(state);
```
ConnorMcMahon: I should probably improve the Error, as this message and stacktrace need to serve two purposes:

1. Embed the OrchestratorState data for the C# extension to read. This data should be easy to grab with regex or other string parsing tools in a unique way, so that custom Errors thrown by customer code will not interfere with extracting this data.
2. Be human readable with clarity of how to resolve the issue, as this is the exception that will be logged in AppInsights.

Proposed error string:
```
Message: <Original-Error-Message>

$OrchestratorOutOfProcState$<Orchestrator-State-Json>$OrchestratorOutOfProcState$
StackTrace: <Original-Error-StackTrace>
```

src/durablefailure.ts
```diff
@@ -0,0 +1,5 @@
+export class DurableFailure extends Error {
```
cgillum: Would the name `DurableError` make more sense this this class derives from `Error`? Just like how in .NET exception types should end with `Exception` I wonder if we should be doing the same for custom error types in TypeScript. It would make the usage of this type more immediately obvious to other TypeScript/JavaScript developers.

src/outofprocerrorwrapper.ts
```diff
@@ -0,0 +1,11 @@
+import { OrchestratorState } from "./classes";
+
+const outOfProcDataLabel = "\n\n$OutOfProcData$:";
+
+export class OutOfProcErrorWrapper extends Error {
```
cgillum: More naming nits: 
* I suggest the suffix `Result` rather than `Wrapper` to make it clear that this data is used as the result of a function execution.
* I suggest we avoid the term "out-of-proc" since it has no meaning from the point of view of a JavaScript app (which does not know or care that it is running "out-of-proc" - only the extension cares about this).

Perhaps a name like `DurableErrorResult` would be more appropriate?

src/outofprocerrorwrapper.ts
```diff
@@ -0,0 +1,11 @@
...
+    constructor(error: Error, state: OrchestratorState) {
+        const message = `${error.message}${outOfProcDataLabel}${JSON.stringify(state)}`;
```
cgillum: I suggest you add a comment explaining that this is an predefined contract understood by the Durable extension, and that changing this could be a breaking change.

src/orchestrator.ts
```diff
@@ -208,7 +209,7 @@ export class Orchestrator {
         const taskFailed = this.findTaskFailed(state, taskScheduled);
         this.setProcessed([taskScheduled, taskCompleted, taskFailed]);
 
-        if (taskCompleted) {
+        if (taskCompleted !== undefined) {
```
cgillum: Why make the distinction between falsey values? Do we intend to treat `null` differently from `undefined`? I'm concerned that explicit checks like these could lead to subtle bugs if a `null` slips through somewhere and we don't correctly detect it.

src/outofprocerrorwrapper.ts
ConnorMcMahon: I am hesitant to call it `DurableErrorResult`, as that implies this only applies to `DurableError`s, when it applies to all errors, including those thrown by the customer.

The name I am leaning towards is `OrchestrationFailureError`. The only error that should be thrown by `orchestrator.handle()` should be this type, and I feel this name captures that.

src/outofprocerrorwrapper.ts
ConnorMcMahon: Adding this as a class level comment.

src/orchestrator.ts
ConnorMcMahon: I think I was misinterpreting the advice given [here](https://basarat.gitbook.io/typescript/recap/null-undefined#final-thoughts), regarding not using `null`.

That being said, the same [section](https://basarat.gitbook.io/typescript/recap/null-undefined#checking-for-either) literally says to use `!= null`, as that properly checks for both, and says that you don't want to make a distinction between the two. 

I'm not sure if it is better condition to do general falsy checks like what existed before, or just do these `!= null` checks to only check for `null` or `undefined`. Knowing JS and the general appeal to brevity, I am presuming that the falsy check we used to have is preferred, so I will revert my change.

src/durableerror.ts
```diff
@@ -0,0 +1,5 @@
+export class DurableError extends Error {
```
cgillum: What is the purpose of this new error class? I suggest we add a comment about why it exists since it doesn't actually add any state or functionality to the base `Error` class.

ConnorMcMahon: It was just to differentiate errors from the "Durable" framework (i.e. a scheduled activity or suborchestration) failed vs any other exception. Theoretically, customers can then make their catch logic more specific by doing an `instanceof` check. If you don't think it's worth adding, I am fine removing it.

cgillum: I’m fine with keeping it. I just think it would be worth adding a comment in the code about it.

Issue #150, opened 2020-01-23
Axios errors are re-thrown as strings


cgillum:
### Problem
In the `DurableOrchestrationClient` class, all axios errors are caught and rethrown as strings. This is not good practice in JavaScript and can make debugging harder.

I discovered this when using **nock** to mock HTTP requests (the library was helpful enough to point out this problem to me).

### Proposal
Instead of catching axios errors, just let the error be raised as-is to the caller. Alternatively we could wrap the error in an error of our own, but I'm not sure there's much value added there.

anthonychu:
May be related to https://github.com/Azure/azure-functions-durable-js/pull/129#issuecomment-572300429. I agree we should throw objects, but ensure they are serialized properly when writing to logs or propagating the errors to other functions.

I think some Axios objects have circular references that need special handling when serializing.

Issue #90, opened 2019-06-10
client.startNew function throws intermittent error when function app runs in Azure


harshithkashyap:
**Describe the bug**
When calling the orchestration function in Azure using `client.startNew(...)` method, durable functions throws unknown error with no error message in the exception. From the stack trace and error message in App Insights, the following message is logged.

`Exception while executing function: Functions.CL_skill-execution-approval-change <--- Result: Failure Exception: Error Stack: Error at DurableOrchestrationClient.<anonymous> (D:\home\site\wwwroot\node_modules\durable-functions\lib\src\durableorchestrationclient.js:284:43) at Generator.next (<anonymous>) at fulfilled (D:\home\site\wwwroot\node_modules\durable-functions\lib\src\durableorchestrationclient.js:4:58) at process._tickCallback (internal/process/next_tick.js:68:7)`

**Investigative information**

- Durable Functions extension version: 1.8.2
- durable-functions npm module version: 1.2.2
- Language (JavaScript/TypeScript) and version: Typescript and 3.5.1
- Node.js version: 10.15.2

***If deployed to Azure App Service***

- Timeframe issue observed:
- Invocation ID: 49cc1fa2-3a28-471b-afae-910a69e2797a
- Region: East US
- Timestamp: 2019-06-10 18:00:16.099
- Orchestration instance ID(s): Instance ID isn't generated and throws error when starting the instance.

> If you don't want to share your Function App name or Functions names on GitHub, please be sure to provide your Invocation ID, Timestamp, and Region - we can use this to look up your Function App/Function. Provide an invocation id per Function. See the [Functions Host wiki](https://github.com/Azure/azure-webjobs-sdk-script/wiki/Sharing-Your-Function-App-name-privately) for more details.

**To Reproduce**

Unfortunately, I have not been able to reproduce this issue locally and this is occurring only in the deployed azure function app.

> While not required, providing your orchestrator's source code in anonymized form is often very helpful when investigating unexpected orchestrator behavior.

```typescript
import { AzureFunction, Context } from '@azure/functions';
import { getClient } from 'durable-functions';
import { SkillExecutionQueueMessage } from '../interfaces/skills';

const queueTrigger: AzureFunction = async (context: Context, queueMessage: SkillExecutionQueueMessage) => {
    try {
        const client = getClient(context);
        const instanceId = await client.startNew('OC_skill-approval-change', undefined, queueMessage);

[...]

Issue #86, opened 2019-05-30
Orchestrator functions error out when run during Function App service warmup period


amclin:
When using the Consumption Tier of Azure Function Apps, the service is put into an idle state. The next request to the service "wakes up" the service, and may take a while to respond (which is expected).

When the service is "hot", then everything works fine, however, if the wakeup request calls an Orchestrator function, then the Orchestrator throws a Service Unavailable error:

```
Exception while executing function: Functions.HttpTrigger Result: Failure
Exception: Error: The service is unavailable.
Stack: Error: The service is unavailable.
    at DurableOrchestrationClient.<anonymous> (D:\home\site\wwwroot\node_modules\durable-functions\lib\src\durableorchestrationclient.js:247:43)
    at Generator.next (<anonymous>)
    at fulfilled (D:\home\site\wwwroot\node_modules\durable-functions\lib\src\durableorchestrationclient.js:4:58)
    at process._tickCallback (internal/process/next_tick.js:68:7) 
```

Stack Trace:
```
Microsoft.Azure.WebJobs.Host.FunctionInvocationException:
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor+<ExecuteWithLoggingAsync>d__19.MoveNext (Microsoft.Azure.WebJobs.Host, Version=3.0.8.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35Microsoft.Azure.WebJobs.Host, Version=3.0.8.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35: C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.csMicrosoft.Azure.WebJobs.Host, Version=3.0.8.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35: 321)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
[...]

Issue #62, opened 2019-03-29
Azure function apps authentication bug


KieranDaviesV:
**Describe the bug**
When durable function is up in Function apps, using the function apps authentication, i receive a 500 error status. along with:

  Exception: Error: You do not have permission to view this directory or page.
Stack: Error: You do not have permission to view this directory or page.
    at DurableOrchestrationClient.<anonymous> (D:\home\site\wwwroot\node_modules\durable-functions\lib\src\durableorchestrationclient.js:247:43)
    at Generator.next (<anonymous>)
    at fulfilled (D:\home\site\wwwroot\node_modules\durable-functions\lib\src\durableorchestrationclient.js:4:58)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)

**Investigative information**

- Durable Functions extension version: 1.7.1
- Function App version (1.0 or 2.0): 2.0
- Programming language used: Javascript

***If JavaScript***

- durable-functions npm module version: 1.1.4

***If deployed to Azure***

- Timeframe issue observed:  2:51:24 pm gmt
- Function App name: sitecreatorv2
- Function name(s): httpRequest
- Region: North Europe
- Orchestration instance ID(s): -- didn't get that far

> If you don't want to share your Function App name or Functions names on GitHub, please be sure to provide your Invocation ID, Timestamp, and Region - we can use this to look up your Function App/Function. Provide an invocation id per Function. See the [Functions Host wiki](https://github.com/Azure/azure-webjobs-sdk-script/wiki/Sharing-Your-Function-App-name-privately) for more details.

**To Reproduce**

1. fire http request
2. view logs to see error

> While not required, providing your orchestrator's source code in anonymized form is often very helpful when investigating unexpected orchestrator behavior.

**Expected behavior**
Should fire off my orchestrators

**Actual behavior**
Errors out with You do not have permission to view this directory or page.

**Screenshots**

**Known workarounds**
Provide a description of any known workarounds you used.

**Additional context**

- Development environment (ex. Visual Studio)
- Links to source
- Additional bindings used
- Function invocation IDs

cgillum:
Unfortunately this is a limitation of the JavaScript orchestration client. It schedules orchestrations by making an HTTP call to the current function app, but if access to that function app is protected by Authentication / Authorization, then these client calls will fail.

In a future update, we hope to resolve issues like this by removing the HTTP call and having the client internally signal the durable extension to schedule orchestrations.

[...]

PR #152, opened 2020-01-24
Update DurableClient to support local RPC

cgillum:
As part of [this Durable Extension PR](https://github.com/Azure/azure-functions-durable-extension/pull/1166), we're exposting a local RPC endpoint so that the Durable Client binding can make local RPC calls into the extension rather than making HTTP requests that route through the FE.

Fixes https://github.com/Azure/azure-functions-durable-js/issues/62
Fixes https://github.com/Azure/azure-functions-durable-js/issues/86
Fixes https://github.com/Azure/azure-functions-durable-js/issues/90
Fixes https://github.com/Azure/azure-functions-durable-js/issues/150

Some important notes:
* This change is intended to be 100% backwards compatible with previous extension versions.
* I created a new test file, `durableclient-spec.ts`, for testing the new code paths.
* The existing `durableorchestrationclient-spec.ts` will remain for validating the back-compat behavior.

Remaining work:
- [x] Implement all client APIs
- [x] End-to-end validation in func.exe
- [x] End-to-end validation in Azure
- [x] Full unit test coverage
- [x] Fix linting warnings
- [x] End to end validation of _all APIs_ in Azure with Easy Auth enabled

### DIFF
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -186,28 +186,24 @@ export class DurableOrchestrationClient {
         showHistory?: boolean,
         showHistoryOutput?: boolean,
         showInput?: boolean,
-        ): Promise<DurableOrchestrationStatus> {
-        try {
-            const options: GetStatusOptions = {
-                instanceId,
-                showHistory,
-                showHistoryOutput,
-                showInput,
-            };
-            const response = await this.getStatusInternal(options);
+    ): Promise<DurableOrchestrationStatus> {
+        const options: GetStatusOptions = {
+            instanceId,
+            showHistory,
+            showHistoryOutput,
+            showInput,
+        };
+        const response = await this.getStatusInternal(options);
 
-            switch (response.status) {
-                case 200: // instance completed
-                case 202: // instance in progress
-                case 400: // instance failed or terminated
-                case 404: // instance not found or pending
-                case 500: // instance failed with unhandled exception
-                    return response.data as DurableOrchestrationStatus;
-                default:
-                    return Promise.reject(new Error(`Webhook returned unrecognized status code ${response.status}`));
-            }
-        } catch (error) {   // error object is axios-specific, not a JavaScript Error; extract relevant bit
-            throw error.message;
+        switch (response.status) {
+            case 200: // instance completed
+            case 202: // instance in progress
+            case 400: // instance failed or terminated
+            case 404: // instance not found or pending
+            case 500: // instance failed with unhandled exception
+                return response.data as DurableOrchestrationStatus;
+            default:
+                return Promise.reject(new Error(`Webhook returned unrecognized status code ${response.status}`));
         }
     }
 
@@ -215,12 +211,8 @@ export class DurableOrchestrationClient {
      * Gets the status of all orchestration instances.
      */
     public async getStatusAll(): Promise<DurableOrchestrationStatus[]> {
-        try {
-            const response = await this.getStatusInternal({});
-            return response.data as DurableOrchestrationStatus[];
-        } catch (error) {   // error object is axios-specific, not a JavaScript Error; extract relevant bit
-            throw error.message;
-        }
+        const response = await this.getStatusInternal({});
+        return response.data as DurableOrchestrationStatus[];
     }
 
     /**
@@ -237,22 +229,18 @@ export class DurableOrchestrationClient {
         createdTimeFrom: Date | undefined,
         createdTimeTo: Date | undefined,
         runtimeStatus: OrchestrationRuntimeStatus[],
-        ): Promise<DurableOrchestrationStatus[]> {
-        try {
-            const options: GetStatusOptions = {
-                createdTimeFrom,
-                createdTimeTo,
-                runtimeStatus,
-            };
-            const response = await this.getStatusInternal(options);
+    ): Promise<DurableOrchestrationStatus[]> {
+        const options: GetStatusOptions = {
+            createdTimeFrom,
+            createdTimeTo,
+            runtimeStatus,
+        };
+        const response = await this.getStatusInternal(options);
 
-            if (response.status > 202) {
-                return Promise.reject(new Error(`Webhook returned status code ${response.status}: ${response.data}`));
-            } else {
-                return response.data as DurableOrchestrationStatus[];
-            }
-        } catch (error) {   // error object is axios-specific, not a JavaScript Error; extract relevant bit
-            throw error.message;
+        if (response.status > 202) {
+            return Promise.reject(new Error(`Webhook returned status code ${response.status}: ${response.data}`));
+        } else {
+            return response.data as DurableOrchestrationStatus[];
         }
     }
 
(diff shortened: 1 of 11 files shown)


src/durableorchestrationclient.ts
```diff
@@ -362,39 +341,55 @@ export class DurableOrchestrationClient {
...
         }
 
-        const idPlaceholder = this.clientData.managementUrls.id;
-        let requestUrl = this.clientData.managementUrls.sendEventPostUri
-            .replace(idPlaceholder, instanceId)
-            .replace(this.eventNamePlaceholder, eventName);
+        let requestUrl: string;
+        if (this.clientData.rpcBaseUrl) {
+            // Fast local RPC path
+            let path = `instances/${instanceId}/raiseEvent/${eventName}`;
```
cgillum: I'm intentionally hardcoding the URL paths here rather than doing templating or variable replacement. I just didn't feel the potential benefit was worth the added complexity. Basically, I'm considering the HTTP API to be a well-known contract and I would expect other language implementations to follow this model as well.

ConnorMcMahon: I'm 100% OK with this. I tried to move closer to this direction with the new Entity APIs, because the maintenance cost of adding a new template everytime we wanted to support a new API seemed not worth the cost.

In fact, in order to keep our legacy/local HTTP logic as close as possible, I would recommend extracting the base url and authorization key from one of the existing templates. Then the path building and query string building logic in the two code paths can be almost entirely shared, and we don't have to maintain any legacy templating logic.

src/orchestrationfailureerror.ts
```diff
@@ -6,17 +6,17 @@ const outOfProcDataLabel = "\n\n$OutOfProcData$:";
...
  */
 export class OrchestrationFailureError extends Error {
     constructor(error: any, state: OrchestratorState) {
-        let errorMessage : String;
+        let errorMessage: string;
         if (error instanceof Error) {
-            errorMessage= error.message;
-        } else if (error instanceof String) {
+            errorMessage = error.message;
+        } else if (typeof(error) === "string") {
```
ConnorMcMahon: Good catch here!

src/durableorchestrationclient.ts
```diff
@@ -294,45 +285,60 @@ export class DurableOrchestrationClient {
...
-        const idPlaceholder = this.clientData.managementUrls.id;
-        let requestUrl = this.clientData.managementUrls.statusQueryGetUri
-        .replace(idPlaceholder, "");
+    ): Promise<PurgeHistoryResult> {
+        let requestUrl: string;
+        if (this.clientData.rpcBaseUrl) {
+            // Fast local RPC path
+            let path = new URL("instances/", this.clientData.rpcBaseUrl).href;
+            const query: string[] = [];
+            if (createdTimeFrom) { query.push(`createdTimeFrom=${createdTimeFrom.toISOString()}`); }
```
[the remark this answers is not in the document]
cgillum: The main reason I added this new logic is because it's simpler, more efficient, and more readable. I didn't want to change the existing logic because it is legacy, should get removed eventually, and I didn't want to risk any regressions.

src/durableorchestrationclient.ts
cgillum: Do you think it's really necessary for us to make any changes to the legacy code? Per my other comment, I didn't see any need to maintain it at all, which is why I opted to leave it untouched.

src/durableorchestrationclient.ts
```diff
@@ -362,39 +368,51 @@ export class DurableOrchestrationClient {
...
+
+        const response = await this.axiosInstance.post(requestUrl, JSON.stringify(eventData));
+        switch (response.status) {
+            case 202: // event accepted
+            case 410: // instance completed or failed
+                return;
+            case 404:
+                return Promise.reject(new Error(`No instance with ID '${instanceId}' found.`));
+            case 400:
+                return Promise.reject(new Error("Only application/json request content is supported"));
```
ConnorMcMahon: I know this isn't new code, but technically a 400 can also indicate invalid JSON content. Why don't we just make the Error message match the body of the response?

src/durableorchestrationclient.ts
```diff
@@ -410,29 +428,42 @@ export class DurableOrchestrationClient {
...
+        } else {
+            // Legacy app frontend path
+            if (!(this.clientData.baseUrl && this.clientData.requiredQueryStringParameters)) {
+                throw new Error("Cannot use the readEntityState API with this version of the Durable Task Extension.");
             }
-        } catch (error) {   // error object is axios-specific, not a JavaScript Error; extract relevant bit
-            throw error.message;
+
+            requestUrl = WebhookUtils.getReadEntityUrl(
+                this.clientData.rpcBaseUrl || this.clientData.baseUrl,
```
ConnorMcMahon: We already know this.clientData.rpcBaseUrl is falsy, so I think we should remove this check.

src/durableorchestrationclient.ts
```diff
@@ -481,29 +525,43 @@ export class DurableOrchestrationClient {
...
+        } else {
+            // Legacy app frontend path
+            if (!(this.clientData.baseUrl && this.clientData.requiredQueryStringParameters)) {
+                throw new Error("Cannot use the signalEntity API with this version of the Durable Task Extension.");
             }
-        } catch (error) {   // error object is axios-specific, not a JavaScript Error; extract relevant bit
-            throw error.message;
+
+            requestUrl = WebhookUtils.getSignalEntityUrl(
+                this.clientData.rpcBaseUrl || this.clientData.baseUrl,
```
ConnorMcMahon:
```suggestion
                this.clientData.baseUrl,
```

test/unit/durableclient-spec.ts
```diff
@@ -0,0 +1,423 @@
...
+            const functionName = "MyFunction";
+            const expectedUrl = new URL(`${testRpcOrigin}/durabletask/orchestrators/${functionName}`);
+            const expectedHeaders = { reqheaders: { "Content-Type": "application/json" } };
+
+            const scope = nock(expectedUrl.origin, expectedHeaders)
+                .post(expectedUrl.pathname)
+                .reply(202, { id: "abc123" });
+
+            const result = await client.startNew(functionName);
+            expect(scope.isDone()).to.be.equal(true);
```
ConnorMcMahon: Is this similar to verify() in C#?

src/entities/entityid.ts
```diff
@@ -24,6 +24,7 @@ export class EntityId {
      */
     constructor(
         // TODO: consider how to name these fields more accurately without interfering with JSON serialization
+        // TODO: entity keys are supposed to be optional
```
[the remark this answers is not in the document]
cgillum: Actually this isn't completely true. We disallow null entity keys. I was thinking of the case where we allow entity keys to be the empty string. I'll remove this comment.

src/durableorchestrationclient.ts
cgillum: I'm not confident about how to reliably transform the response content into a string that can be included in the error message, so I don't feel super comfortable making that change as part of this PR. I'm okay with updating the error message, though.

ConnorMcMahon: Ah, I didn't realize we embed the message as a JSON object. I would say we can just punt that for now.

src/durableorchestrationclient.ts
cgillum: Good catch. I think this is leftover from an earlier attempt to unify the two paths (which I've since abandoned to minimize risk of regression).

src/durableorchestrationclient.ts
cgillum: Reverted

test/unit/durableclient-spec.ts
cgillum: Not sure. I just copied this from other existing nock test code.

ConnorMcMahon: It [looks like it is.](https://www.npmjs.com/package/nock#isdone). If it didn't, I would have recommended we did something to ensure the mock method was actually called, but it does!

cgillum:
I verified the following in an Azure Consumption plan:

* RPC endpoint works as expected for all APIs with Easy Auth enabled or disabled
* I can disable the RPC endpoint by setting `"localRpcEndpointEnabled": false` in host.json
* When RPC endpoint is disabled, client calls fail with Easy Auth enabled (expected) but succeed with Easy Auth disabled (also expected), which verifies we revert to the old behavior correctly.

I think this PR is now fully verified! Merging...

PR #156, opened 2020-02-04
Allow task sets in df.task.all and df.task.any

ConnorMcMahon:
Unify the idea of `Task` and `TaskSet` so that `df.task.any()` and `df.task.all()` can work on the tasksets they generate themselves.

This cleaning of the two concepts also simplifies the code as well.

### DIFF
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -550,6 +550,7 @@ export class Orchestrator {
 
             const failedTasks = completedTasks.filter(TaskFilter.isFailedTask);
             if (failedTasks.length > 0) {
+                // TODO: aggregate all failures into one clean error.
                 return TaskFactory.FailedTaskSet(tasks, completionIndex, (failedTasks[0] as FailedTask).exception);
             } else {
                 const results = completedTasks.map((task) => task.result);
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -11,8 +11,7 @@ import { CallActivityAction, CallActivityWithRetryAction, CallEntityAction, Call
 } from "./classes";
 import { DurableError } from "./durableerror";
 import { OrchestrationFailureError } from "./orchestrationfailureerror";
-import { FailedSingleTask } from "./tasks/taskinterfaces";
-import { Yieldable } from "./tasks/yieldable";
+import { FailedTask, TaskBase } from "./tasks/taskinterfaces";
 import { TokenSource } from "./tokensource";
 
 /** @hidden */
@@ -83,7 +82,7 @@ export class Orchestrator {
         // Setup
         const gen = this.fn(context);
         const actions: IAction[][] = [];
-        let partialResult: Task | TaskSet;
+        let partialResult: TaskBase;
 
         try {
             // First execution, we have not yet "yielded" any of the tasks.
@@ -113,7 +112,11 @@ export class Orchestrator {
                     }
                 }
 
-                partialResult = g.value as Task | TaskSet;
+                partialResult = g.value as TaskBase;
+                let newActions = partialResult.yield();
+                if (newActions && newActions.length > 0) {
+                    actions.push(newActions);
+                }
 
                 // Return continue as new events as completed, as the execution itself is now completed.
                 if (TaskFilter.isSingleTask(partialResult) && partialResult.action instanceof ContinueAsNewAction) {
@@ -539,28 +542,27 @@ export class Orchestrator {
         }
     }
 
-    private all(state: HistoryEvent[], tasks: Task[]): TaskSet {
-        const faulted = tasks.filter(TaskFilter.isFailedTask);
-        if (faulted.length > 0) {
-            return TaskFactory.FailedTaskSet(tasks, (faulted[0] as FailedSingleTask).exception);
-        }
+    private all(state: HistoryEvent[], tasks: TaskBase[]): TaskSet {
+        const completedTasks = tasks.filter(TaskFilter.isCompletedTask);
+        if (completedTasks.length === tasks.length) {
+            const sortedTasks = completedTasks.sort(TaskFilter.CompareFinishedTime) 
+            const completionIndex = sortedTasks[tasks.length-1].completionIndex;
 
-        const succesfulTasks = tasks.filter(TaskFilter.isSuccessfulSingleTask);
-        const isCompleted = succesfulTasks.length === tasks.length;
-        if (isCompleted) {
-            const results = succesfulTasks.reduce((acc, t) => {
-                return [...acc, t.result];
-            }, []);
-
-            return TaskFactory.SuccessfulTaskSet(tasks, results);
+            const failedTasks = completedTasks.filter(TaskFilter.isFailedTask);
+            if (failedTasks.length > 0) {
+                return TaskFactory.FailedTaskSet(tasks, completionIndex, (failedTasks[0] as FailedTask).exception);
+            } else {
+                const results = completedTasks.map((task) => task.result);
+                return TaskFactory.SuccessfulTaskSet(tasks, completionIndex, results);
+            }
         } else {
             return TaskFactory.UncompletedTaskSet(tasks);
         }
     }
 
-    private any(state: HistoryEvent[], tasks: Task[]): TaskSet {
+    private any(state: HistoryEvent[], tasks: TaskBase[]): TaskSet {
         const completedTasks = tasks
-            .filter(TaskFilter.isSuccessfulSingleTask)
+            .filter(TaskFilter.isCompletedTask)
             .sort((a, b) => {
                 if (a.completionIndex > b.completionIndex) { return 1; }
                 if (a.completionIndex < b.completionIndex) { return -1; }
@@ -569,7 +571,7 @@ export class Orchestrator {
 
         const firstCompleted = completedTasks[0];
         if (firstCompleted) {
-            return TaskFactory.SuccessfulTaskSet(tasks, firstCompleted);
+            return TaskFactory.SuccessfulTaskSet(tasks, firstCompleted.completionIndex, firstCompleted);
         } else {
             return TaskFactory.UncompletedTaskSet(tasks);
         }
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -12,7 +12,7 @@ import { CallActivityAction, CallActivityWithRetryAction, CallEntityAction, Call
 } from "./classes";
 import { DurableError } from "./durableerror";
 import { OrchestrationFailureError } from "./orchestrationfailureerror";
-import { TaskBase } from "./tasks/taskinterfaces";
+import { CompletedTask, TaskBase } from "./tasks/taskinterfaces";
 import { TokenSource } from "./tokensource";
 
 /** @hidden */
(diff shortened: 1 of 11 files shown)


cgillum:
Just an observation - no action necessary: I feel that the overall handling of tasks is unnecessarily complicated with our use of timestamps and completion indexes. At a high level, what we need to do in orchestrator.ts is similar to what [TaskOrchestrationExecutor](https://github.com/Azure/durabletask/blob/master/src/DurableTask.Core/TaskOrchestrationExecutor.cs) in the Durable Task Framework is doing, and that implementation is much simpler and more deterministic, IMO.

ConnorMcMahon:
@cgillum 

I'm not opposed to refactoring an approach using a mapping of ids to tasks, and preprocessing the tasks. That being said, I don't think it gives us that much in this scenario. A lot of the complexity in `TaskOrchestrationExecutor` is hidden in `TaskOrchestrationContext`. It gets even more complex considering how closely `TaskOrchestrationContext` depends on the first-class support for `TaskCompletionSource`, an advantage we don't really have in JavaScript.

I also don't think we have any real way of implementing task.any() without keeping track of completion indices (or some other similar metadata), considering we don't have full Task support in JS.

src/orchestrator.ts
```diff
@@ -550,28 +542,28 @@ export class Orchestrator {
...
+    private all(state: HistoryEvent[], tasks: TaskBase[]): TaskSet {
+        const completedTasks = tasks.filter(TaskFilter.isCompletedTask);
+        if (completedTasks.length === tasks.length) {
+            const sortedTasks = completedTasks.sort(TaskFilter.CompareFinishedTime);
+            const completionIndex = sortedTasks[tasks.length - 1].completionIndex;
 
-            return TaskFactory.SuccessfulTaskSet(tasks, results);
+            const failedTasks = completedTasks.filter(TaskFilter.isFailedTask);
+            if (failedTasks.length > 0) {
+                // TODO: aggregate all failures into one clean error.
```
cgillum: Will this `TODO` be done as part of this PR or in another PR?

Also, is this `TODO` talking about the next line where we only consider the first exception and not any subsequent exceptions?

src/tasks/taskset.ts
```diff
@@ -1,12 +1,21 @@
...
+        private readonly tasks: TaskBase[],
+        private readonly completionIndex?: number,
         public result?: unknown,
         public exception?: Error,
     ) { }
+
+    public yieldNewActions(): IAction[] {
+        // Get all of the actions in subtasks and flatten into one array.
+        return this.tasks.map((task) => task.yieldNewActions())
+            .reduce((acc, arr) => acc.concat(arr));
```
cgillum: I can't tell what type of objects `acc` and `arr` are. Can we use more descriptive variable names for those?

src/orchestrator.ts
```diff
@@ -550,28 +542,28 @@ export class Orchestrator {
...
-        const isCompleted = succesfulTasks.length === tasks.length;
-        if (isCompleted) {
-            const results = succesfulTasks.reduce((acc, t) => {
-                return [...acc, t.result];
-            }, []);
+    private all(state: HistoryEvent[], tasks: TaskBase[]): TaskSet {
+        const completedTasks = tasks.filter(TaskFilter.isCompletedTask);
+        if (completedTasks.length === tasks.length) {
+            const sortedTasks = completedTasks.sort(TaskFilter.CompareFinishedTime);
+            const completionIndex = sortedTasks[tasks.length - 1].completionIndex;
```
cgillum: Is it okay that the resulting `TaskSet` has a `completionIndex` that is the same as one of the completed `TaskBase` objects? I would have expected there to be a `+ 1` somewhere in here. Does the relative order of `TaskBase` and `TaskSet` not matter?

src/orchestrator.ts
```diff
@@ -550,28 +542,28 @@ export class Orchestrator {
...
 
-            return TaskFactory.SuccessfulTaskSet(tasks, results);
+            const failedTasks = completedTasks.filter(TaskFilter.isFailedTask);
```
cgillum: Should we use `sortedTasks` here instead of `completedTasks`, or does it not matter?

src/orchestrator.ts
```diff
@@ -550,28 +542,28 @@ export class Orchestrator {
...
+            if (failedTasks.length > 0) {
+                // TODO: aggregate all failures into one clean error.
+                return TaskFactory.FailedTaskSet(tasks, completionIndex, (failedTasks[0] as FailedTask).exception);
```
cgillum: I wonder if it would be safer from a refactoring perspective to use `completedTasks` or `sortedTasks` here instead of `tasks`. Same with line 557.

src/orchestrator.ts
```diff
@@ -550,28 +542,28 @@ export class Orchestrator {
...
             return TaskFactory.UncompletedTaskSet(tasks);
         }
     }
 
-    private any(state: HistoryEvent[], tasks: Task[]): TaskSet {
+    private any(state: HistoryEvent[], tasks: TaskBase[]): TaskSet {
         const completedTasks = tasks
-            .filter(TaskFilter.isSuccessfulSingleTask)
+            .filter(TaskFilter.isCompletedTask)
             .sort((a, b) => {
```
cgillum: It looks like here and also in the `all` method we're sorting just to find the min or max element. If we don't actually need a sorted list, it would be more CPU efficient to just get the min or max of the list. It looks like this should be fairly straightforward to do in ES6: https://stackoverflow.com/a/59511498/2069

test/integration/orchestrator-spec.ts
```diff
@@ -1809,6 +1809,7 @@ describe("Orchestrator", () => {
         it("Task.any proceeds if a scheduled parallel task completes in order", async () => {
             const orchestrator = TestOrchestrations.AnyAOrB;
             const completeInOrder = true;
+            const initialDate = moment.utc().toDate();
```
cgillum: Unused?

test/testobjects/TestOrchestrations.ts
```diff
@@ -45,6 +45,29 @@ export class TestOrchestrations {
...
 
+    public static AnyWithTaskSet: any = df.orchestrator(function*(context: any) {
+        const now = new Date(context.df.currentUtcDateTime).getTime();
+        const deadline = new Date(now + 300000);
+        const timeoutTask = context.df.createTimer(deadline);
+
+        const gate1 = context.df.waitForExternalEvent("A");
+        const gate2 = context.df.waitForExternalEvent("B");
+
+        const winner = yield context.df.Task.any([context.df.Task.all([gate1, gate2]), timeoutTask]);
```
cgillum: I had a hard time understanding the test code (more comments might help :)) but do we have a variation that checks for something like the following?
1. `gate1` completes
2. `timeoutTask` completes
3. `timeoutTask` is the winner

I just want to make sure we're verifying that `all` does not complete early if only a subset of tasks are complete.

src/orchestrator.ts
ConnorMcMahon: I can do it as a part of this PR, just so it isn't forgotten about.

And yes, it is referring to next line. In C#, `Task.WhenAll()` accumulates all exceptions into an `AggregateException`, so I will have to do something similar here. Not sure how that would really really surface App Insights in the JS case though.

src/orchestrator.ts
ConnorMcMahon: A TaskSet returned by `all()` is considered finished when its last subtask finishes, so I think it is fair that it would have the same completion index as that subtask.

Imagine the following C#

```csharp

// Tasks will finish in order task1 then task2
Task firstTask =  task1();
Task secondTask  = task2();

Task bothFinishedTask= Task.WhenAll(firstTask, secondTask);

Task lastTask = Task.WhenAny(bothFinishedTask, secondTask);
```

In C#, technically we would probably expect that `lastTask == secondTask`, just due to the latency it takes for `bothFinishedTask` to realize that it is finished after `secondTask` finishes.

With my current implementation, I think I would expect that `lastTask == bothFinishedTask`, just due to the ordering of parameters in `Task.WhenAny()`. I don't think this is a problem, as long as it is deterministic.

In general, I wouldn't expect the above scenario very often regardless. 

If that is not acceptable, I could see doing a ` + .5`, just to make sure that it doesn't interfere with any other task's completion indices.

src/orchestrator.ts
ConnorMcMahon: It really shouldn't matter, at least after I finish the TODO below.

test/testobjects/TestOrchestrations.ts
ConnorMcMahon: Sounds good.

I just copied the example directly from the Durable Python folks, but it could definitely use some more descriptors for clarity.

I'm also willing to make a more robust scenario like you describe. Will just have to add a bit of complexity to the history generation.

We do already verify that `df.Task.all` should not complete early with the test `Task.all does not proceed if some parallel tasks have completed`, but it's possible there is some unique behavior when it is a parameter of `df.Task.any()`

src/tasks/taskfilter.ts
```diff
@@ -1,42 +1,39 @@
...
-        return task.isCompleted !== undefined && task.isFaulted !== undefined;
+    public static CompareFinishedTime(taskA: CompletedTask, taskB: CompletedTask) {
+        if (taskA.completionIndex > taskB.completionIndex) { return 1; }
+        if (taskA.completionIndex < taskB.completionIndex) { return -1; }
+        return 0;
     }
 
-    public static isSingleTask(task: Task | TaskSet): task is Task {
+    public static isYieldable(task: any): task is TaskBase {
+        return (task instanceof Task) || (task instanceof TaskSet);
```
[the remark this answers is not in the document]
ConnorMcMahon: Unfortunately typescript interfaces are compile time only. Maybe a more robust method would be to check for the existence of all of the current properties on `TaskBase`. That way it would also apply to any new implementations of `TaskBase`. The downside is that if I removed a property from `TaskBase`, it would break the method.

test/testobjects/TestOrchestrations.ts
ConnorMcMahon: Actually, looking at the test history, it looks like that scenario is already covered.

If `eventsBeatsTimer` is false in `TestHistories.GetAnyWithTaskSet`, then order of events is

1. "A" completes
2. Timer fires
3. "B" completes.

If it is true, then the order is

1. "A" completes
2. "B" completes"
3. Timer fires.

ConnorMcMahon: Let me know if the improvements helped here.

src/tasks/taskfilter.ts
```diff
@@ -10,7 +10,11 @@ export class TaskFilter {
     }
 
     public static isYieldable(task: any): task is TaskBase {
-        return (task instanceof Task) || (task instanceof TaskSet);
+        const taskBase = task as TaskBase;
+        return taskBase
+            && taskBase.isCompleted != null
+            && taskBase.isFaulted != null
+            && taskBase.yieldNewActions != null;
```
cgillum: Is the use of `null` here correct? I noticed you typically use `undefined` everywhere. Not sure what the expectation is for members that don't exist.

src/orchestrator.ts
```diff
@@ -562,20 +569,22 @@ export class Orchestrator {
...
+        const completedTasks = tasks.filter(TaskFilter.isCompletedTask);
+        if (completedTasks.length === 0) {
             return TaskFactory.UncompletedTaskSet(tasks);
         }
+
+        // Find the completed task with the lowest completion index
+        let firstCompleted = completedTasks[0];
+        completedTasks.forEach((task) =>
+            firstCompleted = task.completionIndex < firstCompleted.completionIndex ? task : firstCompleted,
+        );
```
cgillum: It looks like we're iterating through the list of tasks twice. Should we try to reduce it to one iteration?

```typescript
let firstCompleted: TaskBase = null;
completedTasks.forEach((task) => {
    if (TaskFilter.isCompletedTask(task)) {
        if (!firstCompleted || task.completionIndex < firstCompleted.completionIndex) {
            firstCompleted = task;
        }
    }
});

if (firstCompleted) {
    return TaskFactory.SuccessfulTaskSet(tasks, firstCompleted.completionIndex, firstCompleted);
} else {
    return TaskFactory.UncompletedTaskSet(tasks);
}
```

src/tasks/taskfilter.ts
ConnorMcMahon: `x != null` is shorthand for `x !== undefined && x !== null` due to type coalescing with the `!=` operator. I couldn't just use a falsy check here since `isCompleted` and `isFaulted` are booleans.

I think I technically could have checked just for `!== undefined` if the members don't exist.

cgillum: Good to know!

src/orchestrator.ts
ConnorMcMahon: There is definitely room to reduce iteration through tasks. I actually think task.all() is the worse offender here. I am going to try to optimize both of these, as the fan-out scenario can cause the internal task arrays to get rather large.

PR #165, opened 2020-02-21
Revamp retry logic

ConnorMcMahon:
The current retry logic had two flaws. First, since it restarted searching from the beginning on each attempt, it could grab the wrong scheduled activity/suborchestration on subsequent events. This could lead to incorrect results. Second, since each retry attempt started from the beginning of the state array, it was not very performant. The new approach only iterates through the state array once per call to callActivityWithRetry() or callSubOrchestrationWithRetry().

### DIFF
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -292,47 +292,90 @@ export class Orchestrator {
     ): Task {
         const newAction = new CallActivityWithRetryAction(name, retryOptions, input);
 
-        for (let attempt = 1; attempt <= retryOptions.maxNumberOfAttempts; attempt++) {
-            const taskScheduled = this.findTaskScheduled(state, name);
-            const taskCompleted = this.findTaskCompleted(state, taskScheduled);
-            const taskFailed = this.findTaskFailed(state, taskScheduled);
-            const taskRetryTimer = this.findRetryTimer(state, taskFailed);
-            const taskRetryTimerFired = this.findTimerFired(state, taskRetryTimer);
-            this.setProcessed([
-                taskScheduled,
-                taskCompleted,
-                taskFailed,
-                taskRetryTimer,
-                taskRetryTimerFired,
-            ]);
-
-            if (!taskScheduled) {
-                break;
+        let attempt = 1;
+        let taskScheduled: TaskScheduledEvent | undefined;
+        let taskFailed: TaskFailedEvent | undefined;
+        let taskRetryTimer: TimerCreatedEvent | undefined;
+        for (let i = 0; i < state.length; i++) {
+            const historyEvent = state[i];
+            if (historyEvent.IsProcessed) {
+                continue;
             }
 
-            if (taskCompleted) {
-                const result = this.parseHistoryEvent(taskCompleted);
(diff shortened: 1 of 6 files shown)


src/orchestrator.ts
```diff
@@ -243,31 +243,80 @@ export class Orchestrator {
...
-                    taskFailed.Timestamp,
-                    taskFailed.TaskScheduledId,
-                    state.indexOf(taskFailed),
-                    new DurableError(taskFailed.Reason),
-                );
+        let attempt = 1;
+        let taskScheduled: TaskScheduledEvent | undefined;
+        let taskFailed: TaskFailedEvent | undefined;
+        let taskRetryTimer: TimerCreatedEvent | undefined;
+        for (let i = 0; i < state.length; i++) {
```
davidmrdavid, line 250: nit: can we do a `foreach` / `for ... in`-style loop here?

davidmrdavid: I guess in TS its a `for...of`

src/orchestrator.ts
```diff
@@ -349,44 +398,81 @@ export class Orchestrator {
         }
 
         const newAction = new CallSubOrchestratorWithRetryAction(name, retryOptions, input, instanceId);
+        let attempt = 1;
+        let subOrchestratorCreated: SubOrchestrationInstanceCreatedEvent | undefined;
+        let subOrchestratorFailed: SubOrchestrationInstanceFailedEvent | undefined;
+        let taskRetryTimer: TimerCreatedEvent | undefined;
+        for (let i = 0; i < state.length; i++) {
```
davidmrdavid, line 405: same nit: `for...of` preferred

src/orchestrator.ts
```diff
@@ -736,15 +808,13 @@ export class Orchestrator {
     }
 
     /* Returns undefined if not found. */
-    private findTaskScheduled(state: HistoryEvent[], name: string): TaskScheduledEvent | undefined {
-        const returnValue = name
-            ? state.filter((val: HistoryEvent) => {
-                return val.EventType === HistoryEventType.TaskScheduled
+    private findTaskScheduled(state: HistoryEvent[], name: string, startIndex: number = 0): TaskScheduledEvent | undefined {
```
davidmrdavid, line 811: Unless I'm mistaken, I don't see `startIndex` being passed explicitly at any point, so I'm unsure of how to use it

src/orchestrator.ts
ConnorMcMahon: I think this was probably a case of me trying to micro-optimize the perf of this loop. Probably not worth it.

ConnorMcMahon: I'm leaning away from this. I went with the explicit for-loop as we use the index in the loop body.

src/orchestrator.ts
ConnorMcMahon: Same here.

src/orchestrator.ts
ConnorMcMahon: I think this was a holdover from a previous iteration. I got rid of it in the rebase.

Issue #144, opened 2020-01-11
Is async supported in Durable Entity?


terurou:
**Describe the bug**
await function doesn't work in Durable Entity.

**Investigative information**

- Durable Functions extension version: 2.1.0
- durable-functions npm module version: 1.8.3
- Language (JavaScript/TypeScript) and version: JavaScript
- Node.js version: 10.18.0

**To Reproduce**
```js
const df = require("durable-functions");

module.exports = df.entity(async function(context) {
    const currentValue = context.df.getState(() => 0);
    switch (context.df.operationName) {
        case "add":
            await new Promise((f, r) => {
                setTimeout(() => {
                    context.df.setState(currentValue + 1);
                    f();
                }, 1000);
            });
            break;
        case "reset":
            context.df.setState(0);
            break;
        case "get":
            context.df.return(currentValue);
            break;
    }
});

```

**Actual behavior**
State isn't updated when "add" is called.

**Additional context**
Entity#handle is seem

https://github.com/Azure/azure-functions-durable-js/blob/bbb0f8b6a2f3fa4b628686f74b196d88edf80a4e/src/entity.ts#L33

cgillum:
Indeed it looks like we aren’t handling async Functions correctly. We’ll investigate.

anthonychu:
Thanks for reporting and identifying the problem! Looks like that line might just need to be changed to `await Promise.resolve(this.fn(context));` but should ensure tests are updated/added for this.

PR #168, opened 2020-02-25
Handling async entities

aaronpowell:
This addresses issue #144 but bubbling the async handling of entities all the way up to the entity shim.

I've added a test and some samples to it as well.

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -30,7 +30,6 @@ export class Entity {
             context.df = this.getCurrentDurableEntityContext(entityBinding, returnState, i, startTime);
 
             try {
-                // this.fn(context);
                 await Promise.resolve(this.fn(context));
                 if (!returnState.results[i]) {
                     const elapsedMs = this.computeElapsedMilliseconds(startTime);
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -30,7 +30,8 @@ export class Entity {
             context.df = this.getCurrentDurableEntityContext(entityBinding, returnState, i, startTime);
 
             try {
-                this.fn(context);
+                // this.fn(context);
+                await Promise.resolve(this.fn(context));
                 if (!returnState.results[i]) {
                     const elapsedMs = this.computeElapsedMilliseconds(startTime);
                     returnState.results[i] = new OperationResult(false, elapsedMs);
--- a/src/shim.ts
+++ b/src/shim.ts
@@ -25,7 +25,7 @@ export function orchestrator(fn: (context: IOrchestrationFunctionContext) => Ite
 export function entity(fn: (context: IEntityFunctionContext) => unknown)
     : (context: IEntityFunctionContext) => void {
     const listener = new Entity(fn).listen();
-    return (context: IEntityFunctionContext) => {
-        listener(context);
+    return async (context: IEntityFunctionContext) => {
+        await listener(context);
     };
 }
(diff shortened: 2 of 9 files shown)


aaronpowell:
@anthonychu There was a few more places than you noted in #144 that needed tweaking, but I think I've got that bug sorted now.

anthonychu:
Thanks for this @aaronpowell. I told you it would be easy to trick you into helping us. It worked!

aaronpowell:
@anthonychu you know how to spike my interest 🤣

aaronpowell:
@cgillum The tests have both sync and async entities in them, the only differences is you _always_ have to `await` the created entity from `shim.ts`.

I didn't test the new sample I added though as I couldn't work out how to trigger it.

ConnorMcMahon:
@aaronpowell 

To clarify, there are no breaking changes for customers who have written their entities as is today, right? The new `await` is handled purely in `shim.ts`, and does not affect existing customer code.

src/entity.ts
```diff
@@ -30,7 +30,8 @@ export class Entity {
             context.df = this.getCurrentDurableEntityContext(entityBinding, returnState, i, startTime);
 
             try {
-                this.fn(context);
+                // this.fn(context);
```
ConnorMcMahon: nit: remove this commented out line.

src/shim.ts
```diff
@@ -25,7 +25,7 @@ export function orchestrator(fn: (context: IOrchestrationFunctionContext) => Ite
 export function entity(fn: (context: IEntityFunctionContext) => unknown)
     : (context: IEntityFunctionContext) => void {
     const listener = new Entity(fn).listen();
-    return (context: IEntityFunctionContext) => {
-        listener(context);
+    return async (context: IEntityFunctionContext) => {
```
ConnorMcMahon: I'm a little confused here, because it looks like we are returning a type of `(context: IEntityFunctionContext) => Promise<void>) instead of `(context: IEntityFunctionContext) => void).  I'm not sure why typescript doesn't yell at us here, and what the implications are of us returning that new type instead.

Assuming the end to end works, I would recommend we just update the return type of `entity` to our new return type.

src/entity.ts
aaronpowell: Whoops, yep I'll remove that line.

src/shim.ts
aaronpowell: As far as I've been able to determine the returned type didn't really make a different, but I'm not 100% sure as to how the trigger is invoked and whether the trigger is concerned about the type.

aaronpowell:
@ConnorMcMahon @cgillum I've created a test application locally that uses this branch and entities that are sync and async worked.

Here's the `async` entity I used:

```js

module.exports = df.entity(async function(context) {
  await new Promise(resolve => setTimeout(() => resolve(), 0));

  context.log("starting entity");
  let currentValue = context.df.getState(() => 0);

  switch (context.df.operationName) {
    case "add":
      context.log("updating entity");
      const amount = context.df.getInput();
      currentValue += amount;
      break;
    case "get":
      context.log("getting entity value");
      context.df.return(currentValue);
      break;
  }

  context.df.setState(currentValue);
});
```

aaronpowell:
Hey folks, is there anything else outstanding on this PR getting merged?

ConnorMcMahon:
Looks good Aaron. Will merge.

PR #171, opened 2020-02-28
Add more exported types

Chroscielski:
Current types exported via index.ts have some missing types if you want to strongly type everything in the code using durable functions to benefit from intellisense. I have added what has been missing from my point of view (I have manually imported that types from src directory in my current solution).

Example typing (df was not visible by intellisense previously):
```TypeScript
const orchestratorTrigger: (ctx: IOrchestrationFunctionContext) => void =
  orchestrator(function* f(ctx: IOrchestrationFunctionContext): IterableIterator<unknown> {
    const message: ICommandMessage = ctx.df.getInput() as ICommandMessage;
}
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,18 +1,21 @@
 import { DurableHttpRequest, DurableHttpResponse, EntityId, EntityStateResponse,
-    OrchestrationRuntimeStatus, RetryOptions } from "./classes";
-import { getClient } from "./durableorchestrationclient";
+    IOrchestrationFunctionContext, OrchestrationRuntimeStatus, RetryOptions, Task } from "./classes";
+import { DurableOrchestrationClient, getClient } from "./durableorchestrationclient";
 import { entity, orchestrator } from "./shim";
 import { ManagedIdentityTokenSource } from "./tokensource";
 
 export {
     DurableHttpRequest,
     DurableHttpResponse,
+    DurableOrchestrationClient,
     entity,
     EntityId,
     EntityStateResponse,
     getClient,
+    IOrchestrationFunctionContext,
     ManagedIdentityTokenSource,
     orchestrator,
     OrchestrationRuntimeStatus,
     RetryOptions,
+    Task,
 };


davidmrdavid:
Hi @Chroscielski , thank you for your contribution! ⚡ 

Your PR caught my eye as we're looking to provide better type hints and IntelliSense for our TypeScript users. A few questions:

* Do you mind sharing a little bit more context around your snippet? 
* Where is `ICommandMessage` coming from?
* If possible, do you have a few more instances of strong typing snippets that you'd like to have available but currently do not work out of the box? Just trying to add them to our testers.

Thanks!

Chroscielski:
Hi @davidmrdavid, thanks for checking out my contribution! ⚡ 

- My snippet is just an example of the orchestrator function with strongly typed `IOrchestrationFunctionContext`, since it was not exported and we had our tslint very restrictive about typing everything (and not using `any`). Additionally it has impact on the comfort of writing the orchestrator itself, since its context is more specific than usual functions context. In our case the code does much more than just getting the input of course, but I wanted to provide some short overview on why I would like to have the type exported. 
- iCommand is our project internal type, since we are expecting to receive a well-known structure, based on which we invoke suitable logic. Some more context: I have used durable functions with my team to integrate external data providers (different APIs) with our system and based on the command we are either getting data from all providers or just invoking some specific logic for previously selected one. That's why we need some metadata along with requests.
-   These were only ones bothering me for now, but I'm more than happy to provide more in another merge request if I'll find some more :)

I hope it's helpful. Do not hesitate to ask more if something is unclear or you have some more questions.

davidmrdavid:
Thanks for your feedback @Chroscielski.
Interesting, I had not thought of the impact of 'nesting' the orchestration definition within the trigger. I wonder if more of our TypeScript users leverage that pattern. I'll follow-up with the rest of the team to give this some thought.

In any case, I do want to confirm to you that we're embarking on designing better TypeScript experience by being more cognizant of what types we export and what docs get picked up by IntelliSense. As a result, I'll put the brakes on merging this PR for now, as we're still designing these changes without sacrificing too much backwards-compatibility.

Some of those efforts have already started in this PR: https://github.com/Azure/azure-functions-durable-js/pull/185
and, in the coming months, I expect that PR to grow in scope from a simple change to a the larger refactoring we've been talking about. It'll just take a bit! 

I'll probably reach out again!

Issue #172, opened 2020-03-18
Upgrade TypeScript version to enable TypeDoc documentation


davidmrdavid:
We currently support TypeScript version `3.2.4`, which appears to interfere with TypeDoc's auto-generated documentation due to Handlebars new strict prototype access policy. This has supposedly been fixed as of TypeDoc version `0.15.1` but the latest version supporting TypeScript `3.2.4` is version `0.14.2`. 

We should attempt updating our TypeScript version to the latest version to support TypeDoc and to avoid analogous issues with other libraries.

davidmrdavid:
Closing and it's been addressed

PR #173, opened 2020-03-20
Upgrade TypeScript to version 3.8.3

davidmrdavid:
This is to address the following issue: https://github.com/Azure/azure-functions-durable-js/issues/172

The PR contains a few subtle changes with the most important one being changing the orchestrator's generator signature from ` IterableIterator<unknown>` to  ` Generator<unknown, unknown, any>`

Let's dive into this a little more.
If I'm not mistaken, `IterableIterator<unknown>` meant that we had a generator that *yields* data of `unknown` type. The return value and its '.next()'s input value are not typed / left as `any`

Meanwhile, `Generator<unknown, unknown, any>` means that we have a generator that *yields* unknowns, returns *unknowns*, and that its '.next()'s input value is `any`. 

For more information of these signatures, please review [this](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-6.html).

Therefore, I don't believe this change makes our code any more more or less robust than before, it is more or less equivalent. In addition, upgrading our typescript version comes with many handy type-safety benefits.

I considered having more concrete types when instantiating generator, but they would have been too restrictive and, worst of all, they would have forced TypeScript end-users to use type-assertions or type guards after each yieldable statement. The use of `any`, while technically unsafe, preserves the programming patterns of exisitng users.

Let me know what you think, I'd be happy to discuss further. Thanks @ConnorMcMahon for your input on this one!

PS: TypeDoc works now :) . At least locally


src/orchestrator.ts
```diff
@@ -25,7 +25,7 @@ export class Orchestrator {
     private newGuidCounter: number;
     private subOrchestratorCounter: number;
 
-    constructor(public fn: (context: IOrchestrationFunctionContext) => IterableIterator<unknown>) { }
+    constructor(public fn: (context: IOrchestrationFunctionContext) => Generator<unknown, unknown, any>) { }
```
ConnorMcMahon: Is there a reason we went with `any` for `TNext`? My understanding was that we should always use `unknown` instead of `any` whenever possible.

test/testobjects/TestOrchestrations.ts
```diff
@@ -306,12 +306,12 @@ export class TestOrchestrations {
     });
 
     public static ThrowsExceptionFromActivity: any = df.orchestrator(function*(context: any)
-    : IterableIterator<unknown> {
+    : Generator<void, void, any> {
```
ConnorMcMahon: I realize you are just editing the type signature to be correct here, but is there a reason we need the type signature here? Many of the other orchestrations in this file are missing them.

src/orchestrator.ts
davidmrdavid: Yeah, and it's quite interesting actually. 

So if we were to use `unknown` for `TNext`, TypeScript would force end-users would to use type-assertions (`x as <type>`) after every `yield` statement in their functions / orchestrators / etc. 

This can be evidenced when writing tests, which simulate user code. In the tests, if we had an `unknown` instead of `any`, each `yield` statement would need to followed by a type-assertions (`output as <-type>`) in order to do pretty much anything with the result. Say, for instance, before accessing property from it. I didn't want to do this because it would break people's expected usage of the service, assuming they were using TypeScript too.

This is why I said that the older version of typescript also leaves `TNext` as any / untyped.

ConnorMcMahon: Makes sense!

davidmrdavid: In the near future, I would love to discuss this with the TypeScript folks because, as you point out, `any` types are an escape hatch from the type-checker.

I imagine some future where you could type your azure functions such that TypeScript would know that each `yield` corresponds to a user-defined type. In that scenario, we could enforce something stricter than `unknown` and `any` without forcing type assertions. It would be so cool!

But that's for another PR, much later in the future 😄

test/testobjects/TestOrchestrations.ts
davidmrdavid: Good observation. I automatically changed the `IterableIterator` to `Generator` because, in the past, this test had been given a type for some reason. 

But removing the type completely still type-checks, so I'm removing it. Thanks for pointing it out!

aaronpowell:
Is the change to the publicly exposed type of the orchestrator not a potentially breaking change?

davidmrdavid:
@aaronpowell, thank you for bringing that up, I didn't consider that.

I think the answer is *yes*. At the same time though, the old type no longer type-checked in the latest TypeScript version. I suspect this change would be required if we wanted TypeScript users to adopt this library without having to use an older TypeScript version.

Perhaps there's a way around this though, I'm no expert. Please let me know what you think.

davidmrdavid:
To further clarify, the old `IterableIterator<T>` type is no longer good enough for our orchestrator's generator as of the latest TypeScript. At the same time, the new `Generator<,,,>` type is not recognized by our current version of TypeScript.

aaronpowell:
Fair enough @davidmrdavid, and besides I don't think people should be relying on the type there.

PR #174, opened 2020-04-03
Allow task sets in df.task.all and df.task.any

CharlesGuillot:
Allow task sets in df.task.all and df.task.any typing

### DIFF
--- a/src/tasks/itaskmethods.ts
+++ b/src/tasks/itaskmethods.ts
@@ -1,4 +1,4 @@
-import { Task } from "./task";
+import { TaskBase } from "./taskinterfaces";
 import { TaskSet } from "./taskset";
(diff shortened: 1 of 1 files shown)


davidmrdavid:
Hi @CharlesGuillot! Thank you for your contribution!! ⚡ ⚡

We're taking a close look at your PR, as we're trying to unify `TaskSet` and `Task` under the same identifier. In the future, we may remove `TaskSet` as a distinguishing identifier from `Task` so we still have to discuss whether or not this exact PR is something we'll want to merge, but we're driving a fix forward that will address this and a handful of other scenarios. 

You can follow our discussion here: https://github.com/Azure/azure-functions-durable-js/pull/185

Issue #175, opened 2020-04-04
durable function Http starter function failed 


andyzhu:
Error when try to the sample code:

const df = require("durable-functions");

module.exports = async function (context, req) {
    const client = df.getClient(context);
    **const instanceId = await client.startNew(req.params.functionName, undefined, req.body);**

    context.log(`Started orchestration with ID = '${instanceId}'.`);

    return client.createCheckStatusResponse(context.bindingData.req, instanceId);
};

The above highlighted code cause the following error

2020-04-04T02:24:40.434 [Information] Executing 'Functions.HttpStart' (Reason='This function was programmatically called via the host APIs.', Id=7dd0fb95-dd62-4781-a002-3ef44620aaad)
2020-04-04T02:24:50.700 [Error] Executed 'Functions.HttpStart' (Failed, Id=7dd0fb95-dd62-4781-a002-3ef44620aaad)
Result: Failure
[...]

PR #176, opened 2020-04-10
Removes synchronous calls when starting a new durableorchestrator

Dc9894:
When undefined is passed as input to _startNew_ to start a new orchestrator function in Azure Functions 3.X an InvalidOperationException is thrown due to a call to a synchronous call to a Read method since synchronous calls have been disabled by default in this version of the SDK. #175

### DIFF
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -603,7 +603,7 @@ export class DurableOrchestrationClient {
                 .replace(this.instanceIdPlaceholder, (instanceId ? `/${instanceId}` : ""));
         }
 
-        const response = await this.axiosInstance.post(requestUrl, input !== undefined ? JSON.stringify(input) : "");
+        const response = await this.axiosInstance.post(requestUrl, JSON.stringify(input));
         if (response.data && response.status <= 202) {
             return (response.data as HttpManagementPayload).id;
         } else  {


davidmrdavid:
:zap::zap: Thank you for your contribution! :zap: :zap:

I'll review this soon and let you know if its ready to merge.

anthonychu:
Interesting find! @ConnorMcMahon any idea why this works?

cgillum:
The exception occurred when reading the HTTP request payload, so my guess is that passing undefined instead of the empty string causes us to send a request with no payload. No payload == no reading == no exception?

Dc9894:
@cgillum I think is the other way around, when passing an empty string into the `startNew()` method, the call to the `axios` library post method is made with empty string instead of an empty JSON. My guess is that when this happens, the library internally tries to read this empty string synchronously and then parse it to JSON.

If we pass a JSON always calling JSON.stringify on the payload the library doesn't ever tries to read the payload synchronously. So, JSON payload == no synchronous reading == no exception

ConnorMcMahon:
@Dc9894 I'm not quite sure if that is the case. I believe the forbidden synchronous IO is only for the .NET Core layer that is the functions runtime. I believe all of the code that is run within Node is done via a GRPC call, so .NET Core should have no visibility into whether axios is doing anything synchronously or not.

davidmrdavid:
@Dc9894, I just tested this change locally, and it did not fix the problem for me.

Some assumptions about your environment:

(1) I'm assuming you're running this on version `v1.8.3` of [Durable-Extension](https://github.com/Azure/azure-functions-durable-extension). Is that right? If I'm not mistaken, you _should_ already using `v1.8.3` if you're using _bundles_ / got your Functions App initialized by the VSCode extension.

(2) I'm assuming you're working on a recent-enough version of `axios`. Do you mind reporting what version of axios you're using? I am running `6.13.4`. You can check this by going to your local Durable Functions JS directory and running `npm -v axios`

A couple notes:

(1) **Location of the error:** As others have pointed out, the exception occurs on the C# end of things, which drives the Durable Functions logic. In particular, take a look at line 486 [here](https://github.com/Azure/azure-functions-durable-extension/blob/v1.8.3/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs#L486). When reproducing this error, on our end, the sample triggers the use of that if-statement, which calls on a synchronous operation if the JSON-body is `null`, which in return triggers the exception.

(2) **Long-term fix:** Thankfully, this issue has already been fixed in an upcoming version of the Durable Functions Extension. The release of it is currently blocked by some delays with the _bundles_ system but it should be coming up nonetheless.

(3) **Short-term fix:** I want to make sure you're not blocked by this problem. If you are, please let us know and we can discuss a workaround for the time being!

Thanks again for your contribution 😄

davidmrdavid:
Hi @Dc9894! Just following-up on this, since I have not heard back. As I mentioned above, I can't seem to replicate your fix so we must be working on top of different dependencies. If you share those details, I can try again to replicate it. In any case, the internal fix should be released soon so hopefully this is not a blocker. If it is, please let us know!

In the meantime, I'll close this issue but please re-open if you'd like to continue this conversation 😄 . Thank you so much for reaching out and submitting a contribution! We're always excited for community PRs so please consider submitting another in the future!! 🚀

PR #181, opened 2020-04-16
Preserve JSDocs on the compiled declaration files

davidmrdavid:
This PR makes changes to our compilation procedure to enable the compiled code to preserve JSDocs, which will enable us to generate documentation from it.

The trick was to turn our compilation into a two-step process. First, we generate declaration files and JS files preserving all comments. This alone would make our package size grow quite large. The next step is to compile again, but this time to only generate JS files *without* comments, therefore overriding the commented-enabled JS files while preserving JSDoc files in the declarations


tsconfig.nocomments
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "declaration": false,
```
ConnorMcMahon, line 4: To clarify, `declaration` to `false` means we won't regenerate our typescript d.ts files, meaning that we won't overwrite those without comments, right?

davidmrdavid: Yes, you're precisely right. 
More than a two-step compilation, it is a 'compile-twice' approach.

davidmrdavid: I was also looking into adding comments somehow, but JSON doesn't officially support comments, so I worry about breaking something if I attempt something brittle like just adding a "comment:" field to the JSON and hoping nothing tries to parse it.

What we can do is create a separate `npm script` called something like `removeComments` and replace the changes I made to the `build` definition to simply call on that other step. 

That way, we attach a proper name to this new step.

PR #185, opened 2020-04-20
Unifying TaskSet and Task under a single identifier

davidmrdavid:
This PR is informed by a problem brought up in: https://github.com/Azure/azure-functions-durable-js/pull/174

We currently export `TaskSet` and `Task` types, which are really internal types. What we want is for users to deal with a generic `Task` type that encompasses the union of those interfaces. We already have that, in `TaskBase`,  but the name `TaskBase` suggests that users should look for something more specific, so it might be a good idea to change it.

So this PR renames `Task` to `SingleTask`, and hides it from JSDoc, meaning that it won't be exported for use in IntelliSense. We also hid `TaskSet` in a previous [PR](https://github.com/Azure/azure-functions-durable-js/pull/182).

This PR then renames `TaskBase` to `Task` and keeps it for IntelliSense use.

This is obviously a breaking change, but we are only recently paying close attention to the IntelliSense usage so this slipped under our radar. I'm happy to discuss alternatives or workarounds, this is just here to open a discussion, unless folks feel comfortable merging.

An outstanding "problem" is that TaskSet and SingleTask are still exported in [src/classes.ts](https://github.com/Azure/azure-functions-durable-js/blob/dev/src/classes.ts) so I believe users still have access to it even if we hide them in JSDocs. We should consider doing a sweep over those exports and removing unwanted declarations.

Thanks!


aaronpowell:
I'll put this on my list for post-Build to review.

davidmrdavid:
Superceded by -https://github.com/Azure/azure-functions-durable-js/pull/305

PR #195, opened 2020-06-16
Improving the exposed type system

aaronpowell:
This PR shows how the generics can be used to make the exposed type interface more consumable to the end user.

There is one main issue, when using `yield` it is not possible to have the type inferred due to a limitation in the TypeScript type system (see https://github.com/microsoft/TypeScript/issues/36967).

An option could be to change the `Task` class to inherit from `Generator<T, T, T>` and thus have a `next` method exposed so if people don't wish to use `yield`, and instead handle the iterations themselves, they can do that, which would give them type safety.


src/entity.ts
```diff
@@ -78,9 +78,11 @@ export class Entity {
             entityId: bindingInfo.self,
             operationName: currentRequest.name,
             isNewlyConstructed: !batchState.entityExists,
-            getState: this.getState.bind(this, batchState),
+            getState: this.getState.bind(this, batchState) as <T>(
```
davidmrdavid, line 81: I must admit that I'm struggling to read the `getState` signature. Do you mind describing it?

test/integration/entity-spec.ts
```diff
@@ -78,13 +79,20 @@ function entityStateMatchesExpected(actual: EntityState, expected: EntityState):
     }
 }
 
-class MockContext {
+class MockContext implements IEntityFunctionContext {
```
davidmrdavid, line 82: Can you remind me why the `IEntityFunctionContext` is needed now?

src/tasks/taskinterfaces.ts
```diff
@@ -1,34 +1,29 @@
 import { IAction } from "../classes";
 
 // Base interfaces
-/** @hidden */
```
davidmrdavid, line 4: why are we removing all these `@hidden` markings?

davidmrdavid:
Great stuff! So, to summarize, this PR improves the type inference around entities which we previously short-circuited via `any` type annotations.  Before I continue, I should say that I'm only just getting started with TypeScript, although I do have experience with type-systems in languages like Haskell, so the syntax here is not immediately obvious to me 😅 .

Am I right in understanding that these changes mostly affect the experience around declaring and implementing entities, but **does not** provide better inference for an entity _client_? This being because the entity client communicates with entities via `yield`-able calls which ,as you point out, are still a challenge to type-annotate.

So what is the current solution for typing `yield`-ed activity call? Do we just cast it forcibly via `<yieldable> as <type>` ?

Finally, I really like your idea of exposing a type-safe alternative API for activities. We could definitively expose an iterator, or alternatively use a continuation-passing-style / Promise / monadic syntax. @ConnorMcMahon , I seem to recall that y'all considered promises as a way to schedule activity calls? Any thoughts on revisiting that idea as an alternative, type-safe, variant for TypeScript?

davidmrdavid:
Ah, one final question for both of y'all.

Other than backwards compatibility concerns, these changes should be directly applicable to orchestrators, right?

davidmrdavid:
Also, I'm surprised that the PR is not requiring a reviewer's approval for merge...

test/integration/entity-spec.ts
aaronpowell: To get the tests to properly understand type inference the context that is used in the tests now implements an actual context. Previously, the arguments to the test functions were `any` typed so a ducktyped class was usable.

src/tasks/taskinterfaces.ts
aaronpowell: Result of a bad merge. I couldn't remember if they were needed or not. I'll go back and put them in where they were removed.

aaronpowell:
Yep, mostly this is around improving consumption of entity functions, and defining them by exposing more of the underlying types (that were previously `any`) or by giving the option to provide a generic argument to operations.

I want to get some time to explore moving where the generic argument is to higher up the definition chain, ideally when you define the entity function:

```typescript
module.exports = df.entity<MyType>(function (context) {
```

That would be nicer if the `MyType` can flow into the `context.df.getState`/etc. functions.

aaronpowell:
By and large they should be non-breaking operations, the only breaks would be when people are relying on the `unknown` or `any` types being returned. From what I understand, I haven't refactored any of the actual functionality.

src/entity.ts
davidmrdavid: Ah, thanks for the playground , I see it it now.
I think this solution makes sense for now, and perhaps we'll revisit the type signature and its associated refactoring in the future, assuming a cleaner syntax is possible.

davidmrdavid:
Ah yes, having an entity declaration-level generic would be really elegant, definitely keep us posted! Let us know when you think you've done enough experimentation and we can decide whether or not to merge then. Also, at that point, we can probably discuss if it makes sense to port these directly into the orchestrator API.

And yeah, I don't think you've changed any functionality, we _should_ be good! As always, many thanks, this is super helpful.

src/entity.ts
aaronpowell: You might be able to make a function yourself that does the currying, but then you're probably going to lose the performance optimisations that the JS runtimes have around how `Function.prototype.bind` works.

aaronpowell:
I've got it working now with strongly typed entities, check out https://github.com/Azure/azure-functions-durable-js/pull/195/files#diff-b5b5544459f11b191866dffca7ebdac8 for a reference example.

Now what happens is that the `entity` function in `shim.ts` takes a generic argument (but defaults to `unknown` so it's `entity<T = unknown>`) and then the `IEntityFunctionContext` and `DurableEntityContext` are both generic, using that same generic argument. This then cascades through the generic on `Entity` and the weird casting stuff no longer is needed when creating the `DurableEntityContext`!

aaronpowell:
And for funsies sake, I've also included generic orchestrators:

```typescript
import * as df from "durable-functions";

module.exports = df.orchestrator<number>(function* (context) {
    const entityId = new df.EntityId("CounterEntityTypeScript", "myCounterTS");

    const currentValue: number = yield context.df.callEntity(entityId, "get");
    if (currentValue < 10) {
        yield context.df.callEntity(entityId, "add", 1);
    }
});
```

Now, you still have the problem that when you `yield` the result the yielded value doesn't know what type it is, but you get type safety if you try and pass in the third argument of `callEntity`. Although, I'm not sure if that's right, because the data you're passing into the entity doesn't have to match the type of the overall orchestrator...

davidmrdavid:
Hurray on typing the entity generically at a declaration level! 🎉 

With regards to the generically-typed orchestrator, can you elaborate on how type safety is checked for when passing the third argument, that is, the input? Since the entity is only referred to by their string name, I'm unsure of how it reason about that the input and output types, unless you're referring to something else.

As a second point, now that I think about it, does the type of the orchestrator refer the value it will get when calling `getInput` or will it refer to the value it returns? Same for the entity. Technically, they do not have to be the same kind of value, right?  _hmmmmm_

aaronpowell:
After writing my the last comment I started thinking more about the input stuff, both with orchestrators and entities.

I think I'll roll back the change on the orchestrator side, since the type doesn't make sense on the cascade through, the input doesn't need to match the type of the underlying entity that you're calling, so a top-level generic is kind of incorrect.

This then led me to think about the `getInput` for an entity. Presently, the way the types cascade down, all generics within an entities context (`getState`, `getInput`, `return`, `setState`) use the same generic type. Now, this works fine on a _Counter_ but if you've some something more complex where the handling of an entity operation has many different _types_ of input (say a Person entity and there are input operations like _add child_, _add pet_, _add partner_ they would all have different typed input).

This leads me to think that `getInput` should be generic using its own generic argument, so that you can call it appropriately each time.

aaronpowell:
Just an aside, I'll pull the linting of `samples` out to its own PR as it's cluttering up here, and it's not really related to this PR anyway.

src/durableorchestrationcontext.ts
```diff
@@ -5,7 +5,7 @@ import { TokenSource } from "./tokensource";
  * Parameter data for orchestration bindings that can be used to schedule
  * function-based activities.
  */
-export class DurableOrchestrationContext {
+export class DurableOrchestrationContext<T> {
```
davidmrdavid, line 8: So we're removing these, right? To avoid propagating the "type" of the orchestration to constrain the type of the Entity state?

src/entity.ts
```diff
@@ -92,23 +92,23 @@ export class Entity {
         batchState.entityState = undefined;
     }
 
-    private getInput(currentRequest: RequestMessage): unknown | undefined {
+    private getInput(currentRequest: RequestMessage): T | undefined {
```
davidmrdavid, line 95: Currently, `getInput` and `getState` are both of type `T`, which I think overly constraints the entity. What about providing two types:
one for input
one for output

I can imagine a scenario where the input to an entity is a number, and that returned value is, say, a string. Think of a fizz-buzz'ing entity :)

davidmrdavid:
Hey @aaronpowell , just following-up on this discussion, which is fascinating by the way...

Just to make sure we're in the same page, is part of what you are saying that entities should be allowed to set their states to values of a type that is different to their input types? If so, I completely agree! Which is why I think we should have two generics in the entity declaration: one for input, one for output.

Then, if we have an entity that takes in multiple inputs, such as in the `addPet`, `addPartner`, `addChild` scenario you propose, and end-user would simply `OR`' them together in a sum-type, no?

Let me know what you think 😄 , just an idea, perhaps it's not idiomatic to TypeScript

src/durableorchestrationcontext.ts
aaronpowell: Yep, haven't had time to do it

src/entity.ts
aaronpowell: Yep, the signature of `getInput` should be `<TInput>(currentRequest: RequestMessage): TInput | undefined`, not inheriting the `T` from the `Entity` signature.

`getState` should use `T` from the `Entity` signature though.

aaronpowell:
The way that I conceptualise it is that `Entity` is a boxing type of your serialized data, which is why it becomes `Entity<T>`. The calling/signaling of an `Entity` allows you to pass in any sort of argument, because that's input into the call/signal handler. Operations on the `Entity` itself (`getState`, `setState`, `return`) are all working with the type of your serialized data, and should be typed accordingly.

Quick bash on the code (I haven't updated the repo yet) would see this:

```typescript
export default df.entity<Cart>(context => {
  switch (context.df.operation) {
    case ADD_TO_CART:
      const item = context.df.getInput<CartItem>();
      const currentCart = context.df.getState() || createNewCart();
      // do stuff
      context.df.setState(...);
      break;
    case REMOVE_FROM_CART:
      const itemToRemove = context.fg.getInput<CartItemId>();
      // do stuff

    case CHECKOUT:
      // do stuff

    // more case statements
  }
});
```

Does that follow how you'd envisage it?

davidmrdavid:
@aaronpowell 
I like it. @ConnorMcMahon  hasn't chimed in yet, although he's quite busy with other things at the moment.

That makes sense to me. As for me, I'd say let's just go with it!

aaronpowell:
Rolled back the orchestrator typing stuff and tweaked `Entity` a bit more. I've also put that "shopping cart" example in with an entity + orchestrator, so you can see how it could work.

davidmrdavid:
Nice, I like it so far. Thanks for the sample!

I see the entity's input and state are still coupled as the same type; though perhaps you're not quite done with it yet. Please let me know when you think this is 100% ready, as to avoid bugging you with reviews over in-progress work 💪  . Many thanks again.

package.json
```diff
@@ -1,85 +1,86 @@
...
-  "bugs": {
-    "url": "https://github.com/Azure/azure-functions-durable-js/issues"
-  },
-  "homepage": "https://github.com/Azure/azure-functions-durable-js#readme",
-  "directories": {
-    "lib": "lib",
-    "test": "test"
-  }
+    "name": "durable-functions",
+    "version": "1.4.1",
```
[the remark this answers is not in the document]
aaronpowell: Yeah, i'll merge in with `dev` fix some of these

samples/package.json
```diff
@@ -1,19 +1,28 @@
...
+        "request": "^2.87.0",
+        "seedrandom": "^2.4.3"
+    },
+    "scripts": {
+        "build": "tsc",
+        "build:production": "npm run prestart && npm prune --production",
+        "watch": "tsc --w",
+        "prestart": "npm run build && func extensions install",
+        "start:host": "func start",
+        "start": "npm-run-all --parallel start:host watch",
```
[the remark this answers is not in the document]
aaronpowell: No idea, it was there already (or maybe it's from the template...)

aaronpowell:
> I see the entity's input and state are still coupled as the same type

Not quite @davidmrdavid, what I've done is defined that there is a new type for the input, `TInput` (same with return and `TReturn`) but the default _type_ is `TState` which is the entity state.

This means that if you're working with simple types for the entity and input, you don't have to explicitly type it on input or return value.

samples/package.json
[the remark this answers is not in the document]
aaronpowell: Wait, I did introduce it... maybe I did a `func new` which adds them 🤔

davidmrdavid (approved):
It's got my LGTM! Sorry for the long delay here. Let me just make sure @ConnorMcMahon is also on-board and then we'll merge this

aaronpowell:
All good, I'd totally forgot it was open 😅

src/durableorchestrationcontext.ts
```diff
@@ -65,7 +65,7 @@ export class DurableOrchestrationContext {
      * @returns A Durable Task that completes when the called activity
      * function completes or fails.
      */
-    public callActivity(name: string, input?: unknown): Task {
+    public callActivity<T>(name: string, input?: T): Task {
```
ConnorMcMahon, line 68: I'm a little concerned by the confusion this will cause in comparison with C#, where `CallActivityAsync<T>` has `T` refer to the `return` type, not the input type.

Unfortunately, with our use of generator functions, we don't have a way to cleanly specify a type for the return value.

src/tasks/task.ts
```diff
@@ -36,7 +36,7 @@ import { TaskBase } from "./taskinterfaces";
  * return firstDone.result;
  * ```
  */
-export class Task implements TaskBase {
+export class Task<T = unknown> implements TaskBase {
```
ConnorMcMahon, line 39: I'm not sure I see the benefit of using generic parameters for `Task` unless we can enforce `yield` on `Task<T>` returns a value of T.

ConnorMcMahon (commented):
I like the overall idea, especially for enforcing the state type for entities, but I think the generic methods on the orchestrator to specify input types will lead to some confusion, and I'm not convinced it provides that much value currently.

src/tasks/task.ts
aaronpowell: Yeah, as I mentioned up top, the use of `yield` leaves you stuck with type inference unless https://github.com/microsoft/TypeScript/issues/36967 is resolved. The only alternative would be to replace the usage of `Task` with a `Promise` and move away from generator functions to Promise-based async, but that'd be a rather huge change that I probably wouldn't advise without a major version bump and breaking changes 😛

ConnorMcMahon: 100% agree. It would be really nice to get better type safety, but it's already somewhat user-enforced in C# even due to the fact that activities and suborchestrations are referenced by name instead of by type.

src/durableorchestrationcontext.ts
aaronpowell: I'll admit I haven't looked at the C#/.NET API for a long time, and wasn't aware of how the type system there was represented (I didn't realise that `input` is `object` there, I assumed it would've been typed).

Admittedly, a lot of this PR was originally written around the Entities site of durable functions, not orchestrators, as I was trying to get them strongly typed so you can do stuff like I demo in https://github.com/Azure/azure-functions-durable-js/pull/195/files/8dc1bf93ebfda83db5ee8bda88764ba18e09e371#diff-efcc2398105c3d67a41a35488e1f984e83a7d96923c24c07bb9fa4e84a2c3fb6

Then I figured I'd apply the same pattern on orchestrators, came across the limitation with `yield` and went "eh, it's not breaking anything" so left it in.

ConnorMcMahon: I think orchestrations are a bit of a mess right now typing wise, and until we find a way around our current `yield` typing, I think it may be best to leave them as is.

I think the entity stuff we should definitely move forward with.

I'm willing to budge on the orchestration stuff if it feels sufficiently "typescripty", but I personally feel the cons of the generic typing outway the pros. 

That being said, it still makes perfect sense for `GetInput` I would say!

davidmrdavid:
So it sounds like the fastest path towards merge is to remove the generic typing of `callActivity` and other orchestrator-specific APIs because it contradicts the typing in the .NET API (we type output instead of input). The only _potential_ exception to this being `getInput`, which might be nice to type. Said differently, let's keep the typing in entities, remove it for orchestrators.

Did I get that right, @ConnorMcMahon?

davidmrdavid:
Ok then @aaronpowell, it sounds like we have a clear path to merge then (outlined in my response above). Since it's been some time since we were last actively working on this, I could merge this into a temporary branch that I own and perform that refactor myself. Alternatively, if you have the cycles, you could take it over the finish line and refactor it yourself. I just want to be mindful of your time, especially considering how long we've kept this in PR-limbo. What would work best for you?

davidmrdavid:
@aaronpowell, just a friendly ping about my question above^

aaronpowell:
Sorry, been busy the last week. I'll try and get it tackled before the end of this week

aaronpowell:
Rollback done.

:shipit:

davidmrdavid (approved):
The orchestrator-specific bits seem to have been roll'ed back. This seems good to me, let's hear from @ConnorMcMahon first and, if he agrees, let's merge this!

src/tasks/task.ts
```diff
@@ -70,7 +70,6 @@ export class Task implements TaskBase {
          */
         public readonly timestamp?: Date,
         /**
-         * @hidden
```
ConnorMcMahon, line 73: Is there a reason we removed this hidden value?

ConnorMcMahon (approved):
LGTM, just want to verify the @hidden removal to see if that was intentional.

src/tasks/task.ts
davidmrdavid: Seems like we're ready to merge @aaronpowell! Do you mind chiming in on this comment so I can finalize this PR? Thanks!

aaronpowell: Added back in

davidmrdavid:
Finally merged! Thank you all so much for the patience here, very excited to ship this! 🚀

aaronpowell:
:shipit:

PR #196, opened 2020-06-17
linting the samples

aaronpowell:
Separating this from PR #195


davidmrdavid:
Looks good to me! Please resolve the conflict in the `package.json` which is probably because we just increased the release version, and we'll merge!

aaronpowell:
Merge conflict fixed

PR #204, opened 2020-07-02
Increasing the package version, for release

davidmrdavid:
As in title.


davidmrdavid:
@ConnorMcMahon : I'm never too sure, but I did just in case

Issue #215, opened 2020-09-14
Number written as string treat as number when call callActivity, so large number will be rounded


yamachu:
**Describe the bug**

The size of numbers that can be handled by the JavaScript number type is limited.
We sometimes treat large-number as BigInt or string for avoiding that limitation.

That large number was treated as number when I used stringified number as `callActivity`' s parameter.

**Investigative information**

- Durable Functions extension version: 
- durable-functions npm module version: 1.4.3
- Language (JavaScript/TypeScript) and version: JavaScript
- Node.js version: 12

***If deployed to Azure App Service***

- Timeframe issue observed:
- Function App name: 
- Function name(s):
- Region: japan-east
- Orchestration instance ID(s): 

**To Reproduce**

See repro repo
https://github.com/yamachu/repro-durable-functions-js-stringify-large-number

1. request [DurableFunctionsHttpStart](https://github.com/yamachu/repro-durable-functions-js-stringify-large-number/tree/master/DurableFunctionsHttpStart) with big_id parameter
2. see logs

https://github.com/yamachu/repro-durable-functions-js-stringify-large-number/blob/master/logs/view-azure-portal-logstream-curl-big_id-12345678901234567890.txt

This log was logged at Azure Portal LogStream.

**Expected behavior**

If `callActivity("activityName", "12345678901234567890")`, 
we got input "12345678901234567890" as string in "activityName" activity.

**Actual behavior**

If `callActivity("activityName", "12345678901234567890")`, 
we got input 12345678901234567000 as number in "activityName" activity.

**Known workarounds**

Add some characters "@", "n" or .... to stringified number, and remove these characters in activity.

**Additional context**

- Development environment (ex. Visual Studio): Visual Studio Code Insiders
- Links to source: https://github.com/yamachu/repro-durable-functions-js-stringify-large-number
- Additional bindings used
- Function invocation IDs

davidmrdavid:
Hi @yamachu ,

Thank you so much for reaching out! Indeed, you seem to have triggered a bug that's actively being discussed with the team at large. In short, somewhere in our pipeline, your string is getting _interpreted_ as a number and therefore changing its precision. We have a similar problem [here](https://github.com/Azure/azure-functions-durable-extension/issues/1476) but instead of interpreting string-represented numbers, its with interpreting string-represented `Date` objects.

[...]

PR #217, opened 2020-09-16
Stringify'ing all inputs to avoid precision loss

davidmrdavid:
**WIP** : _Do not merge, work in progress...._

**Related issues:** https://github.com/Azure/azure-functions-durable-extension/issues/1476, https://github.com/Azure/azure-functions-durable-js/issues/215  

**Problem:** We have encountered a handful of tickets where values are somehow interpreted, and changed, in the pipeline between orchestrator-activity and trigger-activity. A classic example of this is `Date` serialization, where interpreting a Date mid-pipeline could change its encoding. Something similar would occur with `BigInt`s, where interpreting it mid-pipeline could introduce precision problems. These are the cases described in the two related issues above respectively.

**Insight:** 

In Durable Python,  the `Input` field of an `Action` object is serialized before the serializing the `Action` object itself. This means that, in order to interpret an `Input` field as anything other than a string, you have to decode the `Input` field after decoding the `Action` object itself. 

The effect of this is that out-of-proc input data flowing through `durable-extension` is interpreted as string, and left alone. We can see an example of this eager input serialization in python [here](https://github.com/Azure/azure-functions-durable-python/blob/65188f28fc2f6951a91293435bd6d51cef6a193a/azure/durable_functions/models/actions/CallActivityAction.py#L19) and [here](https://github.com/Azure/azure-functions-durable-python/blob/65188f28fc2f6951a91293435bd6d51cef6a193a/azure/durable_functions/models/actions/CallSubOrchestratorAction.py#L16). Note, the `dumps` method there means "dump to a JSON-string". Observe how that occurs at the level of object construction, so when the object itself gets serialized later, the `Input` gets "doubly" serialized.

However, we do not do this in JS. I think this is because JS can natively emit JSON objects, so we the extension can readily accept its data. That being said, it runs the risk of C# and others interpreting JSON-native datatypes like integers and dates, and changing their precision. 

**This PR:**

This PR introduces the same eager-serialization of `Input` fields in `Action` objects that we see in Python. This is quite the invasive change, and somehow none of our tests complained, so we are going to require some serious testing, and adding extra tests, before merging.

### DIFF
--- a/src/actions/callactivityaction.ts
+++ b/src/actions/callactivityaction.ts
@@ -6,14 +6,7 @@ export class CallActivityAction implements IAction {
     public readonly input: unknown;
 
     constructor(public readonly functionName: string, input?: unknown) {
-        // If we fail to stringify inputs, they may get deserialized incorrectly.
-        // For instance: "13131" might get interpreted as a number.
-        // Somehow this doesn't appear to occur with other datatypes, but we should
-        // investigate that further.
-        if (typeof input === "string") {
-            input = JSON.stringify(input);
-        }
-        this.input = input;
+        this.input = Utils.processInput(input);
         Utils.throwIfEmpty(functionName, "functionName");
     }
 }
--- a/src/actions/callactivityaction.ts
+++ b/src/actions/callactivityaction.ts
@@ -10,7 +10,7 @@ export class CallActivityAction implements IAction {
         // For instance: "13131" might get interpreted as a number.
         // Somehow this doesn't appear to occur with other datatypes, but we should
         // investigate that further.
-        if (input instanceof String) {
+        if (typeof input === "string") {
             input = JSON.stringify(input);
         }
         this.input = input;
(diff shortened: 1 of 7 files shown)


davidmrdavid:
Ok so some preliminary testing tells me this works just fine in `callActivity` with inputs as numbers, strings, arbitrary lists, and JSON objects.

davidmrdavid:
Note to self: We can make this PR less intrusive by only stringifying inputs that we know are to be kept as strings.

mhoeger:
@davidmrdavid - Just as a note, i think this bugfix will technically represent a breaking change even for just the string case. ex: my activity function passes in a number as a `string`, but i read it as a `number` on the orchestrator side and do some numeric operations against it. Would definitely be worth calling out in the release notes and maybe a minor version upgrade (although a break technically should only be on major versions - i think you all have a bit more leeway because users rev the version)

For the string case, you are probably already thinking of it but one example you might want to try: A moment.js time object. Today, this will just show up as a string representation of the time instead of the rich object. Not sure what doing JSON.stringify will do to that?

davidmrdavid:
The latest commit to this PR makes this "hack" apply only to strings, where we know they'll be safe.

ConnorMcMahon (approved):
LGTM!

Issue #236, opened 2021-01-14
getStatusBy and getAllStatus brings back only 100 results


pmatthews05:
**Describe the bug**
When calling client.getStatusBy the maximum number of results it calls back is 100 items. There doesn't seem to be a way to page in the node code.

**Investigative information**

- Durable Functions extension version: 
- durable-functions npm module version: ^1.4.4
- Language (JavaScript/TypeScript) and version: Typescript ^3.3.3
- Node.js version: 12.20.0

**To Reproduce**
https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management?tabs=javascript#query-instances-with-filters

Following Microsoft example above, whenthe code is run against a large environment that has many orchestrations, only 100 items come back. Looking at the node code /src/durableorchestrationclient.ts it seems that you are using the Rest API call to get the statuses within the function getStatusInternal.

According to the documentation of the REST API,  

> If more results exist, a continuation token is returned in the response header. The name of the header is x-ms-continuation-token.
> If you set continuation token value in the next request header, you can get the next page of results. This name of the request header is also x-ms-continuation-token.

https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-api#response-2

Cannot see any headers captured or used in the code.

**Expected behavior**
Either the x-ms-continuation-token is returned when you make a call, and the getStatusBy can have the x-ms-continuation-token passed in so that the developer can perform the paging themselves.

Or internally in the getStatusInternal you handle the paging and keep calling until you have all the results.
  
**Actual behavior**
Only a maximum of 100 items are being returned.

**Known workarounds (unsuccessful)**

We tried the below workaround, and although it does bring back lots more results, it doesn't seem to be capturing all orchestration statuses. We are not convinced that the results are being returned in the lastUpdatedTime order of earliest first.

``` TS
  
 //returns minimum date if there is not LastUpdateTime.
   let lastRunTime: Date = await GetLastUpdateTime();

    const runtimeStatus = [
        df.OrchestrationRuntimeStatus.Canceled,
        df.OrchestrationRuntimeStatus.Completed,
        df.OrchestrationRuntimeStatus.ContinuedAsNew,
        df.OrchestrationRuntimeStatus.Failed,
        df.OrchestrationRuntimeStatus.Pending,
        df.OrchestrationRuntimeStatus.Running,
        df.OrchestrationRuntimeStatus.Terminated,
    ];

    do {
[...]

PR #244, opened 2021-02-17
Handle pagination in GetStatus API

davidmrdavid:
Addresses: https://github.com/Azure/azure-functions-durable-js/issues/236

**Context:**
When the `getStatus` / `getStatusBy` APIs are called, their response sizes are capped at 100 results. If more data can be fetched, a continuation token is passed-in in the response header.  Then, to fetch that data, we need to perform another request with that continuation token in the request's header.  This is from: https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-api#response-2

**The problem:**
Previously, the Durable Functions for JavaScript `getStatusBy` API did not handle the case where more than 100 results were available.  So it forced users with many results to manually construct the HTTP request by themselves and to manage the continuation token that explicitly. 

**This PR**
This PR augments the `getStatus` / `getStatusBy` APIs to manage continuation tokens automatically, therefore free'ing users from the burden of managing HTTP pagination themselves.

**Follow-up tasks**
To track this item in the Durable Python SDK


src/durableorchestrationclient.ts
```diff
@@ -870,7 +870,23 @@ export class DurableOrchestrationClient {
...
     }
 
-    private async getStatusInternal(options: GetStatusOptions): Promise<AxiosResponse> {
+    /**
+     * Internal method that gets the status of all orchestration instances that
+     * match the specified conditions. It handles pagination automatically by
+     * calling itself recursively until no more continuation tokens are found.
+     * @param options Return orchestration instances which were created
+     *  after this Date.
+     * @param [continuationToken] Continuation token corresponding to the
```
davidmrdavid, line 879: Just FYI that the bracket-enclosed names denote optional parameters in JSDoc

src/durableorchestrationclient.ts
```diff
@@ -958,7 +974,34 @@ export class DurableOrchestrationClient {
...
+        let axiosConfig = undefined;
+        if (continuationToken) {
+            axiosConfig = {
+                headers: {
+                    "x-ms-continuation-token": continuationToken,
+                },
+            };
+        }
+
+        // We call the getStatus endpoint and construct a promise callback to handle the recursion
```
ConnorMcMahon, line 987: I _really_ don't like that we have to do this pagination logic. This is really a consequence of us having an indirect proxy to the IDurableClient object via the HTTP apis (even with local RPC). The HTTP APIs are a subset of the C# APIs, meaning we have to do workarounds like this on the worker side, and some of these workarounds (like this one), can have major perf implications.

That being said, I think we have to do this, at least in the short term. I think this points to us needing to have a more comprehensive line using the localRpcEndpoint to IDurableClient implementations (both Durable 1.x and Durable 2.x) before we can get rid of this workaround,  and unfortunately we can't wait for that in the meantime.

ConnorMcMahon (approved):
I think this is the best we can do for now, so this gets my reluctant approval.

test/unit/durableclient-spec.ts
```diff
@@ -244,6 +244,49 @@ describe("Durable client RPC endpoint", () => {
...
+            // The getStatusBy() method should do a GET to http://127.0.0.1:17071/durabletask/instances/?createdTimeFrom=2020-01-01T00:00:00Z&createdTimeTo=2020-01-01T23:59:59Z&runtimeStatus=Pending,Running,Completed,Terminated,Failed
+            const expectedUrl = new URL(`${testRpcOrigin}/durabletask/instances/`);
+            const createdTimeFrom = "2020-01-01T00:00:00.000Z";
+            const createdTimeTo = "2020-01-01T23:59:59.000Z";
+            const runtimeStatus = "Pending,Running,Completed,Terminated,Failed";
+
+            const scopeWithTokenResponse = nock(expectedUrl.origin)
+                .get(expectedUrl.pathname)
+                .query({ createdTimeFrom, createdTimeTo, runtimeStatus })
+                .reply(200, [null, null], { "x-ms-continuation-token": "myToken" });
```
ConnorMcMahon, line 261: small nit: Could we at least fill these with dummy objects. Returning an array filled with `null` as a shortcut feels a bit dirty. Ideally we would fill it with actual JSON representations of the data, but I realize that is a lot of work.

davidmrdavid: Of course, I agree the `null` was a little dirty.
Just replaced it with proper JSONs

src/durableorchestrationclient.ts
davidmrdavid: I've created [this](https://github.com/Azure/azure-functions-durable-extension/issues/1693) ticket to continue this conversation. Thanks for bringing this up :)

Issue #167, opened 2020-02-25
Cannot find namespace 'NodeJS'. when using TypeScript 


mhoeger:
Using:
"durable-functions": "^1.3.3"

and TypeScript:
"typescript": "^3.3.3"

Encountered error on `tsc`:
```
node_modules/durable-functions/lib/src/utils.d.ts:8:56 - error TS2503: Cannot find namespace 'NodeJS'.
```

Workaround:
Add to devDependencies:
"@types/node": "10.14.2"

PR #270, opened 2021-06-24
Add missing NodeJS types to dependencies

davidmrdavid:
Resolves: https://github.com/Azure/azure-functions-durable-js/issues/167

This PR adds missing typing dependencies to our TypeScript package. This should allow our TypeScript users to be able to compile their projects without needing to do an extra installation step via: `npm install @types/node`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
         "rimraf": "~2.5.4",
         "typedoc": "^0.17.1",
         "uuid": "~3.3.2",
-        "validator": "~10.8.0"
+        "validator": "~10.8.0",
+        "@types/node": "10.14.2"
     },
     "devDependencies": {
         "@azure/functions": "^1.2.3",
@@ -56,7 +57,6 @@
         "@types/debug": "0.0.29",
         "@types/mocha": "~7.0.2",
         "@types/nock": "^9.3.0",
-        "@types/node": "~6.14.7",
         "@types/rimraf": "0.0.28",
         "@types/sinon": "~5.0.5",
         "@typescript-eslint/eslint-plugin": "^2.21.0",


package.json
```diff
@@ -45,7 +45,8 @@
         "rimraf": "~2.5.4",
         "typedoc": "^0.17.1",
         "uuid": "~3.3.2",
-        "validator": "~10.8.0"
+        "validator": "~10.8.0",
+        "@types/node": "10.14.2"
```
cgillum, line 49: I'm not an experienced TypeScript or Node.js developer, but this change seems counter-intuitive to me. Why do we need to add `@types/node` to `dependencies` instead of to `devDependencies`?  Isn't this something only needed by the compiler at development time, or is there some runtime dependency?

Also, Marie's comment on the linked issue suggests that the workaround to the original issue is to add `@types/node` to `devDependencies`:
> Workaround:
> Add to devDependencies:
> "@types/node": "10.14.2"

Weren't we already doing the right thing prior to this PR?

cgillum (changes requested):

package.json
davidmrdavid: You're right that we were already doing what Marie's comment suggested but, as you could test if you were to scaffold a TypeScript DF app on VSCode right now, the code will not compile without installing `@types/node` :( .

The reason for this is that some of our code has, in its type signatures, types from `@types/node`. When compiling TypeScript, two kinds of files are generated (1) the JS code, and (2) a definitions file with the TypeScript type-signatures. The latter is then used when end-users load our module in their own TypeScript projects. Due to the fact that we don't include `@types/node` in our dependencies, our TypeScript end-users will get compilation errors when loading our module as they will not be able to find the source of some of our datatypes . In that sense, it's definitely a blocker for our TypeScript users.

Of course, many TypeScript end-users are likely to already have `@types/node` installed in their machine, as it's such a common type-definitions module! But, if they do not (which might be rare), then they're in for a surprise and will have to find the source of these missing types on their own :-)

cgillum: Hmm...so perhaps I am confusing the list of dependencies for us to build our SDK with the list of dependencies for customers to build their apps _using_ our SDK. Essentially, anything we put in `devDependencies` won't be automatically reflected in the `devDependencies` of our users, right? And this is why we have to put it in `dependencies` (so that they have what they need to compile against our exposed APIs)?

davidmrdavid: That's correct!

davidmrdavid: Essentially, anything in `devDependencies` is "just for us" while developing the project. This can include things like linters, libraries that accelerate debugging, and anything that allows the code to compile but isn't needed at runtime (JS) _and also is not needed during the development time of an end-user_. In this case, `@types/node` is needed at compilation time for our TypeScript end users.

cgillum: Cool, thanks for the clarification!

cgillum (approved):

PR #278, opened 2021-07-20
Update currentUtcDatetime on failed -WithRetry scenarios

davidmrdavid:
This PR fixes a recently identified bug where an orchestration's deterministic time property, `currentUtcDatetime`, is not correctly updated when a -WithRetry API, such as `callActivityWithRetry` fails. There were 3 logical errors that contributed to this behavior.

(1) On the presence of a failed Task, the JS SDK would feed the exception to the generator _before_ it had the chance to update its currentUtcDatetime value. This meant that the `currentUtcDatetime` value _could_ get out of sync in the presence of failed tasks. 

(2) The HistoryEvent timestamps in our History array were not guaranteed to be treated as `Date` datatypes. Instead, they would sometimes be a `String`. This meant that comparisons between timestamps could sometimes misbehave.

(3) In the presence of a **failed** -WithRetry API, the segment of History events that helps determine the final state of that Task can contain multiple intermediate `OrchestrationStarted` events. Our `currentUtcDatetime`-updating logic did not take this into consideration and was prone to setting its next value to the timestamp of one of these intermediate  `OrchestrationStarted` events instead of choosing the timestamp of the last one in the sequence.

This PR addresses all these issues, and also makes **minor**, but unrelated, tweaks to our node  testing scripts as part of standard maintenance.


src/orchestrator.ts
```diff
@@ -185,6 +180,16 @@ export class Orchestrator {
                     context.df.isReplaying = state[partialResult.completionIndex].IsPlayed;
                 }
 
+                if (TaskFilter.isFailedTask(partialResult)) {
+                    if (!gen.throw) {
```
cgillum, line 184: When would we expect `gen.throw` to be undefined? Can you add a comment explaining this?

src/utils.ts
```diff
@@ -17,14 +17,34 @@ export class Utils {
...
 
+    public static hasOwnProperty<X extends {}, Y extends PropertyKey>(
+        obj: X,
+        prop: Y
+    ): obj is X & Record<Y, unknown> {
+        // informs TS that an object has a property
+        return obj.hasOwnProperty(prop);
+    }
+
+    public static parseDates(obj: unknown): unknown {
```
cgillum, line 36: How about `parseTimestampsAsDates`, since you're checking for a specifically named "Timestamp" field?

src/orchestrator.ts
```diff
@@ -167,11 +159,14 @@ export class Orchestrator {
...
-                        e.Timestamp > decisionStartedEvent.Timestamp
-                );
+                const newDecisionStartedEvent = state
+                    .filter(
+                        (e, index) =>
+                            e.EventType === HistoryEventType.OrchestratorStarted &&
+                            e.Timestamp > decisionStartedEvent.Timestamp &&
+                            index < completionIndex
+                    )
+                    .pop();
```
cgillum, line 169: What's the `.pop()` for? Do we normally change the state (history) inside orchestrator functions? This seems strange to me because I would assume the state should be immutable here.

ConnorMcMahon: If I understand this correctly, it won't edit state. `.filter()` creates a copy of the list.

This is equivalent to a `.Select().Last()` LINQ query in C#.

src/orchestrator.ts
ConnorMcMahon: I believe this is mainly to satisfy the TypeScript type-checker, as it is technically nullable I believe? Probably should never happen.

src/orchestrator.ts
cgillum: No objection if this is just operating on a copy of the history. 👍

src/orchestrator.ts
```diff
@@ -133,21 +133,13 @@ export class Orchestrator {
...
+                const completionIndex = partialResult.completionIndex;
+
                 // The first time a task is marked as complete, the history event that finally marked the task as completed
                 // should not yet have been played by the Durable Task framework, resulting in isReplaying being false.
                 // On replays, the event will have already been processed by the framework, and IsPlayed will be marked as true.
-                if (state[partialResult.completionIndex] !== undefined) {
-                    context.df.isReplaying = state[partialResult.completionIndex].IsPlayed;
-                }
-
-                if (TaskFilter.isFailedTask(partialResult)) {
```
ConnorMcMahon, line 143: I'm trying to think if there are many major implications of moving this. We are essentially moving it past the case where the customer `returns` a non-yielded completed task, which shouldn't affect this code path, as well as after we update the replay state. 

I am pretty sure this is all safe, just wanted to highlight any potential changes this would cause.

src/utils.ts
```diff
@@ -17,14 +17,34 @@ export class Utils {
...
+        if (
+            typeof obj === "object" &&
+            obj !== null &&
```
ConnorMcMahon, line 39: Just curious about the ordering of these. What is `typeof obj` return if `obj` is null?

src/orchestrator.ts
```diff
@@ -167,11 +159,14 @@ export class Orchestrator {
                     return;
                 }
 
-                const newDecisionStartedEvent = state.find(
-                    (e) =>
-                        e.EventType === HistoryEventType.OrchestratorStarted &&
-                        e.Timestamp > decisionStartedEvent.Timestamp
-                );
+                const newDecisionStartedEvent = state
```
ConnorMcMahon, line 162: Can we explicitly add a comment that we grab the last `OrchestratorStarted` event before our current place in the history?

src/orchestrator.ts
```diff
@@ -167,11 +159,14 @@ export class Orchestrator {
...
 
-                const newDecisionStartedEvent = state.find(
-                    (e) =>
-                        e.EventType === HistoryEventType.OrchestratorStarted &&
-                        e.Timestamp > decisionStartedEvent.Timestamp
-                );
+                const newDecisionStartedEvent = state
+                    .filter(
+                        (e, index) =>
+                            e.EventType === HistoryEventType.OrchestratorStarted &&
```
ConnorMcMahon, line 165: I am like 90% sure this is correct (or at least as correct as we can be with our current per-Task linear scan algorithm).

I am a bit uncertain about the case where we have a `df.task.whenAll()` with a bunch of `callActivityWithRetry()` operations that may finish out of order due to retries, but I think the `e.Timestamp > decsionStartedEvent.Timestamp` will help prevent us from traveling "backwards" in time as we process subsequent retried activities that may have finished earlier.

In that case, this filtered list should be empty, and `.pop()` will just return undefined and we will keep the current value for `decisionStartedEvent`.

ConnorMcMahon (approved):
Overall, I am fairly certain this will work, and I can't think of a cleaner way with our current linear scan algorithm.

I think this logic will become substantially easier to reason about with the new replay logic we will implement.

src/orchestrator.ts
davidmrdavid: +1, this is simply because the type of `gen` is `IterableIterator<unknown>`. That `unknown` means that we need to be quite safe about what properties we access, thus the check.

A cleaner way to go about this is probably to assign it a more concrete type (the generator type) earlier in this method. However, this snippet of code already existed and has simply been moved down, so this bit where we check "does `gen` have a `throw` method?" is not being introduced in this PR :)

src/utils.ts
davidmrdavid: Good idea, I've made this change in my latest commit :)

src/orchestrator.ts
davidmrdavid: This is precisely right ;)

src/orchestrator.ts
davidmrdavid: +1, I was also quite worried about this change but I agree that this appears safe. Another way to think about it, which helps in building confidence about this change, is that we've moved it right next to where we feed a successful value to the generator via `gen.next`. It makes sense that `gen.next` would be right next to `gen.throw` and since `gen.next` works correctly, so should this case :)

src/utils.ts
davidmrdavid: Ah yes, this is an interesting one: `typeof null` returns `"object"`
You can try it out here!: 

https://www.typescriptlang.org/play?#code/MYewdgziA2CmB00QHMAUAXAngB1iAZgARgCu00AlANwBQhhQA

src/orchestrator.ts
davidmrdavid: Added!

src/orchestrator.ts
cgillum: Good to know. A comment might be a nice idea since I think this will be non-obvious to other readers like it was for me.

cgillum (approved):

src/orchestrator.ts
davidmrdavid: Done!

PR #290, opened 2021-11-09
Update various dependencies

davidmrdavid:
The node package manager provides very helpful utilities that warn package maintainers of when they should be upgrading various dependencies for performance, security, and correctness reasons. One such tool is `npm audit` which performs an audit of dependencies and raises warnings when those are due for an update.

Running `npm audit fix` automatically updates dependencies that do not introduce breaking changes. This PR is the result of running `npm audit fix`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -514,14 +514,6 @@
                 "follow-redirects": "^1.14.0"
             }
         },
-        "node_modules/backbone": {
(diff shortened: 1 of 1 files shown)


davidmrdavid:
Merging as this a non-breaking, non-feature-changing, PR

PR #292, opened 2021-11-20
Remove ESLint dependency

davidmrdavid:
Our previous ESLint dependency, which we use for auto-formatting, was being flagged by `npm audit`. We're removing it temporarily until we can an alternative configuration or alternative dependency.

I'm also updating our `mocha` version, for running unit tests.
And simplifying our build-script steps.

This is fairly routine PR, no features should be impacted


amdeel (approved):
LGTM!

package.json
```diff
@@ -67,14 +67,12 @@
         "chai": "~4.2.0",
         "chai-as-promised": "~7.1.1",
         "chai-string": "~1.5.0",
-        "eslint": "^6.8.0",
```
ejizba, line 70: Eslint is the standard for Node.js and TypeScript in particular. It does a lot more than just formatting - it's primarily for preventing silly mistakes in code (like not awaiting a promise). I would highly recommend re-enabling eslint ASAP or never removing it in the first place.

Can you quickly try using v7 instead of v6 and see if that works? v7 should get rid of the npm audit complaints and you can turn off any failing rules in [here](https://github.com/Azure/azure-functions-durable-js/blob/dev/.eslintrc) instead of disabling _all_ of eslint. I do think it's fine to get rid of "prettier-eslint", though, because I don't think you're using it. Feel free to ping me if you have issues

davidmrdavid: I'll reach out internally on Monday and block this merge for a little bit. Thanks for reaching out!

ejizba: No problem, just subscribed to the repo like last week 😁 👀

always watching

davidmrdavid: @ejizba: I managed to get this working by upgrading `eslint` and its related dependencies. Could I please get another review? Thanks

ejizba (approved):

PR #298, opened 2021-12-03
Update Azure DevOps pipeline

davidmrdavid:
This PR makes 3 updates to our azure-pipelines config.

1. Our test matrix now matches that of the NodeJS worker. This means adding NodeJS 16 to our tests and dropping tests on Node8.
2. Just like the NodeJS worker, we add an "Artifact publishing" job, which will come in handy when automating npm releases
3. In the artifact publishing job, we add an SBOM-building step, for auditing purposes. This was modeled after this NodeJS worker PR:  https://github.com/Azure/azure-functions-nodejs-worker/pull/489

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -73,6 +73,8 @@ jobs:
             displayName: "npm ci"
           - script: npm build
             displayName: "npm build"
+          - script: npm prune --production
+            displayName: "npm prune --production" # so that only production dependencies are included in SBOM
           - task: ManifestGeneratorTask@0
             displayName: "SBOM Generation Task"
             inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,8 +3,10 @@ variables:
 name: $(MODULE_VERSION)-$(Date:yyyyMMdd)$(Rev:.r)
 
 trigger:
-    - main
-    - dev
+    branches:
+        include:
+            - main
+            - dev
 
 jobs:
     - job: UnitTests
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,11 +2,15 @@ variables:
     { MODULE_VERSION: "1.2.2", NODE_10: "10.x", NODE_12: "12.x", NODE_14: "14.x", NODE_16: "16.x" }
 name: $(MODULE_VERSION)-$(Date:yyyyMMdd)$(Rev:.r)
 
-trigger:
+pr:
     branches:
         include:
             - main
             - dev
+
+trigger:
+    - main
+    - dev
 
 jobs:
     - job: UnitTests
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "typings": "lib/src/index.d.ts",
     "scripts": {
         "clean": "rimraf lib",
-        "lint": "eslint --ext .ts,.json src/",
+        "lint": "eslint --fix --ext .ts,.json src/",
```
ejizba, line 21: The "--fix" option should only be used locally and any resulting changes should be checked in to git

azure-pipelines.yml
```diff
@@ -1,44 +1,76 @@
...
-  - script: npm run test:nolint
-    displayName: 'npm build and test (no linting)'
-    condition: eq(variables['NODE_VERSION'], variables['NODE_8'])
+    - job: UnitTests
+      strategy:
+          matrix:
+              UBUNTU_NODE10:
+                  IMAGE_TYPE: "ubuntu-latest"
+                  NODE_VERSION: $(NODE_10)
+              UBUNTU_NODE14:
```
ejizba, line 18: Are you intentionally skipping NODE12?

azure-pipelines.yml
```diff
@@ -1,44 +1,76 @@
...
+                  IMAGE_TYPE: "macOS-latest"
+                  NODE_VERSION: $(NODE_16)
+      pool:
+          vmImage: $(IMAGE_TYPE)
+      steps:
+          - task: NodeTool@0
+            inputs:
+                versionSpec: $(NODE_VERSION)
+            displayName: "Install Node.js"
+          - script: npm install
```
ejizba, line 49: This should be `npm ci` just like the `BuildArtifacts` job. It's very similar to `npm install` but designed to be used in continuous integration systems like azure pipelines (aka why it's called `npm *ci*`)

azure-pipelines.yml
davidmrdavid: Seems I skipped it because it wasn't in the current `azure-pipelines.yml` of the nodeJS worker :)
I'll add it here, but just a heads up:  https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/azure-pipelines.yml

azure-pipelines.yml
davidmrdavid: Thanks for the heads up. will do

azure-pipelines.yml
ejizba: You can see our version matrix here: https://github.com/Azure/azure-functions-nodejs-worker#azure-functions-nodejs-worker

We support 14/16 on the latest worker, but the previous worker supports 10/12/14. Since you included 10, I assumed you were supporting both the latest worker and previous worker

davidmrdavid: I see. I don't see a reason why we wouldn't include Node12 if we have Node10, so we'll test them both.

ejizba (approved):

Issue #291, opened 2021-11-18
Activity function does not work after retry


noriyukitakei:
**Describe the bug**
I'm using Durable Functions. I encountered an unexpected exception after retry of Activity function.

I have one Orchestrator function and 2 Activity functions as below:

- OrchestratorMain

- ActivityA
- ActivityB

This is my code:

OrchestratorMain
```
const df = require("durable-functions");

module.exports = df.orchestrator(function* (context) {

    const firstRetryIntervalInMilliseconds = 1000;
    const maxNumberOfAttempts = 10;
    const retryOptions = new df.RetryOptions(firstRetryIntervalInMilliseconds, maxNumberOfAttempts);

    const tasks = [];

    for (let i = 0; i < 5; i++) {
        tasks.push(context.df.callActivityWithRetry("ActivityA", retryOptions));
    }

    yield context.df.Task.all(tasks);

    if (!context.df.isReplaying) context.log("All ActivityA functions done");

    const results = yield context.df.callActivityWithRetry("ActivityB", retryOptions);
    
    if (!context.df.isReplaying) context.log("Orchestrator function done");

});
```

ActivityA
```
module.exports = async function (context) {
    context.log('ActivityA start');

    const min = 1;
    const max = 3;
    const a = Math.floor(Math.random() * (max + 1 - min)) + min;

    if (a == 1) throw Error('hoge');

    context.log('ActivityA end');
};
```

ActivityB
```
module.exports = async function (context) {

    context.log('ActivityB start');

    context.log('Do something...');

    context.log('ActivityB End');
};
```

`ActivityA` throws the error randomly.

If `ActivityA` is done with no retry, `ActivityB` is done as I expected.

However, if `ActivityA` is done after some retries, `ActivityB` is not done. Instead of `ActivityB`, `ActivityA` seems to be done according to the logs generated by "context.log".

Why this happens? Please tell me the solution...

**Investigative information**

- Durable Functions extension version:2.6.0
- durable-functions npm module version:1.5.2
- Language (JavaScript/TypeScript) and version:JavaScript
- Node.js version: 14

noriyukitakei:
I guess this behavior is very similar to [this issue](https://github.com/Azure/azure-functions-durable-js/issues/99).

cgillum:
Pinging @davidmrdavid

davidmrdavid:
Thanks for the ping @cgillum.

[...]

PR #300, opened 2021-12-08
Support V2 replay protocol

davidmrdavid:
This PR adds support for the V2 replay protocol in the DF extension. This protocol adds improved reliability and correctness by representing WhenAll and WhenAny tasks as proper Actions within the extension.

To do this, we need to parse the "upperSchemaVersion" field from the extension's input payload, which tells us the maximum protocol version supported by the extension. At time of writing, only 2 protocols exist: V1 and V2.

For V2, we simply change the representation of compound tasks (tasks of tasks) such that WhenAny and WhenAll tasks are not a list of action, but instead a list of tasks within a WhenAny or WhenAll action depending on the API invoked.

Finally, while these PR changes a lot of files, the change is pretty simple: we're just propagating the protocol version throughout various methods to determine the shape of our "Actions" output. 

Should partially address: https://github.com/Azure/azure-functions-durable-js/issues/291

### DIFF
--- a/src/actions/actiontype.ts
+++ b/src/actions/actiontype.ts
@@ -17,4 +17,6 @@ export enum ActionType {
     WaitForExternalEvent = 6,
     CallEntity = 7,
     CallHttp = 8,
+    WhenAny = 11,
+    WhenAll = 12,
 }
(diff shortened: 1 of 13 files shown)


davidmrdavid:
Superceded by -https://github.com/Azure/azure-functions-durable-js/pull/305

PR #305, opened 2022-01-07
Adopt a linear-time replay algorithm

davidmrdavid:
Follow-up to: https://github.com/Azure/azure-functions-durable-python/pull/302

### Problem Statement

A DF SDK's primary function is to rehydrate an orchestrator's state by processing its internal History log. We call this: "replaying an orchestrator". Previously, our approach to replay had a quadratic complexity cost in the number of History events.

This PR makes our replay algorithm linear-time instead. We do this by iterating over the orchestration History array only once, which is possible by making better use of the data within each History event. Based on our experience with this algorithm in the DF Python SDK, this change should give us unprecedented performance, correctness, and scalability improvements for JS/TS DF users.

### Other contributions

This PR is quite intrusive to our Node.JS SDK: most, if not all, the "hot paths" of our SDK have been re-worked. As a result, this was a good opportunity to evaluate dropping support for legacy behavior and interfaces/types that we were supporting mostly for backwards compatibility's sake *and* that newer SDKs such as Python does not include.

As a result, my goal after merging this PR is to release a new **major** version of this SDK.

#### A new replay driver

The `TaskOrchestrationExecutor` is the new driver of orchestration replay. Previously, each DF API was responsible for inferring it's corresponding Task status by searching through the History log. Now, this responsibility is centralized in the `TaskOrchestrationExecutor`. This class interleaves iterating through the History log with resuming the execution of the user-defined generator (the orchestrator code). While iterating through the History, it seeks to update the state of currently-scheduled ("open") tasks which, when resolved ("closed"), allow us to resume running the generator.

#### Support for the Extension-level OOProc replay schema V2

The DF extension "recently" introduced a new replay protocol for OOProc orchestrations. By subscribing to this new extension-level replay protocol, OOProc SDKs enjoy improved correctness when dealing with `WhenAny` and `WhenAll` tasks. This PR adds support to this new protocol.

To do this, it changes its representation of user **actions** depending on the  `upperSchemaVersion` field sent by the extension.

#### New task state machines

Tasks are now represented as state machines that can be signaled to update their state. This is particularly helpful in the case of WhenAny/WhenAll tasks, where the task needs to determine its current status by listening to state changes in its child-tasks.

### Breaking Changes

There are at least 3 breaking changes introduced in this PR.

1. We no longer support returning a Task from an orchestrator to be valid. Previously, a statement such as `return context.CallActivity(...)` would be treated as `return yield context.CallActivity(...)`. This is no longer the case.

2. We no longer support yielding the `ContinueAsNew` API, and doing so would throw an exception. This API is supposed to be _fire-and-forget_ and so it doesn't make sense to support "awaiting" it.

3. All user-facing Task types now inherit from single identifier :`Task`. Additionally, the user-facing Task` types have been simplified and no longer expose various property that were for framework-internal use only. Our users have been asking for a more streamlined experience with Task-types so this is long overdue as well.

To my knowledge, this is the full extent of breaking changes, but I feel that the changes in this PR are disruptive enough that there might be one other minor behavior I could be missing. In any case, the behavior of the Node SDK after this PR should be more consistent with the Python SDK, which I believe is a good thing.

### Follow-ups

This PR will need to be benchmarked and the changes extensively tested before a release.

### Note: 
Some files with major changes (like the orchestration context) are showing up as minimized due to large diffs. Don't miss them while reviewing :) 

Thanks!


package.json
```diff
@@ -30,7 +30,7 @@
         "validate:samples": "npm run build && npm --prefix samples install && npm run build:samples",
         "build:nolint": "npm run clean && npm run stripInternalDocs && echo Done",
         "stripInternalDocs": "tsc --pretty -p tsconfig.nocomments",
-        "test": "npm run validate:samples && npm run build && mocha --recursive ./lib/test/**/*-spec.js",
+        "test": "npm run build && mocha --recursive ./lib/test/**/*-spec.js",
```
davidmrdavid, line 33: This change is just because one of our unit-testing sample doesn't play well with re-building it over and over. We need to do something about that, bu tI think that's outside the scope of this PR

cgillum:
Do we need to make any documentation updates to account for the mentioned breaking changes?

src/actions/actiontype.ts
```diff
@@ -17,4 +17,6 @@ export enum ActionType {
     WaitForExternalEvent = 6,
     CallEntity = 7,
     CallHttp = 8,
+    WhenAny = 11,
+    WhenAll = 12,
```
cgillum, line 21: What happed to `9` and `10`? Should we add a comment?

src/actions/whenanyaction.ts
```diff
@@ -0,0 +1,12 @@
+import { ActionType, IAction } from "../classes";
+import { ProperTask } from "../tasks/internalTasks";
+
+/** @hidden */
+export class WhenAnyAction implements IAction {
+    public readonly actionType: ActionType = ActionType.WhenAll;
```
cgillum, line 6: Should this be `ActionType.WhenAny`?

src/durableorchestrationbindinginfo.ts
```diff
@@ -1,12 +1,25 @@
...
         public readonly instanceId: string = "",
         public readonly isReplaying: boolean = false,
-        public readonly parentInstanceId?: string // TODO: Implement entity locking // public readonly contextLocks?: EntityId[],
-    ) {}
+        public readonly parentInstanceId?: string,
+        upperSchemaVersion = 0 // TODO: Implement entity locking // public readonly contextLocks?: EntityId[],
+    ) {
+        // If the extension-sent upperSchemaVersion is within the range of values
+        // we support, we select it. Otherwise, we conclude it's higher than any
+        // version we support, so we default to our highest version.
```
cgillum, line 18: I'm having trouble understanding these comments because they seem contradictory. If `upperSchemaVersion` isn't specified, do we assume it's higher than any version we support or do we assume they're using the highest version?

src/durableorchestrationbindinginfo.ts
```diff
@@ -1,12 +1,25 @@
...
+        if (Object.values(UpperSchemaVersion).includes(upperSchemaVersion)) {
+            this.upperSchemaVersion = upperSchemaVersion;
+        } else {
+            this.upperSchemaVersion = UpperSchemaVersion.V2;
```
cgillum, line 22: Should there be a `Latest` value specified here instead of hardcoding `V2`?

cgillum (commented):
Partial review feedback. I still have more to review.

test/unit/timertask-spec.ts
```diff
@@ -1,30 +1,44 @@
...
+        const date = new Date();
+        const action = new CreateTimerAction(date, isCancelled);
+        const innerTask = new InnerTimerTask(0, action);
+
+        const task = new TimerTask(innerTask);
         task.cancel();
-        expect(task.action.isCanceled).to.equal(true);
-        expect(task.isCanceled).to.equal(true);
+        expect(task.isCancelled).to.equal(true);
+        expect(task.isCancelled).to.equal(true);
```
cgillum, line 32: Why are we running the same `expect` call twice?

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,493 @@
...
+     */
+    public async execute(
+        context: IOrchestrationFunctionContext,
+        history: HistoryEvent[],
+        schemaVersion: UpperSchemaVersion,
+        fn: (context: IOrchestrationFunctionContext) => IterableIterator<unknown>
+    ): Promise<void> {
+        this.schemaVersion = schemaVersion;
+        this.context = context.df;
+        this.generator = fn(context) as Generator<TaskBase, any, any>; // TODO: what happens if code is not a generator?
```
cgillum, line 112: Regarding this TODO, should we throw an error if it's not a generator?

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,493 @@
...
+
+    /**
+     * @hidden
+     * Determine if the current invocation has finished.
+     *
+     * @returns
+     *  True if the orchestration reached a `return` statement, or a `continueAsNew`.
+     *  False otherwise.
+     */
+    private orchestrationInvocationCompleted(): boolean {
```
cgillum, line 166: nit: I suggest coming up with more precise names for `hasExecutionCompleted` and `orchestrationInvocationCompleted`. An "invocation" and an "execution" don't have clearly distinguished meanings. Perhaps "isDoneExecuting()" and "completedSuccessfully()"?

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,493 @@
...
+     * @hidden
+     * Processes a History event, often by either by updating some deterministic API value, updating
+     * the state of a task, or resuming the user code.
+     *
+     * @param event
+     *  The History event we're currently processing
+     */
+    private processEvent(event: HistoryEvent): void {
+        const eventType = event.EventType;
+        switch (event.EventType) {
```
cgillum, line 180: Reuse `eventType` here?

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,493 @@
...
+                // The EventSent event requires careful handling because it is re-used among
+                // CallEntity and WaitForExternalEvent APIs.
+                // For CallEntity, the EventRaised event that contains that API's result will
+                // expect a TaskID that is different from the TaskID found at the root of this
+                // EventSent event. Namely, the TaskID it expects can be found nested in the
+                // "Input" field of the corresponding EventSent event. Here, we handle that
+                // edge-case by correcting the expected TaskID in our openTask list.
+                const key = event.EventId;
+                if (key in this.openTasks) {
+                    const task = this.openTasks[key];
```
cgillum, line 208: Are the above two lines redundant? Could we save some cycles by fetching the task and checking for `undefined`?  I noticed this in a few other places as well.

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,493 @@
...
+                    this.setTaskValue(event, isSuccess, idKey);
+                    return; // we return because the task is yet to be scheduled
+                };
+                this.deferredTasks[key] = updateTask.bind(this);
+                return;
+            }
+        } else {
+            throw Error(
+                `Task with ID ${key} could not be retrieved from due to its ID-key being of type ${typeof key}` +
+                    `We expect ID-keys to be of type number or string.` +
```
cgillum, line 320: nit: Add a space after the `.`.

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,493 @@
...
+            }
+        } else {
+            if (task.id === false) {
+                // The task needs to be given an ID and then stored.
+                task.id = this.sequenceNumber++;
+                this.openTasks[task.id] = task;
+            } else if (task.actionObj instanceof WaitForExternalEventAction) {
+                // The ID of `WaitForExternalEvent` tasks is the name of
+                // the external event. Given that multiple `WaitForExternalEvent`
+                // tasks can be led for the same event name at once, we need
```
cgillum, line 473: "led"?

cgillum (changes requested):

src/actions/actiontype.ts
davidmrdavid: Ah, good question! `9` and `10` are `SignalEntity` and `ScheduledSignalEntity` respectively. I'll add a clarifying comment

src/actions/whenanyaction.ts
davidmrdavid: I think you're right. Huh, I wonder why no tests caught this. I'll look into this, thanks!

src/durableorchestrationcontext.ts
```diff
@@ -571,46 +253,22 @@ export class DurableOrchestrationContext {
...
      * Restarts the orchestration by clearing its history.
      *
      * @param The JSON-serializable data to re-initialize the instance with.
      */
-    public continueAsNew(input: unknown): Task {
+    public continueAsNew(input: unknown): void {
         const newAction = new ContinueAsNewAction(input);
-
-        return TaskFactory.UncompletedTask(newAction);
+        if (this.taskOrchestratorExecutor !== undefined) {
```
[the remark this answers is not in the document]
davidmrdavid: I believe it's undefined during testing, mostly due to how old tests were implemented. I could try to refactor those old tests, but when first implementing this PR I thought it would be too much work. I can reconsider this decision if we think it is a big deal though :)

src/entities/responsemessage.ts
```diff
@@ -1,7 +1,25 @@
...
-    public result: string; // Result
+    public result?: string; // Result
     public exceptionType?: string; // ExceptionType
+
+    public constructor(event: unknown) {
+        if (typeof event === "object" && event !== null) {
+            if (Utils.hasStringProperty(event, "result")) {
+                this.result = event.result;
+            }
+            if (Utils.hasStringProperty(event, "exceptionType")) {
```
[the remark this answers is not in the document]
davidmrdavid: I guess that depends on how you define type-safe. In this case, `Utisl.hasStringProperty` actually makes heavy use of the TypeScript type system to propagate type information. 

But I see what you mean, this uses quite a lot of reflection. Let me think about this.

src/extensionSchema.ts
```diff
@@ -0,0 +1,5 @@
+/** @hidden */
+export enum ExtensionSchema {
+    V1,
+    V2,
```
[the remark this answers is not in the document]
davidmrdavid: They're documented when translating from the V2 format to V1. I'll move that documentation here.

src/orchestrator.ts
```diff
@@ -1,27 +1,18 @@
...
-
-/** @hidden */
-const log = debug("orchestrator");
+import { TaskOrchestrationExecutor } from "./taskorchestrationexecutor";
+import { UpperSchemaVersion } from "./replaySchema";
 
 /** @hidden */
 export class Orchestrator {
     private currentUtcDateTime: Date;
+    private taskOrchestrationExecutor: TaskOrchestrationExecutor;
```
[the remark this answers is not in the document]
davidmrdavid: If my memory serves me right, this caused some issues during testing. I'll try refactoring this again and seeing what happens.

src/replaySchema.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @hidden
+ * Supported OOProc DF extension protocols
+ */
+export enum UpperSchemaVersion {
+    V1 = 0,
+    V2 = 1,
```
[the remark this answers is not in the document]
davidmrdavid: Ah, seems `ExtensionSchema` is an old name for `UpperSchemaVersion`. It isn't being used anywhere. I'll delete it.

src/durableorchestrationcontext.ts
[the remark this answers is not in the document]
cgillum: Definitely not a big deal. A comment might be good, though, to help maintainers understand that this is not a normal scenario that needs special handling.

src/taskorchestrationexecutor.ts
davidmrdavid: We actually support non-generators in Python. For example, a trivial orchestration that just returns "Hello!" is a valid orchestration.

src/taskorchestrationexecutor.ts
davidmrdavid: Ah yes, I meant to do that. Thanks

src/taskorchestrationexecutor.ts
davidmrdavid: You mean to eagerly index into the array, and then check for undefined? Yeah we could do that, but I'm not particularly fond of how that coding style. Also, in my experience, it takes a lot of convincing to the TypeScript type-checker that a value is no longer `undefined`, so it can make the code harder to read.

Perhaps we can find a happy middle-ground by utilizing some custom utility. Let me think about this.

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,493 @@
...
+    private tryResumingUserCode(): void {
+        // If the current task does not have a result,
+        // then we cannot continue running the user code.
+        const currentTask: TaskBase = this.currentTask;
+        this.context.isReplaying = currentTask.isPlayed;
+        if (currentTask.stateObj === TaskState.Running) {
+            return;
+        }
+
+        // The feed in the result of the current task to the generator
```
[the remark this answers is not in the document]
davidmrdavid: Actually, I think this should have been "The" --> "We" :-) . I'll patch this.

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,493 @@
...
+                : this.generator.throw(taskValue);
+
+            if (generatorResult.done) {
+                // If the generator returned (via a `return` statement),
+                // then we capture the workflow's result result.
+                this.orchestratorReturned = true;
+                this.output = generatorResult.value;
+                return;
+            } else if (generatorResult.value instanceof Task) {
+                // The generator yielded another task.
```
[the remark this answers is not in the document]
davidmrdavid: I think we should. Will do.

src/taskorchestrationexecutor.ts
davidmrdavid: Yeah, it's not the right verb. I'll try with "can refer to"  or something like that.

src/tasks/internalTasks.ts
```diff
@@ -0,0 +1,369 @@
...
+        this.state = state;
+    }
+
+    /** Attempt to set a result for this task, and notifies parents, if any */
+    public SetValue(isError: boolean, value: unknown): void {
+        let newState: TaskState;
+
+        if (isError) {
+            if (value instanceof Error) {
+                if (value instanceof TaskBase && value.result instanceof Error) {
```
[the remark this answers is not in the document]
davidmrdavid: I think you might have mistread this :-).
`value` is a `TaskBase` (a kind of task) and `value.result` is the `Error`.
In other words, this is a task with an error as its result

src/taskorchestrationexecutor.ts
cgillum: Oh interesting. I guess if we wanted to also support that in Python we could just set `fn` as the orchestrator output value if it's not a generator.

src/tasks/internalTasks.ts
```diff
@@ -0,0 +1,369 @@
...
+    constructor() {
+        super(false, "noOp");
+    }
+}
+
+/**
+ * @hidden
+ * A task that should result in an Action being communicated to the DF extension.
+ */
+export class ProperTask extends TaskBase {
```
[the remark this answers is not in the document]
davidmrdavid: Yeah I need to work on the naming here. There's technically already a file defining user-facing tasks that we call "externalTasks.ts", so I need to find a way to make it fit. I'll commit a new version with a better naming scheme.

src/tasks/internalTasks.ts
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * @hidden
+     * Attempts to set a value to this task, given a completed sub-task
+     *
+     * @param child
+     *  The sub-task that just completed
+     */
+    public trySetValue(child: TaskBase): void {
+        if (this.state === TaskState.Running) {
+            this.SetValue(false, child);
```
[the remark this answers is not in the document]
davidmrdavid: Huh, I might be missing that case entirely. My bad, will patch.

src/tasks/internalTasks.ts
```diff
@@ -0,0 +1,369 @@
...
+
+    /**
+     * @hidden
+     * Attempts to set a value to this task, given a completed sub-task
+     *
+     * @param child
+     *  The sub-task that just completed
+     */
+    public trySetValue(child: TaskBase): void {
+        if (this.isWaitingOnTimer) {
```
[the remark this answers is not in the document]
davidmrdavid: I don't think `this.numAttempts > 1` would be equivalent. `isWaitingOnTimer` means that the next task we expect to complete is a Timer-task.

I can certainly rename this though, but I think `isWaitingOnTimer` is fairly descriptive. Let me think of an alternative name though.

src/tasks/internalTasks.ts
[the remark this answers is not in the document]
cgillum: I don't think I'm misreading it... line 81 is checking `value instanceof Error`, which contradicts your statement (and the next line) that `value` is a `TaskBase`.

src/tasks/internalTasks.ts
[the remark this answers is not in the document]
cgillum: Oh, I see. I completely misunderstood what this code is doing. I missed the fact that there's a distinction between waiting on a retry timer and actually retrying. I think you can pretty much disregard this comment and the other related comments.

src/tasks/internalTasks.ts
```diff
@@ -0,0 +1,369 @@
...
+            this.SetValue(false, child.result);
+        } else {
+            // If the sub-task failed, schedule timer to retry again.
+            // Since these sub-tasks are not user-managed, they are declared as internal tasks.
+            const rescheduledTask = new InternalOnlyTask();
+            rescheduledTask.parent = this;
+            this.children.push(rescheduledTask);
+            this.executor.trackOpenTask(rescheduledTask);
+            this.isWaitingOnTimer = true;
+            this.error = child.result;
```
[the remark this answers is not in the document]
cgillum: Disregard - I misunderstood what this code is doing.

src/tasks/internalTasks.ts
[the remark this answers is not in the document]
davidmrdavid: Ah you're right, I didn't see line 81, thanks! Let me review this whole block again, I'll commit a revised version.

test/unit/timertask-spec.ts
davidmrdavid: This was the result of some fancy auto-refactoring. I've removed the duplicate assertion.

src/tasks/internalTasks.ts
```diff
@@ -0,0 +1,369 @@
...
+     * @param child
+     *  The sub-task that just completed
+     */
+    public trySetValue(child: TaskBase): void {
+        if (this.isWaitingOnTimer) {
+            this.isWaitingOnTimer = false;
+
+            // If we're out of retry attempts, we can set the output value
+            // of this task to be that of the last error we encountered
+            if (this.numAttempts >= this.retryOptions.maxNumberOfAttempts) {
```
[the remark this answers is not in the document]
davidmrdavid: The ordering of statements is a bit confusing here, but I believe it's right.
Essentially, if we're on the 3rd attempt out of 3, **and** we're waiting on a timer, then we've already failed the 3rd attempt at re-scheduling the Task. In this state, we're just waiting for the timer to fire before deciding if we need to re-schedule the task again (and increase the attempt counter) or fail, as it's being done within this if-block.

src/tasks/internalTasks.ts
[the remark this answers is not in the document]
davidmrdavid: Looking back, this is on purpose. 
For a `Task` to have `isError = true`, it needs to contain within an Exception/`Error` datatype. However, `WhenAny` only contains `Task`s as its result, so technically it never errors out. The `isError` flag is used simply to determine if the result of the task should be fed in as a value, or as a raised exception to the generator code. 

I'll add a comment to the code to clarify this, but I believe the current logic is correct.

src/orchestrator.ts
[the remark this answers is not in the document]
davidmrdavid: I added a comment for this decision in the code. In short, our testing infrastructure currently doesn't play well with orchestration data being initialized in the constructor. This is the result of tests that were built to match the "old style" of replay.

src/entities/responsemessage.ts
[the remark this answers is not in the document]
davidmrdavid: I'll leave this as-is for now, since `Utils.hasStringProperty`  already makes heavy use of TypeScript's type safety features, as I mentioned. 

Unfortunately, since the input to `ResponseMessage`'s constructor is a task's result, there's no easy way to constraint that datatype. The only alternative would be to move these property checks elsewhere, but that'd be just moving the reflection checks around :)

src/durableorchestrationcontext.ts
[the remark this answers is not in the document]
davidmrdavid: I ended up removing this check, as apparently no tests broke without it

src/durableorchestrationcontext.ts
```diff
@@ -116,69 +104,21 @@ export class DurableOrchestrationContext {
...
-                );
-            } else {
-                return TaskFactory.SuccessfulTaskSet(tasks, completionIndex, results);
-            }
+    public Task = {
+        all: (tasks: Task[]): Task => {
+            const internalTasks = tasks.map((t) => t.internalTask);
+            const action = new WhenAllAction(internalTasks);
+            const innerTask = new WhenAllTask(internalTasks, action);
+            const task = new Task(innerTask);
```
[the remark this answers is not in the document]
davidmrdavid: My intent with this encapsulation was to prevent users from accessing SDK-internal Task-methods in their orchestration code. However, your comment made me realize that I can effectively do that by giving end-users a constrained task `interface` instead. 

I've rewritten the code to do just that.

src/durableorchestrationcontext.ts
```diff
@@ -116,69 +104,21 @@ export class DurableOrchestrationContext {
...
-                    new AggregatedError(errors)
-                );
-            } else {
-                return TaskFactory.SuccessfulTaskSet(tasks, completionIndex, results);
-            }
+    public Task = {
+        all: (tasks: Task[]): Task => {
+            const internalTasks = tasks.map((t) => t.internalTask);
+            const action = new WhenAllAction(internalTasks);
+            const innerTask = new WhenAllTask(internalTasks, action);
```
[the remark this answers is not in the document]
davidmrdavid: I mostly did this for consistency with the rest of the API implementations, which also require an `action` parameter. Given that this is the only place where Compound Task classes are instantiated, I think this is _ok_.

But I'm not immediately opposed to making an exception for WhenAny and WhenAll and instantiating their actions in their constructors. Let me know what you think!

src/actions/actiontype.ts
davidmrdavid: These comments can now be found above ^

src/replaySchema.ts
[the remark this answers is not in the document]
davidmrdavid: I unified those two enums into `ReplaySchema`

src/taskorchestrationexecutor.ts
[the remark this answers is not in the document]
davidmrdavid: Now we do that

src/taskorchestrationexecutor.ts
davidmrdavid: I tested this on a newly added unit test, and it seems the current implementation already supports non-generator orchestrations; which is exactly what we want!

I've removed this TODO comment

src/tasks/internalTasks.ts
```diff
@@ -0,0 +1,369 @@
+import { RetryOptions } from "..";
```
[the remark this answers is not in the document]
davidmrdavid: I'll add this as my last commit, to avoid cluttering this PR too much

davidmrdavid:
The current state of the code should have addressed all PR feedback with the exception of two things:
1. Adding the copyright header, which I'll do later to avoid over-cluttering this PR
2. Saving some cycles by eagerly indexing into our arrays, and then checking for `undefined`. I'm concerned about letting undefined values flow through the code, so my first instinct is to avoid this; unless we feel this will be a real performance concern.

@cgillum: I'd appreciate if you could make another pass over this PR, resolving your old comments and questions if they've been addressed. Thanks!

src/task.ts
```diff
@@ -0,0 +1,450 @@
...
+ *
+ * @example Cancel a timer
+ * ```javascript
+ * // calculate expiration date
+ * const timeoutTask = context.df.createTimer(expirationDate);
+ *
+ * // do some work
+ *
+ * if (!timeoutTask.isCompleted) {
+ *     timeoutTask.cancel();
```
cgillum, line 64: In my previous review, I suggested adding a comment explaining the importance of calling `cancel()` on the task and got a thumbs up. Just checking to see if you were still planning on doing that. Looks like there's another example below too.

src/task.ts
```diff
@@ -0,0 +1,450 @@
...
+        this.state = state;
+    }
+
+    /** Attempt to set a result for this task, and notifies parents, if any */
+    public setValue(isError: boolean, value: unknown): void {
+        let newState: TaskState;
+
+        if (isError) {
+            if (value instanceof Error) {
+                if (value instanceof TaskBase && value.result instanceof Error) {
```
cgillum, line 159: Looks like my comment from a previous iteration about the above two lines got missed in your refactor.  `value` cannot be both an instance of `Error` and an instance of `TaskBase` at the same time.

src/task.ts
```diff
@@ -0,0 +1,450 @@
...
+     * @param child
+     *  The sub-task that just completed
+     */
+    public trySetValue(child: TaskBase): void {
+        if (this.isWaitingOnTimer) {
+            this.isWaitingOnTimer = false;
+
+            // If we're out of retry attempts, we can set the output value
+            // of this task to be that of the last error we encountered
+            if (this.numAttempts >= this.retryOptions.maxNumberOfAttempts) {
```
cgillum, line 424: My previous comment seems to have gotten lost in the refactor. I think `>=` here needs to be `>`.

src/taskorchestrationexecutor.ts
```diff
@@ -0,0 +1,486 @@
...
+                const updateTask = function (): void {
+                    this.setTaskValue(event, isSuccess, idKey);
+                    return; // we return because the task is yet to be scheduled
+                };
+                this.deferredTasks[key] = updateTask.bind(this);
+                return;
+            }
+        } else {
+            throw Error(
+                `Task with ID ${key} could not be retrieved from due to its ID-key being of type ${typeof key}` +
```
cgillum, line 311: Do you need a `. ` at the end here?
```suggestion
                `Task with ID ${key} could not be retrieved from due to its ID-key being of type ${typeof key}. ` +
```

cgillum (commented):
Just a few more things (and a couple old ones too).

src/task.ts
davidmrdavid: Oh my, I had totally mean to fix this! A lot of things changed in my refactor, including the name of this file, so that must be why I missed this. My next commit includes the correct version (just checking for `instanceof` Error)

src/task.ts
davidmrdavid: I had meant to add this comment. My next commits adds it.

src/orchestratorstate.ts
```diff
@@ -7,11 +10,52 @@ export class OrchestratorState implements IOrchestratorState {
...
+     */
+    private flattenCompoundActions(actions: IAction[]): IAction[] {
+        let flatActions: IAction[] = [];
+        for (const action of actions) {
+            // Given any compound action
+            if (action instanceof WhenAllAction || action instanceof WhenAnyAction) {
+                // We obtain its inner actions as a flat array
+                const innerActionArr = this.flattenCompoundActions(action.compoundActions);
+                // we concatenate the inner actions to the flat array we're building
+                flatActions = flatActions.concat(innerActionArr);
```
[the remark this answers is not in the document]
davidmrdavid: Thanks for the heads up, and the interesting post. I've changed this to use `push`

src/orchestratorstate.ts
```diff
@@ -20,5 +64,30 @@ export class OrchestratorState implements IOrchestratorState {
...
+            const actions = this.actions[0];
+            const newActions: IAction[][] = [];
+            // guard against empty array, meaning no user actions were scheduled
+            if (actions !== undefined) {
+                for (const action of actions) {
+                    // Each action is represented as an array in V1
+                    let newEntry: IAction[] = [];
+                    if (action instanceof WhenAllAction || action instanceof WhenAnyAction) {
+                        const innerActionArr = this.flattenCompoundActions(action.compoundActions);
+                        newEntry = newEntry.concat(innerActionArr);
```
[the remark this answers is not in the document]
davidmrdavid: Same as above: will use `push` instead :)

src/task.ts
davidmrdavid: I had responded to your previous comment, but as you point out: the refactor makes that thread hard to find :)

Here's the response I wrote to your original comment:

...
The ordering of statements is a bit confusing here, but I believe it's right.
Essentially, if we're on the 3rd attempt out of 3, and we're waiting on a timer, then we've already failed the 3rd attempt at re-scheduling the Task. In this state, we're just waiting for the timer to fire before deciding if we need to re-schedule the task again (and increase the attempt counter) or fail, as it's being done within this if-block.

davidmrdavid: As a sanity check, I changed that check from `>=` to `>` and various legacy tests failed, which gives me some confidence that this logic is right. But I'm open to being challenged on this!

src/iorchestratorstate.ts
```diff
@@ -7,4 +8,6 @@ export interface IOrchestratorState {
     output: unknown;
     error?: string;
```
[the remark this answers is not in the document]
davidmrdavid: Tangentially related: I had meant to remove `replaySchema?` since we already have a `schemaVersion` field. My next commit will remove it.

As for your question: There's a slight technical difference here. With `error?`, we're saying that the underlying JSON representation of `IOrchestratorState` may not have an `error` *key*. 

Meanwhile, with `error: string | undefined`, we're guaranteeing that the underlying JSON representation will have the `error` *key*, but it may be `undefined`/null.

This matters for serialization purposes. For safety reasons, I'd rather not communicate any `error` fields to the extension when the orchestration did not error out :) . That's what `error?` achieves here. 

Hope that answers it! It's an interesting tidbit

src/taskorchestrationexecutor.ts
davidmrdavid: Fair enough, I'll make the change in all appropiate places (there's about 4 in this file) :) . My next commit includes them.

Tangentially, I wish TypeScript / JS had better support for this scenario - I find it really weird that there's no idiomatic way to safely index into an object without risking nullish types snealing into the code. But oh well, that's a concern for the TypeScript team, not for this PR.

I'm changing this for the performance-friendly alternative

src/durableorchestrationcontext.ts
```diff
@@ -191,36 +154,10 @@ export class DurableOrchestrationContext {
      * @returns A Durable Task that completes when the called activity
      * function completes or fails.
      */
-    public callActivity(name: string, input?: unknown): Task {
+    public callActivity(name: string, input?: unknown): ITask {
```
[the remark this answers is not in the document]
davidmrdavid: There was no reason other than to indicate, with the `I`-  prefix, that this type referred to an interface. 
After thinking about it - I'll be using `Task instead`. My next commit addresses this.

davidmrdavid:
My latest PR responds to the most recent round of feedback. My bad for missing a few of your previous comments, there was a lot to address :)

src/iorchestratorstate.ts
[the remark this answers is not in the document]
cgillum: Got it, makes sense!

src/task.ts
cgillum: Oh right, we did discuss this elsewhere. I really don't like the GitHub web UI for doing PRs - it's hard to keep track of feedback. 😣

Maybe I'm getting confused because there's actually two state variables at play, `isWaitingOnTimer` and `numAttempts`. The semantic meaning of `numAttempts` seems to change depending on the value of `isWaitingOnTimer`, which is really confusing. From a pure semantic standpoint, it's not correct to give up when attempt# == maxAttempts, thus I think most people who read this (and didn't also write the code) will be immediately confused. I wish it could be written in a way that was a bit more intuitive, but I trust you when you say it's correct and don't want to block on this.

cgillum (approved):
I'm good with this iteration. Remaining comments are optional/non-blocking.

src/task.ts
davidmrdavid: Thanks! I do want to make the code clear so I'll give it another go. My latest does a tiny refactor that should clarify the logic here. Please let me know what you think!

davidmrdavid:
Wohoo, we're finally ready to merge :tada:
Super excited to get this out!

_I'll add the copyright headers in a different PR, to avoid cluttering this one_

PR #336, opened 2022-04-08
Properly Convert all Timestamps from Srings to Date Types

hossam-nasr:
The function `getInstancesOf` is used by the `Orchestrator` to get the `DurableOrchestrationBindingInfo` instance on initialization, which includes the list of `HistoryEvent`s. The old implementation of `getInstanceOf` included a function, `parseTimestampsAsDates` to convert the `Timestamp` properties of the desired object from a `string` type to a `Date` type. However, this function was only converting the first level of properties. Since the `Timestamp` properties of `HistoryEvent`s live inside the `history` property of the `DurableOrchestrationBindingInfo` object, this function actually did not convert those `Timestamp` properties, and those remained as strings. This introduced a bug in which the `durableOrchestrationContext.currentUtcDateTime` was never being advanced, because [this line](https://github.com/Azure/azure-functions-durable-js/blob/f337aff952dcb683709b8158b76d24d3ea2aca62/src/taskorchestrationexecutor.ts#L176) in the `TaskOrchestrationExecutor` was comparing a `string` to a `Date`.

This PR updates the `parseTimestampsAsDates` function to a recursive implementation that converts all `Timestamp` properties of the object and its children objects as well. It also cleans up the `getInstancesOf` method a bit to make it slightly easier to understand :)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -18,7 +18,12 @@ export class Utils {
             const candidateObjects = Object.values(collection).filter((value) =>
                 this.hasAllPropertiesOf(value, typeInstance)
             );
+
+            // this recursive step ensures _all_ Timestamp properties are converted propertly
+            // For example, a payload can contain the history as a property, so if we want to
+            // parse each HistoryEvent's Timestamp, we need to traverse the payload recursively
             this.parseTimestampsAsDates(candidateObjects);
+
             return candidateObjects as T[];
         }
         return [];
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -19,7 +19,8 @@ export class Utils {
                 this.hasAllPropertiesOf(value, typeInstance)
             );
 
-            // this recursive step ensures _all_ Timestamp properties are converted propertly
+            // this recursive step ensures _all_ Timestamp properties are converted properly
+
             // For example, a payload can contain the history as a property, so if we want to
             // parse each HistoryEvent's Timestamp, we need to traverse the payload recursively
             this.parseTimestampsAsDates(candidateObjects);
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -14,11 +14,14 @@ export class Utils {
         collection: { [index: string]: unknown },
         typeInstance: T
     ): T[] {
-        return collection && typeInstance
-            ? (Object.keys(collection)
-                  .filter((key: string) => this.hasAllPropertiesOf(collection[key], typeInstance))
-                  .map((key: string) => this.parseTimestampsAsDates(collection[key])) as T[])
-            : [];
+        if (collection && typeInstance) {
+            const candidateObjects = Object.values(collection).filter((value) =>
+                this.hasAllPropertiesOf(value, typeInstance)
+            );
+            this.parseTimestampsAsDates(candidateObjects);
+            return candidateObjects as T[];
+        }
+        return [];
     }
 
     public static getHrMilliseconds(times: number[]): number {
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -46,16 +49,15 @@ export class Utils {
...
-        ) {
-            obj.Timestamp = new Date(obj.Timestamp);
+    public static parseTimestampsAsDates(obj: unknown): void {
+        if (typeof obj === "object" && obj != null) {
+            if (this.hasOwnProperty(obj, "Timestamp") && typeof obj.Timestamp === "string") {
+                obj.Timestamp = new Date(obj.Timestamp);
+            }
+            Object.values(obj).map((value) => {
+                this.parseTimestampsAsDates(value);
+            });
```
davidmrdavid, line 59: Just checking: is it `safe` to access for `.values` even when `obj` is not an array?

davidmrdavid (commented):
Looks great! Just leaving one question

src/utils.ts
hossam-nasr: Are you asking safe from a programming language perspective or in the sense that there may be `Timestamp` properties inside non-array objects that we don't want to mess with?

If it's the former, then yes, it's safe. I guess the main benefit of `Object.values` is to loop over values of a non-array object, otherwise you could loop over the elements of the array itself 😛 

If it's the latter, then I'm not sure. But is there a context in which we _want_ a `Timestamp` property to be a string over a Date?

davidmrdavid: I meant the former :) . I'm just trying to make sure we can always call `Object.values` without causing some exception, which sounds to be the case.

davidmrdavid, line 59: nit: can you add a comment here explaining why we're adding this recursive step? For instance, you could explain that the payload contains the history as JSON property, and so if we want to parse TimeStamps in the history we need to traverse the payload recursively, looking for all TimeStamp properties :)

davidmrdavid (approved):
Looks good to me! Just left a small nitpicky comment to add a comment explaining the change. After that, please squash-and-merge and then remove this branch. Thanks!

src/utils.ts
hossam-nasr: Yep! At least that's my understanding. If we call `Object.values` on an array, we get the same array back, and if we call it on an object it returns a list of values:

```node
> Object.values([1,2,3])
[ 1, 2, 3 ]
> Object.values({a:1, b:2, c:3})
[ 1, 2, 3 ]
```

I think the only problematic scenario would be calling `Object.values` on `null` or `undefined`, but that should be protected against here

hossam-nasr: Done :)

src/utils.ts
```diff
@@ -14,11 +14,19 @@ export class Utils {
...
-            ? (Object.keys(collection)
-                  .filter((key: string) => this.hasAllPropertiesOf(collection[key], typeInstance))
-                  .map((key: string) => this.parseTimestampsAsDates(collection[key])) as T[])
-            : [];
+        if (collection && typeInstance) {
+            const candidateObjects = Object.values(collection).filter((value) =>
+                this.hasAllPropertiesOf(value, typeInstance)
+            );
+
+            // this recursive step ensures _all_ Timestamp properties are converted propertly
```
davidmrdavid, line 22: small typo :P 
```suggestion
            // this recursive step ensures _all_ Timestamp properties are converted properly

```

davidmrdavid (approved):
🚢 it!

src/utils.ts
hossam-nasr: OOPs, thanks haha

Issue #284, opened 2021-08-29
callHttp does no polling


itpropro:
**Describe the bug**
I have two durable function runtimes running locally on different ports. I try to make a durable http call with `yield context.df.callHttp`.
Regarding to the docs, the orchestrator function should poll based on the return as long as 202 and a location header is returned. 
In my case, callhttp does no polling at all and returns the plain result of the request containing status code 202 and a location header:

```
[2021-08-29T20:17:24.596Z] {
[2021-08-29T20:17:24.597Z]   statusCode: 202,
[2021-08-29T20:17:24.597Z]   headers: {
[2021-08-29T20:17:24.598Z]     Date: 'Sun, 29 Aug 2021 20:17:24 GMT',
[2021-08-29T20:17:24.598Z]     Server: 'Kestrel',
[2021-08-29T20:17:24.599Z]     'Transfer-Encoding': 'chunked',
[2021-08-29T20:17:24.599Z] Executed 'Functions.UpdateFunction' (Succeeded, Id=85f4cbae-34da-4ae8-9f27-469041954a81, Duration=6ms)
[2021-08-29T20:17:24.599Z]     Location: 'http://localhost:7072/runtime/webhooks/durabletask/instances/53fccd02ea564e088b5b03c06520aaf3?taskHub=kametlicensingapihub&connection=DfStorageConnectionString&code=jD7lnp5vqS34wvyYTmSKUpKCWxZy5BXmuCLMPAYCsqGXgAYNUvdhDQ==',
[2021-08-29T20:17:24.600Z]     'Retry-After': '10'
[2021-08-29T20:17:24.601Z]   },
[2021-08-29T20:17:24.601Z]   content: '{\r\n' +
[2021-08-29T20:17:24.602Z]     '  "id": "53fccd02ea564e088b5b03c06520aaf3",\r\n' +
[2021-08-29T20:17:24.603Z]     '  "statusQueryGetUri": "http://localhost:7072/runtime/webhooks/durabletask/instances/53fccd02ea564e088b5b03c06520aaf3?taskHub=kametlicensingapihub&connection=DfStorageConnectionString&code=jD7lnp5vqS34wvyYTmSKUpKCWxZy5BXmuCLMPAYCsqGXgAYNUvdhDQ==",\r\n' +
[...]

PR #340, opened 2022-04-09
Support Long Timers ⏳📈🚀🚀🚀

### LINKED ISSUE #325: Support Long Timers
This PR is to support timers longer than 7 days using the [createTimer[(https://docs.microsoft.com/en-us/javascript/api/durable-functions/durableorchestrationcontext?view=azure-node-latest#durable-functions-durableorchestrationcontext-createtimer) API.

To do this, we'll need changes both at the extension-level, and at the DF SDK level.
At the extension level:
* We include in the SDK payload a new field indicating the current maximum time for an individual timer
* Given that we're modifying the SDK payload, we should increase the SDK protocol version number
In the SDK:
* We should increase the SDK protocol version number to support the new payload
* We need to modify our TimerTask state machine to fetch the right history event, the one corresponding to the "final sub-timer"

hossam-nasr:
Resolves #325. This PR leverages these changes in the extension: https://github.com/Azure/azure-functions-durable-extension/pull/2134 to allow JS orchestrators to have timers longer than 6 days. Internally, when user code calls `CreateTimer`, if it is more than 6 days, a `LongTimerTask`, which is essentially a `WhenAll` task of several small sub-timers, is created instead of an atomic `DFTimerTask`. 

Now, the JS SDK will become the first OOProc SDK to support long timers 🥳 This also unblocks #284

### DIFF
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -336,20 +336,23 @@ export class DurableOrchestrationContext {
     public createTimer(fireAt: Date): TimerTask {
         const timerAction = new CreateTimerAction(fireAt);
         const durationUntilFire = moment.duration(moment(fireAt).diff(this.currentUtcDateTime));
-        if (
-            this.schemaVersion >= ReplaySchema.V3 &&
-            this.maximumShortTimerDuration &&
-            this.longRunningTimerIntervalDuration &&
-            durationUntilFire > this.maximumShortTimerDuration
-        ) {
-            return new LongTimerTask(
-                false,
-                timerAction,
-                this,
-                this.taskOrchestratorExecutor,
-                this.maximumShortTimerDuration,
-                this.longRunningTimerIntervalDuration
-            );
+        if (this.schemaVersion >= ReplaySchema.V3) {
+            if (!this.maximumShortTimerDuration || !this.longRunningTimerIntervalDuration) {
+                throw Error(
+                    "Replay schema version >= V3 is being used, but one or more of the properties `maximumShortTimerDuration` and `longRunningTimerIntervalDuration` are not defined. This is likely an issue with the Extension."
+                );
+            }
+
+            if (durationUntilFire > this.maximumShortTimerDuration) {
+                return new LongTimerTask(
+                    false,
+                    timerAction,
+                    this,
+                    this.taskOrchestratorExecutor,
+                    this.maximumShortTimerDuration,
+                    this.longRunningTimerIntervalDuration
+                );
+            }
         }
 
         return new DFTimerTask(false, timerAction);
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -339,7 +339,8 @@ export class DurableOrchestrationContext {
         if (this.schemaVersion >= ReplaySchema.V3) {
             if (!this.maximumShortTimerDuration || !this.longRunningTimerIntervalDuration) {
                 throw Error(
-                    "Replay schema version >= V3 is being used, but one or more of the properties `maximumShortTimerDuration` and `longRunningTimerIntervalDuration` are not defined. This is likely an issue with the Extension."
+                    "A framework-internal error was detected: replay schema version >= V3 is being used, but one or more of the properties `maximumShortTimerDuration` and `longRunningTimerIntervalDuration` are not defined. This is likely an issue with the Durable Functions Extension. Please report this bug here: https://github.com/Azure/azure-functions-durable-js/issues "
+
                 );
             }
 
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -55,8 +55,12 @@ export class DurableOrchestrationContext {
         this.isReplaying = isReplaying;
         this.currentUtcDateTime = currentUtcDateTime;
         this.parentInstanceId = parentInstanceId;
-        this.longRunningTimerIntervalDuration = moment.duration(longRunningTimerIntervalDuration);
-        this.maximumShortTimerDuration = moment.duration(maximumShortTimerDuration);
+        this.longRunningTimerIntervalDuration = longRunningTimerIntervalDuration
+            ? moment.duration(longRunningTimerIntervalDuration)
+            : undefined;
+        this.maximumShortTimerDuration = maximumShortTimerDuration
+            ? moment.duration(maximumShortTimerDuration)
+            : undefined;
         this.schemaVersion = schemaVersion;
         this.input = input;
         this.newGuidCounter = 0;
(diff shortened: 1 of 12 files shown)


src/durableorchestrationcontext.ts
```diff
@@ -293,9 +334,25 @@ export class DurableOrchestrationContext {
...
+        ) {
+            return new LongTimerTask(
+                false,
+                timerAction,
+                this,
+                this.taskOrchestratorExecutor,
+                this.maximumShortTimerDuration,
+                this.longRunningTimerIntervalDuration
+            );
+        }
```
davidmrdavid, line 353: I think it'd be wise to protect against the case where ReplaySchema is V3 but the maximumShortTimerDuration and/or the longRunningTimerIntervalDuration are undefined.

Can we add a check here that throws an exception if that's the case? Otherwise, we risk things going wrong in the extension-level.

src/task.ts
```diff
@@ -377,6 +379,107 @@ export class WhenAnyTask extends CompoundTask {
...
+        id: TaskID,
+        action: CreateTimerAction,
+        orchestrationContext: DurableOrchestrationContext,
+        executor: TaskOrchestrationExecutor,
+        maximumTimerLength: moment.Duration,
+        longRunningTimerIntervalDuration: moment.Duration
+    ) {
+        const currentTime = moment(orchestrationContext.currentUtcDateTime);
+        const finalFireTime = moment(action.fireAt);
+        const durationTillFire = moment.duration(finalFireTime.diff(currentTime));
```
davidmrdavid, line 411: nit: just for style's sake, can we change `durationTillFire` to `durationUntilFire`? Just trying to avoid the abbreviations :)

Alternatively, I also like `timerLength` for this variable.

davidmrdavid (commented):
Good job, I think this is close to merge. Left some comments. My main blocker at this point is that I would prefer not having the timer metadata be exposed in the orchestration context class :-)

src/durableorchestrationcontext.ts
hossam-nasr: I was actually going to ask for feedback regarding this `if` statement. Should we even be checking for schema version? If the schema is less than `V3`, an exception would be thrown in the extension, regardless of whether we create a `LongTimerTask` or a `DFTimerTask`. 

Also, I was going to ask what should we do in the case that `maximumShortTimerDuration` is defined but not `longRunningTimerIntervalDuration`? Should we also complain/throw an exception? Just ignore and create a `DFTimerTask`? Default to using `maximumShortTimerDuration` as the `longRunningTimerIntervalDuration`?

src/task.ts
hossam-nasr: Done, removed the abbreviation :)

src/durableorchestrationcontext.ts
```diff
@@ -101,6 +114,34 @@ export class DurableOrchestrationContext {
...
+
+    /**
+     * Gets the current schema version that this execution is
+     * utilizing, based on negotiation with the extension.
+     *
+     * Different schema versions can allow different behavior.
+     * For example, long timers are only supported in schema version >=3
+     */
+    public schemaVersion: ReplaySchema;
+
```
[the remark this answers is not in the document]
hossam-nasr: I made them private and readonly

src/durableorchestrationcontext.ts
hossam-nasr: Add the check! On second thought, I actually think we need to wrap this in a check for the replay version if we want to throw this error. We don't want to complain about timer-related metadata not being set if schema version < 3 is being used anyway.

The alternative I guess would be to ignore this case entirely and let it error out in the extension, but I think throwing the error here is more helpful.

Also, would appreciate thoughts on the error message 😛

src/durableorchestrationcontext.ts
```diff
@@ -293,9 +334,28 @@ export class DurableOrchestrationContext {
...
     public createTimer(fireAt: Date): TimerTask {
-        const newAction = new CreateTimerAction(fireAt);
-        const task = new DFTimerTask(false, newAction);
-        return task;
+        const timerAction = new CreateTimerAction(fireAt);
+        const durationUntilFire = moment.duration(moment(fireAt).diff(this.currentUtcDateTime));
+        if (this.schemaVersion >= ReplaySchema.V3) {
+            if (!this.maximumShortTimerDuration || !this.longRunningTimerIntervalDuration) {
+                throw Error(
+                    "Replay schema version >= V3 is being used, but one or more of the properties `maximumShortTimerDuration` and `longRunningTimerIntervalDuration` are not defined. This is likely an issue with the Extension."
```
davidmrdavid, line 342: It would be good to include actionable feedback for errors like this. In this case, let's ask users to report the bug :)

```suggestion
                    "A framework-internal error was detected: replay schema version >= V3 is being used, but one or more of the properties `maximumShortTimerDuration` and `longRunningTimerIntervalDuration` are not defined. This is likely an issue with the Durable Functions Extension. Please report this bug here: https://github.com/Azure/azure-functions-durable-js/issues "

```

davidmrdavid: Additionally, it would be good if you could log the values for `maximumShortTimerDuration` and `longRuninngTimerIntervalDuration`. It's a bit verbose, but it will help us greatly if a bug is filed.

davidmrdavid (commented):
We're almost there!

src/durableorchestrationcontext.ts
davidmrdavid: Left some comments on the error message in my latest review!

src/durableorchestrationcontext.ts
hossam-nasr: I could log them, but they won't be the raw JSON values that we get from the extension. Those are parsed [here](https://github.com/Azure/azure-functions-durable-js/blob/e2f32c0a65821e53bc24c53b704e1ac3811f7edb/src/durableorchestrationcontext.ts#L58) into either a `moment.Duration` object, or `undefined`, so if we hit this line, it pretty much means they were `undefined`.

hossam-nasr: Although I guess it would help decipher which one was undefined, or if it was both

davidmrdavid: Yup, I suggested this to figure out which one was undefined :)

hossam-nasr: Added the logging

davidmrdavid (commented):
This looks pretty much ready to me. Just 2 more things:
1. I realized last night this was missing a test-case. Can you please add the appropiate long timer test cases using the following segment of timer tests as an example? https://github.com/Azure/azure-functions-durable-js/blob/3182ada64ddcd7acfeef9bf6b8eca34d6deeaf75/test/integration/orchestrator-spec.ts#L1536

For the above, you'll probably have to modify the `TestHistories` class to include a new history for long timers. That will be a little manual, but essentially you need to mock the expected History for a simple long timer scenario. My tip here is not to overcomplicate it, just mock the history for an orchestrator that schedules 1 long timer with, say, 2 sub-timers, and that's it.

As for how many long timer test cases you need, I'd say 2: one that tests that the orchestrator waits for the sub-timers, and another that tests that the orchestrator proceeds when all sub-timers finish.

2. Can you do a manual check testing that long timers can be cancelled successfully? Now that timers can be arbitrarily long, it is important to make sure they can be cancelled at-will. The easiest way to test for cancellable timers it to utilize the [human in the loop](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-overview?tabs=javascript#human) sample and have the activity in it always finish immediately. See the linked sample for more clarity. Just want to make sure this edge-case is at least manually tested as well. Thanks!

hossam-nasr (commented):
Added unit tests and also manually tested and verified that long timers get canceled :)

davidmrdavid (approved):
The time has come ... 
to merge this PR 🧙‍♂️ 

✨ _it looks good to me_ ✨

PR #344, opened 2022-04-12
Avoid using mutable `moment.add` method

hossam-nasr:
`Moment` objects are mutable, and the `moment().add()` method mutates the original moment object. This is sometimes counterintuitive and can lead to some nasty bugs. This PR avoid mutating any moment objects that are reused. Instead, we wrap Date objects in a new moment object before calling `add()`.

Thankfully, use of `moment.add` was limited to tests, and did not affect any test result (with the exception of one test). However, I still think that these uses of `moment.add` did not intend to use the mutability of the objects, and even when it did, it's cleaner to explicitly state that to avoid any misunderstandings :)

### DIFF
--- a/test/integration/orchestrator-spec.ts
+++ b/test/integration/orchestrator-spec.ts
@@ -818,7 +818,10 @@ describe("Orchestrator", () => {
                         actions: [
                             [new CallActivityWithRetryAction("Hello", retryOptions, "World")],
                         ],
-                        output: [startingTime, moment(startingTime).add(1, "m").toDate()],
+                        output: [
+                            startingTime,
+                            moment(startingTime).add(1, "m").add(30, "s").toDate(),
+                        ],
                         schemaVersion: ReplaySchema.V1,
                     },
                     true
--- a/test/integration/orchestrator-spec.ts
+++ b/test/integration/orchestrator-spec.ts
@@ -818,10 +818,7 @@ describe("Orchestrator", () => {
                         actions: [
                             [new CallActivityWithRetryAction("Hello", retryOptions, "World")],
                         ],
-                        output: [
-                            startingTime,
-                            moment(startingTime).add(1, "m").add(30, "s").toDate(),
-                        ],
+                        output: [startingTime, moment(startingTime).add(1, "m").toDate()],
                         schemaVersion: ReplaySchema.V1,
                     },
                     true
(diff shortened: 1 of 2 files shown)


test/integration/orchestrator-spec.ts
```diff
@@ -818,10 +818,7 @@ describe("Orchestrator", () => {
                         actions: [
                             [new CallActivityWithRetryAction("Hello", retryOptions, "World")],
                         ],
-                        output: [
-                            startingTime,
-                            moment(startingTime).add(1, "m").add(30, "s").toDate(),
-                        ],
+                        output: [startingTime, moment(startingTime).add(1, "m").toDate()],
```
davidmrdavid, line 821: As you point out, the previous usage of `moment` can be pretty unintuitive. Can you please remind me why this diff right here works? In the previous code, it seems we're adding 1 minute 30 seconds, but now we're just adding 1 minute. Where did the 30 seconds go?

test/testobjects/testhistories.ts
```diff
@@ -1313,127 +1293,148 @@ export class TestHistories {
...
                 isPlayed: false,
             }),
             new TimerFiredEvent({
                 eventId: -1,
-                timestamp: firstMoment.add(4, "s").add(retryInterval, "ms").toDate(),
+                timestamp,
                 isPlayed: false,
-                fireAt: firstMoment.add(4, "s").add(retryInterval, "ms").toDate(),
+                fireAt: timestamp,
                 timerId: 3,
```
davidmrdavid, line 1426: nitpick: It seems that the timers are firing _exactly_ at their `fireAt` value (value for `timestamp` is the same as `fireAt`), which doesn't usually match reality. In reality, the timestamp of a timer might be a few seconds after it was scheduled to fire.

Just for good practice, can you add some artificial delay in the `timestamp` value of timer fired events? I want our tests to catch errors when timers are "delayed" in their firing for whatever reason.

davidmrdavid (approved):
Looks almost right to me! Left a clarifying question, and an edit request.

test/testobjects/testhistories.ts
hossam-nasr: Done :)

test/integration/orchestrator-spec.ts
hossam-nasr: Because the original test's behavior was hard to understand and replicate, for this test in particular, I didn't try to replicate the old behavior, but rather I wrote a new history without using `moment.add()`, and in that history it just so happened that the output is 1 min after starting time not 1:30.

I wouldn't think of this diff as changing 1:30 to 1min, but rather rewriting this test, and the new expected value here is 1min.

That being said, when I added those artificial delays you talked about, I added an additional 30 seconds to remove this diff :P

test/testobjects/testhistories.ts
```diff
@@ -1352,7 +1352,8 @@ export class TestHistories {
             })
         );
 
-        timestamp = moment(timestamp).add(retryInterval, "ms").toDate();
+        let fireAt = moment(timestamp).add(retryInterval, "ms").toDate();
+        timestamp = moment(fireAt).add(10, "s").toDate();
```
davidmrdavid, line 1356: My only nitpick is to add a comment to these artificial delays explaining that you're adding these extra seconds because, in reality, timers don't fire exactly at the same time as their `fireAt` value.

davidmrdavid (approved):
LGTM conditional on adding the clarifying comment I list below.

test/testobjects/testhistories.ts
hossam-nasr: Done. 

Just FYI, there are other places where test histories are generated with timers, and those also have their `timestamp` the same as their `fireAt` value. It might be worthwhile to go and add artificial delays to those too, but I feel like that's outside the scope of this PR

PR #345, opened 2022-04-13
cancelled -> canceled

hossam-nasr:
Changing instances of `isCancelled` to `isCanceled` to match naming used by the extensions. Fixes a bug where actions were not being canceled, e.g. timers.


davidmrdavid (approved):
LGTM!

PR #347, opened 2022-04-20
"@types/node": "10.14.2" is no longer available

StrSimon:

PR #351, opened 2022-04-22
Remove unused dependencies

ejizba:
Related to https://github.com/Azure/azure-functions-durable-js/pull/347, I was inspired to go through the `dependencies` and clean them up. This is something every node package should regularly do, and we should strive to have as few dependencies as possible. I often look at a package's list of dependencies on npm when deciding if I should use it and my level of trust in the package is directly related to the number of dependencies. Each additional dependency increases the size of an app (affecting startup times, etc.) and increases the chances of security vulnerabilities reported on packages halfway down the dependency tree that are difficult to get rid of.

My basic process for cleaning these was to use VS Code's search functionality to find any references to these packages. I excluded the "node_modules" and "samples" folders because those would not be direct references that needed to be listed in the package.json. I will add a comment for each case

For the types packages, the only reason those _may_ be needed to be listed as a dependency is if you are explicitly referencing the packages in your exported types (aka the `*.d.ts` files). You might say "What if I accidentally export a type in the future, though?", but honestly you should be very intentional about the types you export because it is an official API and changing those types would usually constitute a breaking change. It is very difficult to align versions of npm packages, so exporting a type from another package would exponentially increase the chances you cause a breaking change to the api and I would almost never recommend doing that.


package.json
```diff
@@ -40,28 +40,25 @@
     },
     "dependencies": {
         "@azure/functions": "^1.2.3",
-        "@types/lodash": "^4.14.119",
-        "@types/node": "10.14.2",
-        "@types/uuid": "~3.4.4",
-        "@types/validator": "^9.4.3",
         "axios": "^0.21.1",
-        "commander": "~2.9.0",
```
ejizba, line 48: Removed completely - no references found
Screen Shot 2022-04-22 at 2 45 43 PM

package.json
```diff
@@ -40,28 +40,25 @@
...
         "@azure/functions": "^1.2.3",
-        "@types/lodash": "^4.14.119",
-        "@types/node": "10.14.2",
-        "@types/uuid": "~3.4.4",
-        "@types/validator": "^9.4.3",
         "axios": "^0.21.1",
-        "commander": "~2.9.0",
         "debug": "~2.6.9",
         "lodash": "^4.17.15",
-        "rimraf": "~2.5.4",
```
ejizba, line 51: Only one reference found in a package.json script. However, these scripts are never run in production so `rimraf` can safely be a dev dependency.
Screen Shot 2022-04-22 at 2 45 34 PM

package.json
```diff
@@ -40,28 +40,25 @@
     },
     "dependencies": {
         "@azure/functions": "^1.2.3",
-        "@types/lodash": "^4.14.119",
-        "@types/node": "10.14.2",
```
ejizba, line 44: This one is the most complicated. There is one reference to a Node.js type (NodeJS.Timer), but I still don't think that means we should list `@types/node` as a dependency. A few reasons:

1. The durable-functions package appears to support Node 10,12,14, & 16 so forcing users to use v10 would be very limiting and often incorrect. Users need to specify the version of node types they are actually running on in their own package.json, which will likely not be 10 because it's pretty old.
1. Most users will include `@types/node` in their own package.json, so it's not like the type for `NodeJS.Timer` will be missing
2. Long term I would recommend you try to remove this reference to improve you compatibility across Node versions, but realistically I doubt the `NodeJS.Timer` type has changed in forever

Screen Shot 2022-04-22 at 2 44 01 PM

package.json
```diff
@@ -40,28 +40,25 @@
     },
     "dependencies": {
         "@azure/functions": "^1.2.3",
-        "@types/lodash": "^4.14.119",
-        "@types/node": "10.14.2",
-        "@types/uuid": "~3.4.4",
-        "@types/validator": "^9.4.3",
```
ejizba, line 46: No references found in *.d.ts files - moved to a dev dependency
Screen Shot 2022-04-22 at 2 44 56 PM
.

package.json
```diff
@@ -40,28 +40,25 @@
     },
     "dependencies": {
         "@azure/functions": "^1.2.3",
-        "@types/lodash": "^4.14.119",
-        "@types/node": "10.14.2",
-        "@types/uuid": "~3.4.4",
```
ejizba, line 45: No references found in *.d.ts files - moved to a dev dependency. Only result was a comment in the code, not an actual reference
Screen Shot 2022-04-22 at 2 44 45 PM

package.json
```diff
@@ -40,28 +40,25 @@
     },
     "dependencies": {
         "@azure/functions": "^1.2.3",
-        "@types/lodash": "^4.14.119",
```
ejizba, line 43: No references found in *.d.ts file - moved to a dev dependency
Screen Shot 2022-04-22 at 2 44 17 PM

package.json
davidmrdavid: Don't we need some reference to `@types/node` to annotate the sleep method? Not saying that it's worth it for a single method, just trying to understand how this works is working.

davidmrdavid (commented):
Thanks @ejizba, really appreciate this, and I'm sure our users will appreciate it greatly as well. This one's long overdue.

I left a small comment below, just a small question on how the `NodeJS.Timer` annotation is working if we don't have the node types. Thanks!

package.json
ejizba: It depends what you mean by "need" or "work". Once the typescript files get compiled to javascript, the types don't matter at all so the underlying javascript will always work. The only thing they matter for is a user's build step and that will fail if users don't have `@types/node` in their package.json.

However, the build will pass if a user adds `@types/node` in their _own_ package.json, and my point was the vast majority of users have already done that. It would actually be very difficult to create a Node.js app that _didn't_ reference node types haha (as soon as you reference `path`, `http`, `fs`, `process`, `os` etc. you'll need the types). Even in the rare case they don't have the types, I think the error is pretty self explanatory (or at least, TypeScript users will hit this many times in their life and know how to fix):

```
node_modules/durable-functions/lib/src/utils.d.ts:1:23 - error TS2688: Cannot find type definition file for 'node'.
 
1 /// <reference types="node" />
```

davidmrdavid: I see. I do know that the types are erased at compile-time, but for our TS users I do feel a little strange not including a types dependency (like @types/node) whose exclusion will cause a compile type error.

But, I trust you when you say that TS users will know how to deal with this, as well as the fact that most users will get this dependency as soon as they import pretty much anything else. So I'll defer to your expertise with this one and be ok with removing it :)

davidmrdavid (approved):
It's got my thumbs-up, please squash and merge ahead.

ejizba:
Fyi, I still don't have permissions to merge. Can someone merge for me? Thanks

davidmrdavid:
Thanks for merging, Hossam.
@ejizba - I increased your contributor role, hopefully you'll be able to merge PRs moving forward. Thanks again for this PR

PR #354, opened 2022-04-27
Remove moment types from exported types

### LINKED ISSUE #353: TypeScript build fails due to missing moment types
# Repro steps

This only repros with the bits on the dev branch, not the released package. So first do this in the durable repo:
1. `npm install`
2. `npm run build`
3. `npm pack`

In a new folder:
1. Create a sample typescript function project
2. `npm install <path to the the tgz file created in the durable repo above>`
3. Add a durable function (honestly not sure this is necessary - just having the package in your project might be enough)
1. `npm run build`

> NOTE: The sample project cannot be in a sub folder of the durable repo. The way node.js works, it'll search all parent folders for a "node_modules" folder, so if `@types/moment` package exists in a parent folder you won't get the error.

# Expected

Build passes

# Actual

You should get build errors complaining about references to moment in a types file
```
node_modules/durable-functions/lib/src/task.d.ts:4:25 - error TS2307: Cannot find module 'moment' or its corresponding type declarations.

4 import moment = require("moment");
```

# Additional info

It would be nice if we could add some sort of test to the build pipeline that essentially replicates my repro steps from above. Until #352 is fixed, it will be easy to accidentally reference types from external packages like this

hossam-nasr:
Partially addresses #353 . Removes the `moment.Duration` type from the `LongTimerTask` constructor.

### DIFF
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -354,8 +354,8 @@ export class DurableOrchestrationContext {
                     timerAction,
                     this,
                     this.taskOrchestratorExecutor,
-                    this.maximumShortTimerDuration,
-                    this.longRunningTimerIntervalDuration
+                    this.maximumShortTimerDuration.toISOString(),
+                    this.longRunningTimerIntervalDuration.toISOString()
                 );
             }
         }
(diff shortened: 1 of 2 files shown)


davidmrdavid (approved):
🚢 it!

Issue #334, opened 2022-04-04
Metadata generation failed


letzfets:
**Describe the bug**
Trying to run the sample project

**Investigative information**

- Durable Functions extension version: 1.8.3
- durable-functions npm module version: ??
- Language (JavaScript/TypeScript) and version: -> sample
- Node.js version: v16.14.2
- dotnet version: 6.0.103

**To Reproduce**

1. func extensions install -p Microsoft.Azure.WebJobs.Extensions.DurableTask -v 1.8.32.

**Expected behavior**
install durabel functions extensions

**Actual behavior**
crashes with
````
...dataGenerator.targets(20,5): error : Metadata generation failed. [/<PATH-TO_REPO>/azure-functions-durable-js/samples/extensions.csproj]
````

**Comment by cgillum** [comment]
It looks like the samples in this repo are incredibly out of date. I recommend instead following the instructions here for getting started: https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode

@davidmrdavid FYI, we should look into updating all these samples. It looks like they haven't been touched in 2-3 years.

**Comment by letzfets** [comment]
True. The samples in [azure-functions-durable-extension](https://github.com/Azure/azure-functions-durable-extension/tree/dev/samples) seem a bit newer. But the issue tracker for _javascript_ refers to this repo.

And that one creates the same error anyways.

I'm trying to run the samples with **function core tools** from command line (without VS Code): do you have a link to a tutorial / sample / documentation for that as well? I know how to run functions - just want to extend my knowledge to durable functions.

**Comment by cgillum** [comment]
I see. I suspect the reason for the crash is that the extension version being referenced is really old or doesn't exist (not sure if there's a typo in your command above). The latest extension version is **2.6.1**.

However, in most cases this step shouldn't be needed. Durable Functions can be used with extension bundles, removing the need for an extensions.csproj file or the `func extensions install` command.

I couldn't find instructions for how to do this with just Core Tools alone, but I tried a few things locally on my box and put together some instructions. Give this a try:

```bash
# Create a new function app
func init --worker-runtime node --language javascript

# Add the core Durable Functions functions
func new --name "DurableFunctionsHttpStart" --template "Durable Functions HTTP starter"
func new --name "DurableFunctionsOrchestrator" --template "Durable Functions orchestrator"
func new --name "Hello" --template "Durable Functions activity"

# Install the Durable Functions SDK
npm install durable-functions
```

[...]

PR #355, opened 2022-04-27
Add a smoke test in build pipeline

### LINKED ISSUE #353: TypeScript build fails due to missing moment types

hossam-nasr:
Resolves #353 (per suggestion in additional info), related to #354 and #356.

I've never used Azure Pipelines (or even YAML files at all) before, so I'm not sure if this is the right or the most efficient way to do this, but at least it works 😛

## Description 

This new build job essentially replicates the repro steps in #353:
1. builds and packs `durable-functions` into a `.tgz` file
2. copies the `samples/UnitTesting` directory into a `test-app` directory, outside the durable SDK directory.
    - I chose to create the test app by copying the `samples/UnitTesting` directory, but if we make the whole directory runnable (see discussions here: https://github.com/Azure/azure-functions-durable-js/issues/334#issuecomment-1091919762 -- I've also created an issue here to track this: #356), then we could instead copy the whole samples directory:
         * I didn't want to create the test app from scratch (mostly because it seemed more complicated) and preferred to just copy from the samples.
         * This was the only sample that was a fully working function app not just a sample function
         * I discovered later there are benefits to copying our samples (see below).
3. npm installs the generated `.tgz` file from step 1 in the test app
4. runs `npm run build` on the test app 

## Validation
Before merging #354 into this branch, the build pipeline failed (as expected) with this error (see [here](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=66438&view=logs&j=d25dc844-164d-552e-f1e9-8647caa62262&t=578a2197-efdf-58e0-879d-404438bf2588)): 
```
node_modules/durable-functions/lib/src/task.d.ts(4,25): error TS2307: Cannot find module 'moment' or its corresponding type declarations.
```

After merging, the build pipeline succeeds (see [here](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=66439&view=logs&j=d25dc844-164d-552e-f1e9-8647caa62262))

## Benefits

Besides checking for these kinds of type exporting mistakes in the future, while testing out this new pipeline, I realized there are other kinds of errors that can also be detected by this addition! I have fixed some of them in this PR too:

- If we break a previous types contract, but is not evident from our own code. For example, all arguments to `DummyOrchestrationContext` were supposed to be optional, but I didn't realize that I broke this contract with #340. It didn't cause any errors because none of our own code relied on the fact that those arguments are optional. Meanwhile, the `samples/UnitTesting` directory inside the durable directory referenced types from the npm `durable-functions` module, not from the local build. So this was only revealed when doing this kind of test.
- If some of our sample code is not runnable (related to #356). For example, the `samples/UnitTesting` app didn't import the `mocha` module, even though it referenced it. This also wasn't detected before, because `mocha` _is_ imported by the root durable SDK directory.


azure-pipelines.yml
```diff
@@ -65,6 +65,32 @@ jobs:
...
+          - task: Bash@3
+            displayName: "Create test app"
+            inputs:
+                targetType: "inline"
+                script: "mkdir $(System.DefaultWorkingDirectory)/../test-app && cp -r $(System.DefaultWorkingDirectory)/samples/UnitTesting $(System.DefaultWorkingDirectory)/../test-app"
+          - task: Bash@3
+            displayName: "Build test app"
+            inputs:
+                targetType: "inline"
+                script: "cd $(System.DefaultWorkingDirectory)/../test-app/UnitTesting && npm install && npm uninstall durable-functions && pckg=$(ls $(System.DefaultWorkingDirectory)/*.tgz) && npm install $pckg && npm run build"
```
davidmrdavid, line 93: Should we make the value of `pckg` more concrete? So far it just grabs any `.tgz` file. Maybe we can give it the durable functions package prefix or some known substring that is a part of the generated `tgz` name to ensure we're grabbing the right payload

azure-pipelines.yml
```diff
@@ -65,6 +65,32 @@ jobs:
             displayName: "npm build and test"
           - script: npm run test:nolint
             displayName: "npm build and test (no linting)"
+    - job: TestExportedTypes
```
davidmrdavid, line 68: I think we could safely rename this to `SmokeTest` as it seems to test not just types, but any error that would show up when setting up a DF app.

davidmrdavid (commented):
Left some small suggestions. I'll defer to @ejizba to for the final approval here, but I'd love to hear your thought son my suggestions below.

azure-pipelines.yml
hossam-nasr: Makes sense, I added the `durable-functions-` prefix

azure-pipelines.yml
hossam-nasr: Yes! I knew `TestExportedTypes` wasn't the best name but I couldn't figure out what to name it. `SmokeTest` is what I was looking for!

davidmrdavid (approved):
LGTM!

samples/UnitTesting/package.json
```diff
@@ -19,6 +20,7 @@
   },
   "devDependencies": {
```
ejizba, line 21: > I didn't want to create the test app from scratch (mostly because it seemed more complicated) and preferred to just copy from the samples.

My preference is to have a separate test app in the `test` folder. I think a sample app will always have a different purpose than a test app and so we will run into conflicting goals pretty quickly. For example, for the purposes of https://github.com/Azure/azure-functions-durable-js/issues/353, we don't want _any_ `@types/` packages listed in the package.json otherwise it could hide a missing dependency error in `durable-functions`. But for a sample app you _would_ want to have `@types/` packages because that's what all users want in real life

When I think of a "smoke test", I think of a simple hello world app. I would suggest you use VS Code or core tools to create a project and add all three of the durable function triggers to create the simplest "working" app possible. Here's a tutorial if you want to follow exactly what a user would likely do: https://docs.microsoft.com/azure/azure-functions/durable/quickstart-js-vscode (Just use TypeScript instead of JavaScript and you can skip everything after you create the projects/functions)

azure-pipelines.yml
```diff
@@ -65,6 +65,32 @@ jobs:
             displayName: "npm build and test"
           - script: npm run test:nolint
             displayName: "npm build and test (no linting)"
+    - job: SmokeTest
```
ejizba, line 68: I would prefer to include this smoke test in the above "UnitTests" job (You can rename it to just "Test" if you want). Then you get the benefit of testing the whole matrix and it simplifies the yaml because it gets rid of some duplicate steps

azure-pipelines.yml
```diff
@@ -65,6 +65,32 @@ jobs:
...
+            displayName: "Install Node.js Dependencies"
+          - script: npm run build
+            displayName: "Build project"
+          - script: npm pack
+            displayName: "Pack npm package"
+          - task: Bash@3
+            displayName: "Create test app"
+            inputs:
+                targetType: "inline"
+                script: "mkdir $(System.DefaultWorkingDirectory)/../test-app && cp -r $(System.DefaultWorkingDirectory)/samples/UnitTesting $(System.DefaultWorkingDirectory)/../test-app"
```
ejizba, line 88: Try to use the built in tasks when possible instead of a script. That way it works better on all OS's. There's examples of the npm task above and there's a copy files task with an example here: https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/azure-pipelines/build.yml#L76

Ping me offline and I can give you some tips on authoring/testing these yaml files if you want. Azure DevOps has some UI for editing pipelines that frankly isn't that intuitive, but is a lot easier than just modifying the yaml and hoping for the best.

samples/UnitTesting/package.json
hossam-nasr: Created a new test app in the test folder and using that for the smoke test now

azure-pipelines.yml
hossam-nasr: Switched to built-in tasks! It took me a while to get it to work on all platforms (see the number of commits :p) but should be good to go now!

azure-pipelines.yml
hossam-nasr: Done

test/test-app/package.json
```diff
@@ -0,0 +1,16 @@
+{
```
ejizba, line 1: Can you add a simple README.md to the "test-app" folder explaining what this app is used for? Honestly just pointing to https://github.com/Azure/azure-functions-durable-js/issues/353 might be enough

test/test-app/package.json
```diff
@@ -0,0 +1,16 @@
...
+    "name": "test-app",
+    "version": "1.0.0",
```
ejizba, line 3: > Before merging https://github.com/Azure/azure-functions-durable-js/pull/354 into this branch, the build pipeline failed (as expected) with this error (see [here](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=66438&view=logs&j=d25dc844-164d-552e-f1e9-8647caa62262&t=578a2197-efdf-58e0-879d-404438bf2588)):

Can you verify you did this for your final changes as well? It appears the link is for the first set of changes and just wanna make sure this still causes a build failure when expected

ejizba (approved):
Looks good!

test/test-app/package.json
hossam-nasr: Done

test/test-app/package.json
hossam-nasr: Did this, see the `hossamnasr/build-pipeline-test-2` branch: https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=71235&view=results.

PR #375, opened 2022-08-02
Update ts and pass audit

davidmrdavid:
The CI has been failing with Node16 apparently due to some unmet dependency of ts-doc, which required a newer version of the typescript dev-dependency.

This PR updates our typescript dev-dependency, and also performs an `npm audit fix`


davidmrdavid:
@hossam-nasr: FYI

Try running a CI on `dev` and you'll see that it fails due to some conflict on Node16.  The step that's failing is `npm ci`.
This PR fixes that step, but now we fail at the building/testing stage

ejizba (approved):

PR #380, opened 2022-08-30
Make entities use async/await only

### LINKED ISSUE #373: Error: Choose either to return a promise or call 'done'. Do not use both in your script.
**Describe the bug**
The following error log is shown in the local console when running and executing durable entity functions:

```
Error: Choose either to return a promise or call 'done'. Do not use both in your script. Learn more: https://go.microsoft.com/fwlink/?linkid=2097909
```

**Investigative information**

```
Azure Functions Core Tools
Core Tools Version:       4.0.4590 Commit hash: N/A  (64-bit)
Function Runtime Version: 4.5.2.18383
```

```
node -v
v16.15.0
npm -v
8.12.2
```

**To Reproduce**

function.json:
```json
{
  "bindings": [
    {
      "name": "context",
      "type": "entityTrigger",
      "direction": "in"
    }
  ],
  "scriptFile": "../dist/entity-name/entity.js"
}
```

function.ts:
```ts
export default entity<EntityType>(context => {
   // handleOperation(context)
});
```

**Expected behavior**
No error should be logged.

**Actual behavior**
An error is logged.

**Screenshots**
image

**Additional context**

This seems to be related to how the `entity()` function and `Entity` class are implemented internally (?).

- [Here](https://github.com/Azure/azure-functions-durable-js/blob/dev/src/shim.ts#L35-L37) a promise is returned.
- [Here](https://github.com/Azure/azure-functions-durable-js/blob/dev/src/entity.ts#L70) `context.done(...)` is called.

hossam-nasr:
Fixes #373. The `df.entity()` function before was returning an async function that also calls `context.done()`, causing the error:

```
Error: Choose either to return a promise or call 'done'. Do not use both in your script. Learn more: https://go.microsoft.com/fwlink/?linkid=2097909
```

To be thrown, which was hidden from logs before, but was surfaced after a node worker bug was fixed in https://github.com/Azure/azure-functions-nodejs-worker/pull/555. 

This PR removes the `context.done()` call and replaces it with the return value of the function. Potentially could be considred a "breaking" change since it changes the type signature of the customer-facing `df.entity()` function

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -20,11 +20,11 @@ const log = debug("orchestrator");
 export class Entity<T> {
     constructor(public fn: (context: IEntityFunctionContext<T>) => void) {}
 
-    public listen(): (context: IEntityFunctionContext<T>) => Promise<void> {
+    public listen(): (context: IEntityFunctionContext<T>) => Promise<EntityState> {
         return this.handle.bind(this);
     }
 
-    private async handle(context: IEntityFunctionContext<T>): Promise<void> {
+    private async handle(context: IEntityFunctionContext<T>): Promise<EntityState> {
         const entityBinding = Utils.getInstancesOf<DurableEntityBindingInfo>(
             context.bindings,
             new DurableEntityBindingInfoReqFields(
@@ -67,7 +67,7 @@ export class Entity<T> {
             }
         }
 
-        context.done(null, returnState);
+        return returnState;
     }
 
     private getCurrentDurableEntityContext(
--- a/src/shim.ts
+++ b/src/shim.ts
@@ -1,5 +1,6 @@
 import {
     Entity,
+    EntityState,
     IEntityFunctionContext,
     IOrchestrationFunctionContext,
     Orchestrator,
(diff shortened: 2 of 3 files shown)


src/entity.ts
```diff
@@ -67,7 +67,7 @@ export class Entity<T> {
             }
         }
 
-        context.done(null, returnState);
+        return returnState;
```
ejizba, line 70: I'm leaning towards this _would_ be breaking. Since we have an upcoming breaking change release soon (for the new programming model), I'd prefer to postpone fixing until then.

I have not been able to get an entity function to work locally yet, but I believe previously you could ignore the result of the function and it would still work because of this `context.done`. Like this:
```typescript
const myEntityFunc = df.entity(function* (context) {
    // do stuff
});

export default (context) => {
    myEntityFunc(context); // note no `return` here
};
```

I don't know why you'd wrap `myEntityFunc` like that, but doesn't really matter from a breaking change perspective.

ejizba (changes requested):

src/entity.ts
hossam-nasr: Yes you're right. I am able to run entities locally, and I was able to verify that the pattern you had above does work normally after this change but will stop working after it. Is there a way around that while still not using `context.done()` or are we okay with breaking this scenario?

P.S. not sure if that's why you couldn't get entities to work locally for you, but `df.entity()` takes a normal function as an argument not a generator like `df.orchestrator()`

ejizba: >  Is there a way around that while still not using context.done() or are we okay with breaking this scenario?

I don't think there is a way without `context.done()`. And no, we should not do breaking changes to existing functionality without a major version release. It's worth noting that this bug is only about cleaning up the logs - the existing error doesn't actually block the function from working.

> P.S. not sure if that's why you couldn't get entities to work locally for you, but df.entity() takes a normal function as an argument not a generator like df.orchestrator():

Yes that was my problem, thanks!

hossam-nasr:
Closing this in favor of #373 and #397

Issue #381, opened 2022-08-30
Unhelpful error message if error is thrown before work is scheduled


hossam-nasr:
**Describe the bug**
If an error is thrown inside an orchestrator function before `yield`ing any `Task`s, then 2 things happen:

1. No error is thrown in Core Tools
2. Navigating to the `statusQueryGetUri` will output an unhelpful failure message as below:

```
"runtimeStatus":"Failed","input":null,"customStatus":null,"output":"With OOProc schema V2, expected actions array to be of length 1 in outer layer but got size: 0"
```

By contrast, if the error is thrown inside the client directly, an error is thrown in core tools, with the expected error message rather than the one above.

**To Reproduce**

1. Create a dummy orchestration function that immediately throws an error as below:

```TS
const orchestrator = df.orchestrator(function* (context) {
    throw new Error("test")
});
```

2. Create a client to call your dummy orchestration function
3. Try running your app locally and running the orchestration
4. No error is thrown in Core Tools and navigating to the `statusQueryGetUri` will result in an output of `With OOProc schema V2, expected actions array to be of length 1 in outer layer but got size: 0` rather than my custom error message `test`

**Expected behavior**
An error is thrown in Core Tools and the output in `statusQueryGetUri` matches the error message thrown in the orchestration

**Actual behavior**
We see the above error message instead

PR #383, opened 2022-09-01
Add signalEntity API to orchestrations

### LINKED ISSUE #327: Support SignalEntity API
This allows orchestrations to send messages to Entities without needing to `yield` / `await` on an API. The implementation should be similar to `ContinueAsNew`, which is also a fire-and-forget operation.

hossam-nasr:
Resolves #327 . Allows orchestrations to have a "fire-and-forget" `signalEntity` operation, in addition to the existing blocking `callEntity` API.

## Design decisions/questions

Some design decisions made in this PR which could require additional discussions:

### An error encountered after a `signalEntity` call
The entity is still signaled if an error is encountered after the `signalEntity` call.

Code like the below:

```TS
const orchestrator = df.orchestrator(function* (context) {
    const id:string = context.df.getInput()
    const entityId = new df.EntityId("Counter", id);
    context.df.signalEntity(entityId, "add", 1)
    throw new Error("test");
});
```

Will result in:

* The orchestrator fails with an error message "test"
* The entity `Counter` is signaled once

### Yielding a `signalEntity` call
`signalEntity` is a fire-and-forget operation, and therefore it doesn't return a `Task` object and cannot be `yield`ed. Attempting to yield `signalEntity` call will cause your orchestration to fail with an error, but the signal call will still go through.

Code as the following:

```TS
const orchestrator = df.orchestrator(function* (context) {
    const id:string = context.df.getInput()
    const entityId = new df.EntityId("Counter", id);
    yield context.df.signalEntity(entityId, "add", 1)
});
```

Will result in:
* An error thrown in Core Tools (if running locally) with a message like the below:

```
Orchestration yielded data of type undefined. Only Task types can be yielded.Please refactor your orchestration to yield only Tasks.
```

* Orchestration status failed with the output message similar to the one above
* The `Counter` entity is still signaled once

### DIFF
--- a/src/taskorchestrationexecutor.ts
+++ b/src/taskorchestrationexecutor.ts
@@ -400,10 +400,17 @@ export class TaskOrchestrationExecutor {
                 newTask = generatorResult.value;
             } else {
                 // non-task was yielded. This isn't supported
-                throw Error(
-                    `Orchestration yielded data of type ${typeof generatorResult.value}. Only Task types can be yielded.` +
-                        "Please refactor your orchestration to yield only Tasks."
-                );
+                let errorMsg = `Durable Functions programming constraint violation: Orchestration yielded data of type ${typeof generatorResult.value}.`;
+                errorMsg +=
+                    typeof generatorResult.value === "undefined"
+                        ? ' This is likely a result of yielding a "fire-and-forget API" such as signalEntity or continueAsNew.' +
+                          " These APIs should not be yielded as they are not blocking operations. Please remove the yield statement preceding those invocations." +
+                          " If you are not calling those APIs, p"
+                        : " Only Task types can be yielded. P";
+                errorMsg +=
+                    "lease check your yield statements to make sure you only yield Task types resulting from calling Durable Functions APIs.";
+
+                throw Error(errorMsg);
             }
         } catch (exception) {
             // The generator threw an exception
--- a/src/taskorchestrationexecutor.ts
+++ b/src/taskorchestrationexecutor.ts
@@ -443,6 +443,13 @@ export class TaskOrchestrationExecutor {
         }
     }
 
+    public recordFireAndForgetAction(action: IAction): void {
+        if (!this.willContinueAsNew) {
+            this.addToActions(action);
+            this.sequenceNumber++;
+        }
+    }
+
     /**
      * @hidden
      * Tracks this task as waiting for completion.
--- a/src/actions/actiontype.ts
+++ b/src/actions/actiontype.ts
@@ -17,8 +17,8 @@ export enum ActionType {
     WaitForExternalEvent = 6,
     CallEntity = 7,
     CallHttp = 8,
-    // ActionType 9 and 10 correspond to SignalEntity and ScheduledSignalEntity
-    // Those two are not supported yet.
+    SignalEntity = 9,
+    // ActionType 10 corresponds to ScheduledSignalEntity, which is not supported yet
     WhenAny = 11,
     WhenAll = 12,
 }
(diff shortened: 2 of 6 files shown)


hossam-nasr:
@davidmrdavid ping

src/taskorchestrationexecutor.ts
```diff
@@ -113,7 +113,7 @@ export class TaskOrchestrationExecutor {
         }
 
         // Construct current orchestration state
-        const actions: IAction[][] = this.actions.length == 0 ? [] : [this.actions];
+        const actions: IAction[][] = this.actions.length == 0 ? [[]] : [this.actions];
```
davidmrdavid, line 116: Why do we need this change?

davidmrdavid (commented):
LGTM but, before approving, I'd like to understand the change below:

src/taskorchestrationexecutor.ts
davidmrdavid: @hossam-nasr: Friendly ping on this.

davidmrdavid:
@hossam-nasr:

I agree with design decision (1).

For design decision (2), I think we have room to improve the error message. In theory, I would like if the error message could say something like: 

"Durable Functions programmign constraint violation: You seem to have `yield`'ed signalEntity, which is a "fire-and-forget API". These APIs should not be `yielded` as they are not blocking operations. To avoid this exception, please remove the `yield` preceding the signalEntity invocation."

Obviously, this is quite involved, but I think it will save us the trouble of having to explain this default exception in the long run. You should be able to detect this if you make `signalEntity` return a special kind of object, say a "FireAndForgetTask" and for you to check against this type in the yield handler.

Please let me know what you think!

src/taskorchestrationexecutor.ts
hossam-nasr: This should've been a fix for #381 that was already merged in #382. I'm confused why it's still showing up in the diff here tbh

hossam-nasr:
@davidmrdavid I get the sentiment that the error message isn't the most detailed or helpful, but it feels like it goes against the spirit of having it be a "fire and forget" operation to have it return anything at all. By that same argument, we could have it just return a dummy task that _can_ be yielded, and just silently ignoring it in the yield handler, than throwing any error to begin with. Perhaps, however, as a compromise, we can make that generic error message more descriptive? For example, we can detect if the type of the object the user is trying to yield is `undefined`, and warn that this is likely because they are trying to yield a fire-and-forget operation, such as `continueAsNew()` or `signalEntity()`?

hossam-nasr:
@davidmrdavid I get what you are saying, but what I was trying to say is that having `context.df.signalEntity` return anything at all may in and of itself be violating this pattern we're trying to encourage, or sending the wrong message so to speak. For example, if a user types:

```TS
   const myTask = context.df.signalEntity(/* whatever */)
   console.log(typeof myTask)
```

You'd expect this to print `undefined` if it was a fire-and-forget operation, but now the type of `myTask` would be an object, which might send the wrong message. Does that make sense?

davidmrdavid:
@hossam-nasr. Yep, it does. I'm ok with returning nothing for that API so long as we generalize the error message that gets thrown today to include the possibility that the error is due to `yielding` a fire-and-forget API. Let me sync with you internally just to make sure we're on the same page.

hossam-nasr:
@davidmrdavid Done. I update the error message so that when the type attempted to be yielded is undefined, the error message is:

Durable Functions programming constraint violation: Orchestration yielded data of type undefined. This is likely a result of yielding a "fire-and-forget API" such as signalEntity or continueAsNew. These APIs should not be yielded as they are not blocking operations. Please remove the yield statement preceding those invocations. If you are not calling those APIs, please check your yield statements to make sure you only yield Task types resulting from calling Durable Functions APIs.

If it's any other type, the error message is:

Durable Functions programming constraint violation: Orchestration yielded data of type {type}. Only Task types can be yielded. Please check your yield statements to make sure you only yield Task types resulting from calling Durable Functions APIs.

davidmrdavid (approved):
:shipit:

Issue #384, opened 2022-09-02
Support the new Node.js programming model in durable


ejizba:
More info here: https://aka.ms/AzFuncNodeV4

Here are my recommended steps to do this in an incremental fashion:
1. This will be a breaking change, there's no way around that. So to start, I'd create a new branch, maybe "v3.x" and update the version of this package to "3.0.0-alpha.1"
2. I would update the reference of "@azure/functions" in the `package.json` to "4.0.0-alpha.2"
3. I would fix all the build errors (there will be a lot lol) and submit a PR
4. Now, I would try to get it to actually run locally 😅 That means following steps [here](https://aka.ms/AzFuncNodeV4) and using `app.generic` to register durable functions. Once it works locally, I'd submit another PR as I doubt the changes from step 3 will work on the first attempt
5. Finally, I would add fancy intellisense for `app.durable` methods. Currently tracked by [this issue](https://github.com/Azure/azure-functions-nodejs-library/issues/8), but open for discussion where that code lands. Regardless, ideally this is a separate PR

hossam-nasr:
First alpha v3 release of the SDK with programming model support is out now! 🎉

For further work on the programming model, discussions and improvements, closing this in favor of #393, #394, #414, #415, #416, #417, #418, #419, #420, #421

PR #389, opened 2022-09-27
Add basic support for durable triggers in new programming model

hossam-nasr:
Partially addresses #384 , corresponding to step 4. This PR allows registering of durable triggers in the new programming model using `app.generic()` interface, without any fancy wrappers around it in the meantime. Here is an example of a working Durable app using this build of the Durable SDK:

```TS
import { app, HttpRequest, input, InvocationContext, trigger } from "@azure/functions";
import * as df from "durable-functions";

app.generic('DurableFunctionsOrchestratorJS', {
    trigger: trigger.generic({
        type: 'orchestrationTrigger'
    }),
    handler: df.orchestrator(function* (context) {
        const outputs = [];
        outputs.push(yield context.df.callActivity("Hello", "Tokyo"));
        outputs.push(yield context.df.callActivity("Hello", "Seattle"));
        outputs.push(yield context.df.callActivity("Hello", "Cairo"));
        
        return outputs;
    })
})

app.generic('Hello', {
    trigger: trigger.generic({
        type: 'activityTrigger',
    }),
    handler: async function(context: InvocationContext, name: string): Promise<string> {
        return `Hello ${name}`
    }
})

const clientInput = input.generic({
    type: 'orchestrationClient'
});

app.generic('DurableFunctionsHttpStart', {
    trigger: trigger.generic({
        type: 'httpTrigger',
        authLevel: 'function',
        route: 'orchestrators/{functionName}',
        methods: [
            'post', 'get'
        ],
    }),
    extraInputs: [clientInput],
    return: {name: "$return", type:'http'},
    handler: async function (context: InvocationContext, req: HttpRequest): Promise<any> {
        const client = df.getClient(context, clientInput);
        const instanceId = await client.startNew(req.params.functionName, undefined, req.body);
    
        return client.createCheckStatusResponse(req, instanceId);
    }
})
```

For now, this PR fixes and amends tests for orchestrations and entities to be compliant with the new programming model, but I've disabled tests for `getClient` until the new public-facing API is finalized. Also, the PR references a local build of the Node framework because support for passing in a return value as-is to the extension is needed, and is currently unavailable in a public release of the Framework.

I wanted to split up the new programming model work into multiple PRs to avoid one large unwieldy PR, which this PR is already becoming. But, I've added an issue here: #392 to track updating the `@azure/functions` reference to a public release once one is available, and #393 to update and re-enable `getClient` tests.


src/durableorchestrationclient.ts
```diff
@@ -46,8 +47,14 @@ const URL = url.URL;
  * };
  * ```
  */
-export function getClient(context: unknown): DurableOrchestrationClient {
-    let clientData = getClientData(context as IOrchestrationFunctionContext);
+export function getClient(
+    context: IOrchestrationFunctionContext,
+    clientInputOptions: ClientInput
+): DurableOrchestrationClient {
+    if (!(context instanceof InvocationContext)) {
```
ejizba, line 54: If `IOrchestrationFunctionContext` already extends `InvocationContext`, this check seems unnecessary

Similar comment for line 71 `isClientInput` below

src/testingUtils.ts
```diff
@@ -71,20 +69,21 @@ export class DummyOrchestrationContext implements IOrchestrationFunctionContext
         // Set this as undefined, let it be initialized by the orchestrator
         this.df = (undefined as unknown) as DurableOrchestrationContext;
     }
-    public doneValue: IOrchestratorState | undefined;
-    public err: string | Error | null | undefined;
+    functionName: string;
+    extraInputs: InvocationContextExtraInputs;
+    extraOutputs: InvocationContextExtraOutputs;
```
ejizba, line 74: Would be nice if `DummyOrchestrationContext` extended `InvocationContext` instead of copy/pasting all this stuff

ejizba (changes requested):
Generally looks to be on the right track - clean it up and turn it into a non-draft PR and I'll take another pass (fix merge conflicts, get a build passing, etc.)

src/durableorchestrationclient.ts
hossam-nasr: The reason I added those checks here is because `getClient` is a customer-facing API, so even though the type here is `IOrchestrationFunctionContext` and `ClientInput`, there is no guarantee that the customer actually supplies input with those types

test/unit/getclient-spec.ts
```diff
@@ -11,115 +11,97 @@ const expect = chai.expect;
...
-                },
-            };
-
-            const client = getClient(badContext);
-
-            const expectedUniqueWebhookOrigins: string[] = [Constants.DefaultLocalOrigin];
-            expect(client.uniqueWebhookOrigins).to.deep.equal(expectedUniqueWebhookOrigins);
-        });
-    });
+    // const defaultTaskHub = "TestTaskHub";
```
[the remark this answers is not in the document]
hossam-nasr: I commented them out because currently I get build errors for bad types when I uncomment them. Fixing the types would entail about as much work as fixing the tests altogether, which I'd rather do in a separate PR as this one is already getting too big

types/index.d.ts
```diff
@@ -0,0 +1,5 @@
+import { FunctionInput } from "@azure/functions";
+
+export interface ClientInput extends FunctionInput {
```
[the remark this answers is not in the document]
hossam-nasr: Updated to `OrchestrationClientInput`

src/testingUtils.ts
hossam-nasr: Done!

ejizba:
> I wanted to split up the new programming model work into multiple PRs to avoid one large unwieldy PR, which this PR is already becoming. 

I think the current PR size is perfectly fine - no need to worry about being "unwieldy" yet

> I've added an issue here: https://github.com/Azure/azure-functions-durable-js/issues/392 to track updating the @azure/functions reference to a public release once one is available

I think the CI build for this PR needs to pass before merging, so I don't really see how you can separate #392

test/unit/getclient-spec.ts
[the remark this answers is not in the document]
ejizba: That makes sense, although I'm still not a fan of merging commented code. How about we delete it for now? It'll still be in the git history and looks like you already have the issue to add it back.

hossam-nasr:
> I think the CI build for this PR needs to pass before merging, so I don't really see how you can separate #392

Yeah good point lol. I updated the package version in this PR and updated #392 to instead track updating the package to a version that supports returning a value to the extension.

src/durableorchestrationclient.ts
ejizba: I see what you're saying, but I think it's unreasonable for an npm package to validate every single input it gets like this. I'm guessing this is related to this discussion: https://github.com/Azure/azure-functions-nodejs-library/pull/14#discussion_r964309579

I don't have a strict answer on why, but this case just feels excessive to me where the original doesn't. One reason is that the user is manually crafting the return value in the original and it can take many different formats (http response, buffer, string, number, etc.) but in this case they're just passing along a context object already created for them. In other words it's not that they _can_ pass anything in the original it's that there are valid scenarios where they _will_ pass virtually anything.

Another thing is I don't like `instanceof` because it can prevent valid mock testing scenarios

hossam-nasr: Yeah that makes sense, thanks! Do you think I should also remove the check for `isClientInput` or keep it?

test/unit/getclient-spec.ts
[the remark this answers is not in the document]
hossam-nasr: Yeah that's fair. I removed the commented code

src/durableorchestrationclient.ts
ejizba: Probably yeah. Ultimately this isn't unique to the new programming model stuff anyways, so if you decide to do it (which I'm not advocating for) it would be better to do a separate PR and cherry-pick to the other branch

src/durableorchestrationclient.ts
```diff
@@ -46,8 +47,14 @@ const URL = url.URL;
  * };
  * ```
  */
-export function getClient(context: unknown): DurableOrchestrationClient {
-    let clientData = getClientData(context as IOrchestrationFunctionContext);
+export function getClient(
+    context: IOrchestrationFunctionContext,
+    clientInputOptions: OrchestrationClientInput
```
davidmrdavid, line 52: What is this new `clientInputOptions` parameter?

test/unit/getclient-spec.ts
```diff
@@ -2,124 +2,8 @@ import chai = require("chai");
...
-describe("getClient()", () => {
-    const defaultTaskHub = "TestTaskHub";
-    const defaultConnection = "Storage";
-
-    const defaultClientInputData = TestUtils.createOrchestrationClientInputData(
-        TestConstants.idPlaceholder,
-        Constants.DefaultLocalOrigin,
-        defaultTaskHub,
-        defaultConnection
-    );
```
davidmrdavid, line 22: should we comment this out instead of deleting them? Might make it easier to recover once we re-enable them.

src/orchestrator.ts
```diff
@@ -22,19 +22,25 @@ export class Orchestrator {
...
+    private async handle(
+        context: IOrchestrationFunctionContext,
+        orchestrationTrigger: DurableOrchestrationBindingInfo
+    ): Promise<OrchestratorState> {
         this.taskOrchestrationExecutor = new TaskOrchestrationExecutor();
         const orchestrationBinding = Utils.getInstancesOf<DurableOrchestrationBindingInfo>(
-            context.bindings,
+            { trigger: orchestrationTrigger },
             new DurableOrchestrationBindingInfoReqFields() as DurableOrchestrationBindingInfo
         )[0];
```
davidmrdavid, line 41: I'm confused about how `getInstancesOf` is working after this change. 
My understanding is that `getInstancesOf` used to search for any values of its generic type (in this case that's `DurableOrchestrationBindingInfo` inside of `context`.

However, now it seems like we're getting the binding info as a parameter itself, so do we still need to call `getInstancesOf`?

davidmrdavid (commented):
Left some questions

src/durableorchestrationclient.ts
hossam-nasr: Well, it is unique for the programming model because the old one doesn't even have this second `FunctionInput` argument. Regardless, I've remove the validation for both

azure-pipelines.yml
```diff
@@ -73,22 +45,6 @@ jobs:
...
-                CleanTargetFolder: true
-          - script: npm install $(System.DefaultWorkingDirectory)/package.tgz
-            displayName: "Install packed durable-functions module (test app)"
-            workingDirectory: $(Agent.BuildDirectory)/test-app
-          - script: npm install
-            displayName: "npm install (test app)"
-            workingDirectory: $(Agent.BuildDirectory)/test-app
-          - script: npm run build
-            displayName: "Build smoke test app"
-            workingDirectory: "$(Agent.BuildDirectory)/test-app"
```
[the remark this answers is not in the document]
hossam-nasr: The smoke test step currently would try to build an old programming model app with this build of the SDK, which would fail. Ideally, I'd like to replace the smoke test step with a new app that follows the new programming model, but it's hard to do this in this PR when the API for the new programming model isn't finalized. I've created an issue to track it though: https://github.com/Azure/azure-functions-durable-js/issues/394

src/durableorchestrationclient.ts
hossam-nasr: Unlike the orchestrator, activity, and entity functions, client functions don't have a special trigger. Instead, they are normal triggered functions and the client comes from an extra input binding. In the old model, the `getClient` function would just loop over all the available bindings on the `Context` object and find the one that matches the `durableClient` input binding. The new programming model doesn't work like that anymore. Instead, to register an input binding, this is how you would register and access it:

```TS
const clientInput = input.generic({
    type: 'orchestrationClient'
});

app.generic('DurableFunctionsHttpStart', {
    trigger: trigger.http(),
    extraInputs: [clientInput],
    handler: async function (context: InvocationContext, req: HttpRequest): Promise<any> {
        const clientInputBinding = context.getInput(clientInput)
    }
})
```

So essentially, the `getClient` method would need to have access to the `FunctionInput` object that the customer added to their `extraInputs` property, so that it can call `context.getInput()` on behalf of the customer. The type `OrchestrationClientInput` is just a `FunctionInput` object with the `type` property set to `durableClient`.

Regardless, this shouldn't matter because all of this should be abstracted away from the customer, and we would provide an API that allows them to register a client function that just gets passed the client directly as its third parameter, like this:

```TS
df.httpClient('DurableFunctionsHttpStart', async (context: InvocationContext, req: HttpRequest, client: DurableOrchestrationClient) => {
    const instanceId = await client.startNew(req.query.get('functionName'), undefined, req.body);
    return client.createCheckStatusResponse(req, instanceId);
})
```

This is just a step to get there

test/unit/getclient-spec.ts
hossam-nasr: Initially, I had it commented out, but instead removed it as per Eric's suggestion here: https://github.com/Azure/azure-functions-durable-js/pull/389#discussion_r989458594. I don't mind either way. Regardless, I also have an issue to make sure we don't forget about these tests: https://github.com/Azure/azure-functions-durable-js/issues/393

src/orchestrator.ts
hossam-nasr: You're right that the main function of `getInstancesOf` was to search for the value in a collection. However, it also had validation of types, and, crucially, conversion of timestamps: https://github.com/Azure/azure-functions-durable-js/blob/7320b2d57ceaa53fbc0ac4212fcca21141acbfb9/src/utils.ts#L22. I figured I don't wanna mess with this right now, and just wanted to make the fewest changes necessary to get to a working POC. I do agree there could be a lot of refactoring of this function now that we're not searching for a value anymore. I could do that in this PR if you feel strongly about it.

davidmrdavid: Makes sense, good call. Thanks!

davidmrdavid (approved):

package.json
```diff
@@ -39,13 +39,14 @@
         "e2etst": "npm run "
     },
     "dependencies": {
-        "@azure/functions": "^1.2.3",
+        "@azure/functions": "^4.0.0-alpha.3",
```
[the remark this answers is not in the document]
hossam-nasr: Done

package.json
```diff
@@ -39,13 +39,14 @@
         "e2etst": "npm run "
     },
     "dependencies": {
-        "@azure/functions": "^1.2.3",
+        "@azure/functions": "^4.0.0-alpha.3",
         "axios": "^0.21.1",
         "debug": "~2.6.9",
         "lodash": "^4.17.15",
+        "moment": "^2.29.2",
+        "undici": "^5.9.1",
```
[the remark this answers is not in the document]
hossam-nasr: True, that was just a temporary fix for https://github.com/Azure/azure-functions-nodejs-library/issues/29. Removed

ejizba (approved):
🎉🎉🎉

PR #397, opened 2022-11-02
Fix entity context.done() error log

### LINKED ISSUE #373: Error: Choose either to return a promise or call 'done'. Do not use both in your script.

hossam-nasr:
Resolves #373 . This PR does a few things:

1. Updates the `@azure/functions` package to the most recent version, which includes the most updated Functions types, including the new `suppressAsyncDoneError` flag on the `Context` object, which unblocks fixing this error. See here: https://github.com/Azure/azure-functions-nodejs-library/pull/33
2. Updates tests to ensure compliance with the most up-to-date types (to allow this to build)
3. Sets the `suppressAsyncDoneError` flag to true before calling `context.done()` to suppress the error seen in #373
4. Adds `@types/node` as a devDependency in the smoke test app to allow the build pipeline to pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,13 +39,13 @@
         "e2etst": "npm run "
     },
     "dependencies": {
-        "@azure/functions": "^1.2.3",
+        "@azure/functions": "^3.5.0-alpha.5",
         "axios": "^0.21.1",
         "debug": "~2.6.9",
         "lodash": "^4.17.15",
+        "moment": "^2.29.2",
         "uuid": "~3.3.2",
-        "validator": "~13.7.0",
-        "moment": "^2.29.2"
+        "validator": "~13.7.0"
     },
     "devDependencies": {
         "@types/chai": "~4.1.6",
--- a/test/testobjects/testutils.ts
+++ b/test/testobjects/testutils.ts
@@ -1,4 +1,6 @@
+import { Form, HttpMethod, HttpRequest, HttpRequestUser } from "@azure/functions";
 import {
+    Constants,
     HttpCreationPayload,
     HttpManagementPayload,
     IOrchestratorState,
@@ -6,6 +8,9 @@ import {
 } from "../../src/classes";
 import { OrchestrationFailureError } from "../../src/orchestrationfailureerror";
 import { TestConstants } from "./testconstants";
+
+const defaultOrchestrationName = "TestOrchestration";
+const defaultRequestUrl = `${Constants.DefaultLocalOrigin}/orchestrators/${defaultOrchestrationName}`;
 
 export class TestUtils {
     public static createOrchestrationClientInputData(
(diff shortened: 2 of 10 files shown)


test/testobjects/testutils.ts
```diff
@@ -98,3 +103,37 @@ export class TestUtils {
...
     }
 }
+
+export class DummyHttpRequest implements HttpRequest {
+    public method: HttpMethod | null;
+    public user: HttpRequestUser | null;
+    public form: Form | null;
+
+    constructor(
+        public url = defaultRequestUrl,
```
davidmrdavid, line 113: Is this new class just to simplify the logic, or does it also perform a change needed by the types upgrade?

hossam-nasr: Both. Initially, the `HttpRequest` type was simple enough that it could be constructed on the fly when necessary, with something like:

```TS
 const requestObj: HttpRequest = {
                method: "GET",
                url: defaultRequestUrl,
                headers: {},
                query: {},
                params: {},
            };
```

But now, the `HttpRequest` type also includes methods such as `parseFormBody()`, `get()`, etc., such that it would be untenable to construct a brand new instance compliant with the type each time

package.json
```diff
@@ -39,13 +39,13 @@
         "e2etst": "npm run "
     },
     "dependencies": {
-        "@azure/functions": "^1.2.3",
+        "@azure/functions": "^3.5.0-alpha.5",
```
ejizba, line 42: I don't wanna release durable referencing an alpha version here. My plan was to release the real 3.5.0 package as soon as host 4.14 finished as that is the first version supporting our new npm package pattern. 4.14 is hopefully finishing soon, but let me know if you want the real 3.5.0 out sooner and we can probably do that for you

hossam-nasr: I think that's fine. This fix probably won't make a difference until the new flag rolls out in Azure anyway

hossam-nasr:
Discussed offline. I will leave this PR on the back burner for now, until we have a new stable 3.5.0 release of the library and have the fix for the flag we need roll out in Azure. In the meantime, if we agree on a solution/pattern to follow for #400 before then, then I will follow that in this PR as well. If not, then I will just cheese the types to get the new flag to work (add the flag to our own context type definitions) until we agree on a more sustainable solution for the types

hossam-nasr:
Since not updating the types package anymore, closing this in favor of #513

PR #399, opened 2022-11-03
Add APIs to register durable functions in new programming model

hossam-nasr:
This PR adds public-facing APIs to register durable functions (orchestrations, entities, activities, and durable clients) in the new Node.js programming mode. This step 5 and the last step of resolving #384 and supporting the new programming model in durable, in terms of functionality. The only remaining work would be on the documentation, tooling and testing side (updating our samples, re-enabling tests that were disabled, adding new tests for these APIs, adding back the smoke test to our build pipeline, etc.). 

The comments on each function describe what they do. Also find working examples in these two PRs: https://github.com/ejizba/func-nodejs-prototype/pull/14 and https://github.com/ejizba/func-nodejs-prototype/pull/15. The naming and design choices made in this PR are of course up for discussion 😛


ejizba:
For a PR like this I like to start by testing it out myself. For that reason, you'll see my first round of comments on the working example here: https://github.com/ejizba/func-nodejs-prototype/pull/14

davidmrdavid:
I'll add some questions to that prototype repo as well, to help centralize feedback.

src/shim.ts
```diff
@@ -6,43 +30,293 @@ import {
...
- * @example Initialize an orchestrator
+ * @example Register an orchestrator using `app.generic()`
  * ```javascript
+ * const { app } = require("@azure/functions")
  * const df = require("durable-functions");
  *
- * module.exports = df.orchestrator(function*(context) {
- *     // function body
- * });
+ * app.generic('DurableFunctionsOrchestratorJS', {
```
ejizba: I don't think I've seen people do a big example like this in intellisense and the editor isn't the best at displaying that much info in a tooltip:
Screen Shot 2022-11-09 at 10 43 30 AM

I think I would rather move these to docs, a wiki, or a samples repo.

src/shim.ts
```diff
@@ -6,43 +26,211 @@ import {
...
+ * })
  * ```
  */
-export function orchestrator(
-    fn: (context: IOrchestrationFunctionContext) => Generator<unknown, unknown, any>
-): (
-    context: IOrchestrationFunctionContext,
-    orchestrationTrigger: DurableOrchestrationBindingInfo
-) => Promise<OrchestratorState> {
+export function createOrchestrator(fn: OrchestrationHandler): OrchestrationFunction {
```
ejizba, line 46: I think the difference in naming between `orchestrator` and `createOrchestrator` is pretty confusing. My solution would actually be to combine them into one named `orchestration` - can we do that? If not, what is the use-case for `createOrchestrator`? If it's only used by `app.generic`, keep in mind that's primarily intended for trigger types we _don't_ support

src/shim.ts
```diff
@@ -6,43 +26,211 @@ import {
...
+    app.generic(functionName, {
+        ...options,
+        handler,
+    });
+}
+
+/**
+ * The root namespace to help create durable trigger configurations
+ */
+export namespace trigger {
```
ejizba, line 159: I'm curious about the choice to have a `trigger` namespace, but not an `app` namespace. I understand it may be confusing to have duplicate namespaces across packages, but it would be nice to at least have consistency within the durable package

If we duplicate namespaces, I assume people would do something like this:
```typescript
import * as func from '@azure/functions';
import * as df from 'durable-functions';

func.app.http(...);
df.app.orchestration(...);
```

Otherwise, I could see us adding a prefix to the durable namespaces like this:
```typescript
import { app } from '@azure/functions';
import { dfApp } from 'durable-functions'; // or durApp or durableApp

app.http(...);
dfApp.orchestration(...);
```

When deciding our path forward, would be nice if we had examples from other frameworks/packages.

src/types.ts
```diff
@@ -1,5 +1,75 @@
...
+import { DurableEntityBindingInfo } from "../src/durableentitybindinginfo";
+import { IEntityFunctionContext } from "../src/ientityfunctioncontext";
+import { IOrchestrationFunctionContext } from "../src/iorchestrationfunctioncontext";
+import { OrchestratorState } from "../src/orchestratorstate";
+import { DurableOrchestrationInput } from "../src/testingUtils";
 
+// orchestrations
+export type OrchestrationHandler = (
+    context: IOrchestrationFunctionContext
+) => Generator<unknown, unknown, any>;
```
ejizba, line 14: Can we get more specific on these types?

src/shim.ts
hossam-nasr: I guess we could combine them, but then we would be overloading the implementation, and I think that might be _more_ confusing, because they do subtly different but distinct things. `df.createOrchestrator()` takes in an `OrchestrationHandler` type (generator), and returns a `FunctionHandler` type -- something that can be passed to `app.generic()`. Meanwhile, `df.orchestration()` takes in an `OrchestrationHandler` generator and _registers_ it as a function on the function app, returning `void`. I don't see `createOrchestrator()` used much in production, but it could be useful for unit testing, for example see our own tests. Would be harder to have the same kind of tests with just `df.orchestration()`

src/shim.ts
hossam-nasr: It didn't seem like it made much sense to have a separate `app` namespace for Durable, since the methods that would be there should be the only main methods that the user interacts with to begin with. Conversely, I don't see `df.trigger` and `df.input()` being used outside of `app.generic()` use-cases, which aren't nearly as common. Another reason against it is that it avoids duplication of namespaces for most use-cases. 

In terms of duplicating namespaces, I would be more in favor of recommending the first pattern over the second. I don't think the SDK (a separate package) should be in the business of avoiding the most natural names for our namespaces to accommodate `@azure/functions` or any other package.

I could also see a third pattern being used:

```TS
import { input as azFuncApp } from '@azure/functions';
import { input as dfInput } from 'durable-functions';
```

Or they could choose to keep one as just `input` and rename just the other

src/types.ts
hossam-nasr: What do you mean?

ejizba: Can we use types better than `unknown` and `any`? We should avoid those whenever possible - it means you get zero intellisense/build errors/etc.

src/shim.ts
ejizba: `@azure/functions` isn't just "any other package" though. It's the parent package that everything in durable-functions is based off of. It feels like this should be a common enough pattern where one package is the clear parent of another, which is why I was hoping we could find other examples instead of designing a pattern from scratch. For example, I would think there are many packages that extend the express or fastify frameworks

When looking for examples, please keep in mind that we can totally change `@azure/functions` if necessary (of course, in a separate PR). You are serving as a trailblazer for people using/extending the package and I expect some bumps.

src/shim.ts
ejizba: I don't want multiple overloads - I was suggesting we get rid of the "subtle but distinct things" in favor of one function that did one thing. It sounds like you still want "subtle but distinct" for unit tests? If so, I would challenge that a little bit. Both `createOrchestrator` and `orchestration` are only 2-3 lines each - I don't think the unit tests are really testing those functions in particular, but rather the stuff they're wrapping. Regardless, I wouldn't expose `createOrchestrator` or any relevant types outside of this package if its only needed for unit tests

hossam-nasr: Well, I meant more for users' unit-testing too, not necessarily just our unit-testing. You're right, our unit tests aren't testing `df.createOrchestrator()` itself, but rather what it's wrapping, and same for users too -- it would help their unit tests

src/types.ts
hossam-nasr: I will actually defer to @davidmrdavid on this one. This type signature for the generator `Genrator<unknown, unknown, any>` predates me, I just gave it a name. I don't think I know enough about generators or their type constraints to answer that question. @davidmrdavid, do you know if we can enforce stricter types here?

davidmrdavid: I've been trying to answer this question myself for a while, but have never had enough JS/TS experts around me to give it a serious shot. Let's try it now...

So the signature of a Generator is `Generator<InputType, ReturnType, YieldType>`.

Some definitions
*  The input type I believe stands for the inputs that you give the generator every time you ask it to run up to the next yield or the return statement. 
* The return type is the type of data at the return statement
* The YieldType is the type of intermediate values

So I think we may be able to constrain `YieldType` to be of type "Task" (concretely, it may be `DFTask`, can't remember at this time). Users are only allowed to yield `Task` types, and if they're not yield'ing `Task`s then they should encounter an exception. So this is a type we could try to constraint.

For the `ReturnType`, that's hard. Orchestrators may return anything, but they may return only once. Is it possible to parametrize `OrchestrationHandler` with a generic parameter `T` that corresponds to the return type? If so, then yes: we could rename `ReturnType` with that generic `T` so that it's fixed.

For `InputType` - I don't think we can constraint that one at all. The orchestration generator is given as **inputs** the results of each DF API that is yielded. Those DF APIs may truly return anything.

@ejizba/@hossam-nasr: thoughts?

src/shim.ts
ejizba: Ah didn't realize you were trying to help with _users'_ unit testing. Are there existing guides/docs on how users should do unit testing in the old model? If so, please forward that to me and we can figure out a way to maintain the status quo. If not, unfortunately I think "users' unit testing" is going to be a way larger feature outside the scope of this PR and we should hide `createOrchestrator`/`OrchestrationFunction` because the names are too confusing.

src/types.ts
ejizba: I don't have a quick answer, and also I'm not very familiar with Generators. For this PR I'm fine sticking with the status quo since Hossam said it predates his changes

src/shim.ts
davidmrdavid: The testing story for DF in out-of-proc is unfortunately largely undocumented and community driving at this point. However, since this is a breaking change release, I'm not too worried about starting with a unit testing story from the get go.

At this time, `createOrchestrator` is a bit confusing for me as well. I would also like to punt `createOrchestrator` for this PR, and to discuss it's design in the context of unit testing in a follow-up PR.

src/shim.ts
davidmrdavid: If we're looking for precent: In PyStein, we're exporting the **entire** API (triggers and input bindings) through a `dfApp` namespace. Personally, I like it as it keeps the Function-declaration methods all in the same place and it also mimicks the baseline PyStein code organization

src/types.ts
```diff
@@ -1,5 +1,75 @@
...
+// clients
 export interface DurableClientInput extends FunctionInput {
     type: "orchestrationClient";
 }
+
+export type DurableClientHandler = (
+    context: InvocationContext,
+    triggerInput: any,
+    orchestrationClient: DurableOrchestrationClient
+) => FunctionResult<any>;
```
davidmrdavid: This is only required because we have a `df.client` convenience API, right? From what I gather, these `-Handler` types usually correspond to Triggers ("the handler is called by the trigger"), so I'm confused to see a handler for the DF client which is an input Binding.

src/types.ts
```diff
@@ -1,5 +1,75 @@
...
+export type OrchestrationHandler = (
+    context: IOrchestrationFunctionContext
+) => Generator<unknown, unknown, any>;
+
+export type OrchestrationFunction = FunctionHandler &
+    ((
+        context: IOrchestrationFunctionContext,
+        orchestrationTrigger: DurableOrchestrationInput
+    ) => Promise<OrchestratorState>);
+
```
davidmrdavid: We have an `OrchestrationHandler` and an `OrchestrationFunction` because we make a difference between `df.orchestration` and `df.createOrchestration`, right? If we only had one of these methods, would we also only need one of these types?

src/shim.ts
hossam-nasr: I'm happy with no longer exporting `createOrchestrator()` or revealing it as a public function, but I'd still like to keep it as a helper function, at least for this PR, because it's easier to keep our own existing unit tests as they are

hossam-nasr: Done

src/shim.ts
hossam-nasr: So I guess there are two different questions here:

1. Should APIs for registering functions be included in some kind of `app` namespace or in the global DF package directly?
2. Should DF namespaces differentiate themselves from `@azure/functions` namespaces by including `df` prefix? 

For those two questions, my personal preference is basically what I have in this PR: APIs for registering functions should be accessed by `df.` directly, and no to the second question. But I don't have strong feelings about this. Do you folks have strong feelings one way or the other?

src/types.ts
hossam-nasr: > This is only required because we have a `df.client` convenience API, right? 

Yes, and since we decided to remove the `df.client()` helper for this PR, I have also removed this type

src/types.ts
hossam-nasr: More or less, yes. It might be argued that those steps could still be useful internally though even if we don't export/expose them to users

src/shim.ts
```diff
@@ -6,43 +18,181 @@ import {
...
+ *     // orchestrator body
+ * });
+ * ```
+ */
+export function orchestration(functionName: string, handler: OrchestrationHandler): void {
+    app.generic(functionName, {
+        trigger: trigger.orchestration(),
+        handler: createOrchestrator(handler),
+    });
+}
```
davidmrdavid: Just a thought: After staring at this implementation, I think I finally understand what throws me off about the subtle distinction between `orchestration` and `createOrchestrator`.

The source of my confusion is in that the `handler: ` field in the options to `app.generic` does not accept the `OrchestrationHandler` type. Instead, it accepts an `OrchestratorFunction` type, which is unintuitive to me. I would expect that an `OrchestrationHandler` would be a valid type to pass in to the `handler: ` field here, but it is not. 

Put differently, `createOrchestrator` transforms an `OrchestratorHandler` into an `OrchestratorFunction`, which can then be passed as a `handler: ` to app generic. In my mind, it would be more intuitive if the names were reverse.

Put differently, it would make more sense to me if `createOrchestrator` transformed a type `OrchestratorFunction` into an `OrchestratorHandler` which naturally fits as the type accepted by the `handler: ` field of `app.generic`. Thoughts? @hossam-nasr

src/shim.ts
davidmrdavid: I don't fully understand the subtlety between questions (1) and (2). Aren't these just two different ways of importing a namespace? @hossam-nasr?

davidmrdavid (commented):
Some thoughts and questions

src/shim.ts
ejizba: > Done

Looks like you still need to remove `OrchestrationFunction` and `EntityFunction` from the public types.

Btw, aren't these the same?
```typescript
export function createOrchestrator(fn: OrchestrationHandler): OrchestrationFunction {
    const listener = new Orchestrator(fn).listen();

    return async (
        context: IOrchestrationFunctionContext,
        orchestrationTrigger: DurableOrchestrationInput
    ): Promise<OrchestratorState> => {
        return await listener(context, orchestrationTrigger);
    };
}
```

```typescript
export function createOrchestrator(fn: OrchestrationHandler): OrchestrationFunction {
    return new Orchestrator(fn).listen();
}
```

I just think we're being a bit generous talking about "subtle distinction"/"helper function"/etc. for a function that barely does anything

src/shim.ts
```diff
@@ -6,43 +18,181 @@ import {
...
+ * const df = require("durable-functions");
+ *
+ * app.generic('DurableEntity', {
+ *   trigger: df.trigger.entity(),
+ *   handler: df.createEntityFunction(function* (context) {
+ *       // orchestrator body
+ *  })
+ * })
+ * ```
+ */
```
ejizba, line 61: Please remove all of these extra docs as well since we're not exposing these methods publicly

src/shim.ts
ejizba: > 1. Should APIs for registering functions be included in some kind of app namespace or in the global DF package directly?

The current answer is yes for the programming model. I don't necessarily have a strong preference what we do, but I do have a strong preference that durable and non-durable should be consistent. If you would like to change that for both, I would suggest you jump in on this discussion: https://github.com/Azure/azure-functions-nodejs-library/issues/12. As stated earlier, I prefer examples from other frameworks as opposed to just personal preference, though.

> 2. Should DF namespaces differentiate themselves from @azure/functions namespaces by including df prefix?

No strong preference. Partially because unlike question 1, I feel like the answer to this is specific to durable

src/shim.ts
hossam-nasr: > Looks like you still need to remove OrchestrationFunction and EntityFunction from the public types.

Done.

> Btw, aren't these the same?

I'm not sure honestly. Idk why they were done the first way bc you're right the second one looks to be doing the same thing. @davidmrdavid do you know?

src/shim.ts
hossam-nasr: Okay, I've added the `app` namespace, but kept the names as-is for now. I'm happy to rename them to `dfApp`, `dfInput`, and `dfTrigger` (or similar) if we get strong feedback for this (or if anyone has strong feelings about this)

src/shim.ts
hossam-nasr: Discussed offline. `createOrchestrator` and the `OrchestrationFunction` type have been removed from public access so some of this discussion is obsolete now anyway

src/shim.ts
hossam-nasr: I've simplified the examples a lot, so that they're not same complexity or less than what we had earlier. Lmk if you think this is better or if we should get rid of the examples entirely

src/shim.ts
hossam-nasr: Done

src/shim.ts
davidmrdavid: I don't know, these decision predate me.
I prefer the simplified second version.

package.json
```diff
@@ -17,6 +17,9 @@
     ],
     "main": "lib/src/index.js",
     "typings": "lib/src/index.d.ts",
+    "engines": {
+        "node": ">=18.0"
+    },
```
davidmrdavid, line 22: I assume the new new prog-model is only available in Node18+, right?

src/shim.ts
```diff
@@ -6,43 +21,166 @@ import {
...
+/**
+ * The root namespace to help create durable input configurations
+ */
+export namespace input {
+    /**
+     * @returns a durable client input configuration object
+     */
+    export function durableClient(): DurableClientInput {
+        return azFuncInput.generic({
+            type: "orchestrationClient",
```
davidmrdavid: I think the preferred `type` here should be `"durableClient"`.
I think `orchestrationClient` is a bit of a legacy name.,

davidmrdavid (approved):
The API makes sense to me, although I think the naming of certain things is still to be discussed with the DF Team. Please do not make a public preview announcement before demo'ing and gathering feedback from the DF team. Internal testing is fine though.

My only request is that you try changing the value of `type:` in the client from `orchestrationClient` to `durableClient`. After validating that works (I'm certain it should), and getting @ejizba's approval, I think you can merge this.

src/shim.ts
ejizba: Intellisense is a priority for us in the new model, but not a blocker for this PR. I will spend time looking at the docs/examples at a later time

src/shim.ts
ejizba: Suffice to say I think there's still plenty of room for improvement for our own internal sake, but you removed the most confusing bits from being exported externally, so I will consider this resolved

src/shim.ts
```diff
@@ -6,43 +21,166 @@ import {
...
+     * Registers a generator function as a Durable Orchestrator for your Function App.
+     *
+     * @param functionName the name of your new durable orchestrator
+     * @param handler the generator function that should act as an orchestrator
+     *
+     * @example Register an orchestrator
+     * ```javascript
+     * const df = require("durable-functions");
+     *
+     * df.orchestration('durableOrchestration1', function* (context) {
```
ejizba: Yeaahh you should probably just remove these examples. I don't think they work (shouldn't it be `df.app.orchestration` now?) and I don't like how we have sample JavaScript code without sample TypeScript code (but don't do both because that would also be bad for length reasons)

package.json
hossam-nasr: Correct

src/shim.ts
hossam-nasr: I actually don't think this is possible (or at least it would require more changes at the extensions level). This string translates to what you would put in a `function.json` file. Since the name of the binding is `"orchestrationClient"`, this is what I have here. Otherwise, it wouldn't register this binding. Although I'm not sure where the names of bindings come from to begin with? 🤔

hossam-nasr: My bad I didn't realize both were already valid names for this binding! We should update our templates to reflect the new preferred name because I've only ever seen `orchestationClient` (at least for node templates)

davidmrdavid: I can confirm that there's a "durableClient" binding that is equivalent to the "orchestrationClient" :). See here: https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-bindings?tabs=python%2C2x-durable-functions#client-usage

hossam-nasr: Updated this to `durableClient` though

src/types.ts
```diff
@@ -1,5 +1,45 @@
...
+import { FunctionInput, FunctionHandler, FunctionOutput, FunctionTrigger } from "@azure/functions";
+import { IEntityFunctionContext } from "../src/ientityfunctioncontext";
+import { IOrchestrationFunctionContext } from "../src/iorchestrationfunctioncontext";
 
+// orchestrations
+export type OrchestrationHandler = (
+    context: IOrchestrationFunctionContext
+) => Generator<unknown, unknown, any>;
+
+export type OrchestrationOptions = {
```
ejizba: This should be an interface that `extends Partial<FunctionOptions>`. If we add something in `FunctionOptions`, we want durable to respect it as well. For example, right now I believe you can only set a function timeout in your `host.json` file. In the future, it would be nice if we allow users to configure the timeout per-function on this options object

src/types.ts
```diff
@@ -1,5 +1,45 @@
...
+    handler: OrchestrationHandler;
+};
+
+export interface OrchestrationTrigger extends FunctionTrigger {
+    type: "orchestrationTrigger";
+}
+
+// entities
+export type EntityHandler<T> = (context: IEntityFunctionContext<T>) => void;
```
ejizba, line 25: Ideally entity shouldn't use generics either. Can we change it to `IEntityFunctionContext<T = unknown>` in `ientityfunctioncontext.ts` and remove the entity generics from this file? If not, I won't block on this since I know `IEntityFunctionContext` already had generics before your PR

src/shim.ts
```diff
@@ -6,43 +23,220 @@ import {
...
+        functionName: string,
+        handlerOrOptions: OrchestrationHandler | OrchestrationOptions
+    ): void {
+        const handler: OrchestrationHandler =
+            typeof handlerOrOptions === "function" ? handlerOrOptions : handlerOrOptions.handler;
+
+        azFuncApp.generic(functionName, {
+            trigger: trigger.orchestration(),
+            handler: createOrchestrator(handler),
+        });
```
ejizba: We want to pass along the user's options, so it should be something like this:
```typescript
const options: OrchestrationOptions =
    typeof handlerOrOptions === "function"
        ? { handler: handlerOrOptions }
        : handlerOrOptions;

azFuncApp.generic(functionName, {
    trigger: trigger.orchestration(),
    ...options,
    handler: createOrchestrator(options.handler),
});
```

src/types.ts
hossam-nasr: This might work for activities, but for entities and orchestrations, won't that allow them to add extra inputs and extra outputs? Which they shouldn't be allowed to do

src/shim.ts
hossam-nasr: Done

src/shim.ts
hossam-nasr: Done

src/types.ts
hossam-nasr: Done

ejizba (approved):

PR #405, opened 2022-11-12
Patch isReplaying on WhenAll

### LINKED ISSUE #404: isReplaying returns unexpected result after fan out
**Description**
context.df.isReplaying will return an unexpected result (true) the first time the line of code is hit after tasks fan out, in case there was another activity called before.

**Investigative information**

- Durable Functions extension version: 1.8.3 (latest at the moment of creating an issue)
- durable-functions npm module version: 2.1.0 (latest at the moment of creating an issue
- Language (JavaScript/TypeScript) and version: Typescript 4.0.0
- Node.js version: v14.19.3

Repository to reproduce: https://github.com/danyltsiv/isReplayingCheck
Orchestrator: https://github.com/danyltsiv/isReplayingCheck/tree/master/orchestrator

**To Reproduce**

Just put a breakpoint right after the tasks are awaited:
IsReplayTrueWhen1stTime
at first hit context.df.isReplaying will say true. So, if you want to log something under it, it won't appear:
image

Also, there should be at least one call to any activity function before:
https://github.com/danyltsiv/isReplayingCheck/blob/master/orchestrator/index.ts#L17
The weird thing is that if there is no activity called before, the issue won't reproduce.

**Expected behavior**
The context.df.isReplaying should return false as this is the first hit during orchestrator executing

**Actual behavior**
The context.df.isReplaying returns true, and if we want to put some logs right under it they won't appear

davidmrdavid:
Resolves: https://github.com/Azure/azure-functions-durable-js/issues/404

We currently have a bug where in WhenAll where ithe value of `isPlayed` for the `WhenAll` task will be set to `True` if any of its sub-tasks has a `isPlayed` value of `True`. The correct behavior is that the value of `WhenAll's` `IsPlayed` should be `True` when *all* of its sub-tasks are `True`. 

This PR addresses this.

Note: for `WhenAny` it suffices that only one of its Tasks have `isPlayed` as `True` for it's `isPlayed` to be `True` as well

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -245,7 +245,6 @@ export abstract class CompoundTask extends DFTask {
         // If the task has no children, then it's completed by definition.
         if (children.length == 0) {
             this.state = TaskState.Completed;
-            // TODO: how should isPlayed change here?
         }
 
         // Sub-tasks may have already completed, so we process them
--- a/src/task.ts
+++ b/src/task.ts
@@ -344,8 +344,7 @@ export class WhenAllTask extends CompoundTask {
     }
 
     trySetIsPlayed(): void {
-        const isPlayed = this.children.every((c) => c.isPlayed);
-        this.isPlayed = isPlayed;
+        this.isPlayed = this.children.every((c) => c.isPlayed);
     }
 
     /**
@@ -380,8 +379,7 @@ export class WhenAllTask extends CompoundTask {
  */
 export class WhenAnyTask extends CompoundTask {
     trySetIsPlayed(): void {
-        const isPlayed = this.children.some((c) => c.isPlayed);
-        this.isPlayed = isPlayed;
+        this.isPlayed = this.children.some((c) => c.isPlayed);
     }
     /**
      * @hidden
--- a/src/task.ts
+++ b/src/task.ts
@@ -125,6 +125,7 @@ export abstract class TaskBase {
      */
     constructor(public id: TaskID, protected action: BackingAction) {
         this.state = TaskState.Running;
+        this.isPlayed = false;
     }
 
     /** Get this task's backing action */
@@ -244,6 +245,7 @@ export abstract class CompoundTask extends DFTask {
         // If the task has no children, then it's completed by definition.
         if (children.length == 0) {
             this.state = TaskState.Completed;
+            // TODO: how should isPlayed change here?
         }
 
         // Sub-tasks may have already completed, so we process them
@@ -261,11 +263,16 @@ export abstract class CompoundTask extends DFTask {
      *  A sub-task of this task.
      */
     public handleCompletion(child: TaskBase): void {
-        if (!this.isPlayed) {
-            this.isPlayed = child.isPlayed;
-        }
+        this.trySetIsPlayed();
         this.trySetValue(child);
     }
+
+    /**
+     * @hidden
+     *
+     * Task-internal logic for setting this task's isPlayed flag
+     */
+    abstract trySetIsPlayed(): void;
 
     /**
      * @hidden
@@ -336,6 +343,11 @@ export class WhenAllTask extends CompoundTask {
         super(children, action);
     }
 
+    trySetIsPlayed(): void {
+        const isPlayed = this.children.every((c) => c.isPlayed);
+        this.isPlayed = isPlayed;
+    }
+
     /**
      * @hidden
      * Attempts to set a value to this task, given a completed sub-task
@@ -367,6 +379,10 @@ export class WhenAllTask extends CompoundTask {
  * A WhenAny task.
  */
 export class WhenAnyTask extends CompoundTask {
+    trySetIsPlayed(): void {
+        const isPlayed = this.children.some((c) => c.isPlayed);
+        this.isPlayed = isPlayed;
+    }
     /**
      * @hidden
      * Attempts to set a value to this task, given a completed sub-task
@@ -382,6 +398,7 @@ export class WhenAnyTask extends CompoundTask {
         // in the result as a value.
         if (this.state === TaskState.Running) {
             this.setValue(false, child);
+            this.isPlayed = child.isPlayed;
         }
     }
 }
@@ -419,6 +436,11 @@ export class CallHttpWithPollingTask extends CompoundTask {
             defaultHttpAsyncRequestSleepTimeMillseconds,
             "ms"
         );
+    }
+
+    trySetIsPlayed(): void {
+        const isPlayed = this.children.every((c) => c.isPlayed);
+        this.isPlayed = isPlayed;
     }
 
     public trySetValue(child: TaskBase): void {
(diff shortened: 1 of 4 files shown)


davidmrdavid:
@hossam-nasr: a friendly ping here. Would appreciate a review to get this hotfix out :) Thanks!

src/task.ts
```diff
@@ -244,6 +245,7 @@ export abstract class CompoundTask extends DFTask {
         // If the task has no children, then it's completed by definition.
         if (children.length == 0) {
             this.state = TaskState.Completed;
+            // TODO: how should isPlayed change here?
```
hossam-nasr: There is still a TODO here?

src/task.ts
```diff
@@ -336,6 +343,11 @@ export class WhenAllTask extends CompoundTask {
         super(children, action);
     }
 
+    trySetIsPlayed(): void {
+        const isPlayed = this.children.every((c) => c.isPlayed);
+        this.isPlayed = isPlayed;
```
hossam-nasr: Nit; this could be merged into one line

src/task.ts
```diff
@@ -367,6 +379,10 @@ export class WhenAllTask extends CompoundTask {
  * A WhenAny task.
  */
 export class WhenAnyTask extends CompoundTask {
+    trySetIsPlayed(): void {
+        const isPlayed = this.children.some((c) => c.isPlayed);
+        this.isPlayed = isPlayed;
```
hossam-nasr: Nit: same here

src/task.ts
davidmrdavid: Yeah, but I think that `TODO` is not directly related to the bug we're fixing here :)
Essentially, that TODO is about how to update the Task's isPlayed flag when it is either a WhenAll or WhenAny with no sub-tasks. I haven't fully thought about whether the resulting Task should have `isPlayed` equals to `true` or `false`, especially as that will affect the `context.IsReplaying` flag.

So the way I'm thinking about this is simply as a note for us to investigate later, but it's not a remaining "to do" item for this bug fix. Does that seem reasonable?

davidmrdavid:
Thanks @hossam-nasr, I've included your suggestions and responded to your feedback. Would appreciate another review!

src/task.ts
hossam-nasr: Your reasoning does sound reasonable yes, but I wouldn't want to merge code with a TODO in it. It's very easy to lose track of it and lose context. (there's already a couple places in the code where I've run into a TODO and not sure what it was intending to do). That being said, in order to not block this PR, I'd be more in favor of creating an issue for us to investigate this further (basically just saying what you wrote above), and removing this comment from the PR

hossam-nasr: Also, I'm curious, what is the current behavior of this scenario (the `isPlayed` flag of a compound task that has no children)? Does this PR change the current behavior of this scenario?

src/task.ts
```diff
@@ -261,12 +263,17 @@ export abstract class CompoundTask extends DFTask {
      *  A sub-task of this task.
      */
     public handleCompletion(child: TaskBase): void {
-        if (!this.isPlayed) {
-            this.isPlayed = child.isPlayed;
-        }
+        this.trySetIsPlayed();
```
hossam-nasr, line 265: From a design perspective, wouldn't it be easier to pass in the child here, same as `trySetValue()`? I understand for `WhenAll`, we would need to loop over all the children anyway, so we can ignore that argument, but for a `WhenAny`, we can just set the compound task's `isPlayed` to that of the child? Or am I missing something here?

src/task.ts
```diff
@@ -125,6 +125,7 @@ export abstract class TaskBase {
      */
     constructor(public id: TaskID, protected action: BackingAction) {
         this.state = TaskState.Running;
+        this.isPlayed = false;
```
hossam-nasr, line 128: This checking: do we need to make sure this is updated for non-compound tasks too, or is that already handled?

davidmrdavid: Well this is being updated in the `TaskBase` class, so this change should apply to compound and non-compound tasks alike. Does that answer your question or did I misunderstand it?

src/task.ts
davidmrdavid: I think you're slightly onto something. For `WhenAny`, we can't just set the `isPlayed` flag to that of the child being passed in. _If_ we wanted to make us of a `child` argument, then we would need to guarantee that we only set `isPlayed` to be equal to the `isPlayed` value of the **first** child to complete. Alternatively, we could set `isPlayed` such that it may only go from `False` to `True` but not the other way around. If we were to use this, it would be an optimization for sure, although it's unclear to me if we gain much from it.

For now, I personally like the declarative style of this implementation because it gives me confidence in its correctness and I don't think we've had a performance complaint from this side. But yes, I think we could consider optimizing this scenario, although I'd rather make that part of a different PR. Would you like to file an issue for it?

src/task.ts
davidmrdavid: I've created that issue [here](https://github.com/Azure/azure-functions-durable-js/issues/423) and will remove the `TODO` comment on my next commit.

As for your question, the `isPlayed` flag today remains with its default value of `False`. This is an arbitrary and accidental choice, which is why I added the `TODO` comment to investigate further.

hossam-nasr: Sounds good, thank you!

src/task.ts
hossam-nasr: I'm not sure I understand this:

> If we wanted to make use of a child argument, then we would need to guarantee that we only set isPlayed to be equal to the isPlayed value of the first child to complete.

I thought we already made that guarantee? In order words, once _any_ child of a `WhenAny` completes, doesn't the value of the entire compound `WhenAny` task equal that _first_ completed child (including, I'm assuming, its `isPlayed` flag)? If that's not the case, then I'm not sure I understand the semantics of the `isPlayed` flag on a `WhenAny` task

src/task.ts
hossam-nasr: Hmm so I get that here we are setting the default initial value to be `false`, which will apply to both compound and non-compound tasks. However, in addition to that, this PR also provides a way for compound tasks to update that default value through `trySetIsPlayed()`. I'm wondering if we need an equivalent for non-compound tasks too?

src/task.ts
hossam-nasr: Either way, I don't want to be pedantic, so I won't block on this and I will trust that you know what you're doing here

davidmrdavid: I've filed the request to optimize this logic here: https://github.com/Azure/azure-functions-durable-js/issues/425

src/task.ts
davidmrdavid: Ah, I found where this is set for atomic (i.e not compound) tasks.

Essentially, as we iterate over History events, we set the `IsPlayed` flag of those atomic events, because atomic results are stored in individual History events. See [here](https://github.com/Azure/azure-functions-durable-js/blob/f716f842401ccfdeaf00c989a317944116110627/src/taskorchestrationexecutor.ts#L357-L360)

davidmrdavid:
I'd appreciate another review @hossam-nasr. Thanks!

src/task.ts
hossam-nasr: Makes sense, thank you for looking into it!

hossam-nasr (approved):
LGTM :) Thanks @davidmrdavid! :shipit:

PR #429, opened 2023-01-24
Switch `client.startNew()` to using object for optional arguments

### LINKED ISSUE #415: Simplify the `client.startNew()` experience
Based on discussion [here](https://github.com/ejizba/func-nodejs-prototype/pull/14#discussion_r1044721159). Currently, the default HTTP starter `client.startNew()` call looks like this:

```TS
const instanceId = await client.startNew(request.params.orchestratorName, undefined, request.text());
```

Ideally, `client.startNew()` should accept one single object as an argument, rather than 3 distinct arguments, so that:

1. It's clearer what each argument is for
2. Order doesn't matter
3. Unnecessary arguments can be omitted instead of passing in `undefined`.

Since this would be a breaking change, it might be better to include this change with the rest of the v3.x new programming model changes.

hossam-nasr:
Fixes #415. See issue for details

### DIFF
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -34,17 +34,7 @@ const URL = url.URL;
  * Returns an OrchestrationClient instance.
  * @param context The context object of the Azure function whose body
  *  calls this method.
- * @example Get an orchestration client instance
- * ```javascript
- * const df = require("durable-functions");
  *
- * module.exports = async function (req, context) {
- *     const client = df.getClient(context, clientInput);
- *     const instanceId = await client.startNew(req.params.functionName, { input: req.text() });
- *
- *     return client.createCheckStatusResponse(req, instanceId);
- * };
- * ```
  */
 export function getClient(
     context: InvocationContext,
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -24,7 +24,7 @@ import {
     PurgeHistoryResult,
     Utils,
 } from "./classes";
-import { DurableClientInput } from "./types";
+import { ClientStartNewOptions, DurableClientInput } from "./types";
 import { WebhookUtils } from "./webhookutils";
 
 /** @hidden */
@@ -38,9 +38,9 @@ const URL = url.URL;
  * ```javascript
  * const df = require("durable-functions");
  *
- * module.exports = async function (context, req) {
- *     const client = df.getClient(context);
- *     const instanceId = await client.startNew(req.params.functionName, undefined, req.body);
+ * module.exports = async function (req, context) {
+ *     const client = df.getClient(context, clientInput);
+ *     const instanceId = await client.startNew(req.params.functionName, { input: req.text() });
  *
  *     return client.createCheckStatusResponse(req, instanceId);
  * };
@@ -648,17 +648,13 @@ export class DurableOrchestrationClient {
      * existing instance will be silently replaced by this new instance.
      * @param orchestratorFunctionName The name of the orchestrator function to
      *  start.
-     * @param instanceId The ID to use for the new orchestration instance. If
-     *  no instanceId is specified, the Durable Functions extension will
-     *  generate a random GUID (recommended).
-     * @param input JSON-serializable input value for the orchestrator
-     *  function.
+     * @param options optional object specifying the orchestration instance ID and any optional input
+     * @param input
      * @returns The ID of the new orchestration instance.
      */
     public async startNew(
         orchestratorFunctionName: string,
-        instanceId?: string,
-        input?: unknown
+        options?: ClientStartNewOptions
     ): Promise<string> {
         if (!orchestratorFunctionName) {
             throw new Error("orchestratorFunctionName must be a valid string.");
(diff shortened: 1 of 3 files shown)


src/durableorchestrationclient.ts
```diff
@@ -648,42 +648,37 @@ export class DurableOrchestrationClient {
...
                 .replace(this.functionNamePlaceholder, orchestratorFunctionName)
-                .replace(this.instanceIdPlaceholder, instanceId ? `/${instanceId}` : "");
+                .replace(this.instanceIdPlaceholder, instanceIdPath);
         }
 
-        const response = await this.axiosInstance.post(
-            requestUrl,
-            input !== undefined ? JSON.stringify(input) : ""
-        );
+        const response = await this.axiosInstance.post(requestUrl, input);
```
davidmrdavid: has this been tested with `input` as `undefined`? I'm asking since I noticed our validation for this has been removed

src/durableorchestrationclient.ts
```diff
@@ -648,42 +648,37 @@ export class DurableOrchestrationClient {
      * existing instance will be silently replaced by this new instance.
      * @param orchestratorFunctionName The name of the orchestrator function to
      *  start.
-     * @param instanceId The ID to use for the new orchestration instance. If
-     *  no instanceId is specified, the Durable Functions extension will
-     *  generate a random GUID (recommended).
-     * @param input JSON-serializable input value for the orchestrator
-     *  function.
+     * @param options optional object specifying the orchestration instance ID and any optional input
```
davidmrdavid: Can we provide a bit more detail in the comment on what these optional inputs are? Something as small as adding "(e.g the new orchestration's instance ID)" would do.

src/types.ts
```diff
@@ -49,3 +49,20 @@ export interface ActivityTrigger extends FunctionTrigger {
 export interface DurableClientInput extends FunctionInput {
     type: "durableClient";
 }
+
+/**
+ * Options object provided as an optional second argument to the `client.startNew()` method
+ */
+export interface ClientStartNewOptions {
```
davidmrdavid: I would just call this `StartNewOptions` as this method only exists in the Client anyways, so I don't think the prefix "Client" is adding all that much. What do you think?

davidmrdavid (commented):
Left some feedback

src/durableorchestrationclient.ts
hossam-nasr: It hasn't been removed, it's just been moved up a step. Here, `input`, isn't the raw input coming from the user, but rather is defined here:

```TS
const input: unknown = options?.input !== undefined ? JSON.stringify(options.input) : "";
```

src/durableorchestrationclient.ts
```diff
@@ -38,9 +38,9 @@ const URL = url.URL;
  * ```javascript
  * const df = require("durable-functions");
  *
- * module.exports = async function (context, req) {
- *     const client = df.getClient(context);
- *     const instanceId = await client.startNew(req.params.functionName, undefined, req.body);
+ * module.exports = async function (req, context) {
+ *     const client = df.getClient(context, clientInput);
+ *     const instanceId = await client.startNew(req.params.functionName, { input: req.text() });
```
ejizba: Isn't this supposed to have `await` before `req.text()`? To bring it back to previous comments I've made, this is another reason why I prefer to remove examples from intellisense because I just think they get out-of-date and inaccurate very quickly. Also, remember that if a user has an old version of the npm package they will be stuck with the old examples, as opposed to online docs and templates that should always be up to date.

src/durableorchestrationclient.ts
ejizba: Maybe just move line 666 down a bit closer to where `input` is used. I think that'll make it a tad less confusing

ejizba (approved):

src/durableorchestrationclient.ts
hossam-nasr: Done

src/durableorchestrationclient.ts
hossam-nasr: I already mention the orchestration instance ID in the comment, but I updated the comment to make it clearer what I mean by "input" here is input to the orchestration

src/types.ts
hossam-nasr: How about `StartNewClientOptions`? `StartNewOptions` seems a bit vague to me, but I also am not married to the name so if you feel strongly about it I can go with that

src/durableorchestrationclient.ts
hossam-nasr: Yeah you're right. I've removed this example from the comments

src/types.ts
hossam-nasr: Sure, done :)

src/types.ts
```diff
@@ -49,3 +49,20 @@ export interface ActivityTrigger extends FunctionTrigger {
...
 }
+
+/**
+ * Options object provided as an optional second argument to the `client.startNew()` method
+ */
+export interface StartNewOptions {
+    /**
+     * The ID to use for the new orchestration instance. If
+     *  no instanceId is specified, the Durable Functions extension will
+     *  generate a random GUID (recommended).
```
davidmrdavid: nitpick: I noticed the 2nd and 3rd line are at a different indentation. Can you please normalize these?

src/durableorchestrationclient.ts
```diff
@@ -648,42 +638,36 @@ export class DurableOrchestrationClient {
      * existing instance will be silently replaced by this new instance.
      * @param orchestratorFunctionName The name of the orchestrator function to
      *  start.
-     * @param instanceId The ID to use for the new orchestration instance. If
-     *  no instanceId is specified, the Durable Functions extension will
-     *  generate a random GUID (recommended).
-     * @param input JSON-serializable input value for the orchestrator
-     *  function.
+     * @param options optional object specifying the orchestration instance ID and any input to the orchestration
```
davidmrdavid: This parameter may grow over time. So I suggest this more general wording:

```suggestion
     * @param options optional object to control the scheduled orchestrator (e.g provide input, instanceID)

```

davidmrdavid (approved):
LGTM, left non-blocking suggestions

src/types.ts
hossam-nasr: Done

src/durableorchestrationclient.ts
hossam-nasr: Done

PR #432, opened 2023-01-25
Default release pipeline to retrieve latest build from v3.x branch instead of main

hossam-nasr:
As in title


davidmrdavid (approved):
LGTM

PR #435, opened 2023-01-25
Make callHttp accept one options object argument

hossam-nasr:
Following the lead of #415 and #429, this PR also switches the `context.df.callHttp()` on orchestration contexts from using explicit arguments to using one options argument for all parameters, following the lead of popular frameworks like `axios`

Before (actual sample from our docs [here](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-features?tabs=javascript#managed-identities)):

```TS
const restartResponse = yield context.df.callHttp(
        "POST",
        `https://management.azure.com/subscriptions/${subscriptionId}/resourceGroups/${resourceGroup}/providers/Microsoft.Compute/virtualMachines/${vmName}/restart?api-version=${apiVersion}`,
        undefined, // no request content
        undefined, // no request headers (besides auth which is handled by the token source)
        tokenSource
);
```

After:

```TS
const restartResponse = yield context.df.callHttp({
        method: "POST",
        url: `https://management.azure.com/subscriptions/${subscriptionId}/resourceGroups/${resourceGroup}/providers/Microsoft.Compute/virtualMachines/${vmName}/restart?api-version=${apiVersion}`,
        tokenSource,
});
```

Changes that were made in this PR:

* Now accepting an options argument inside of 5 different arguments
* Renamed `uri` to `url`
* Renamed `content` to `body` (I'm assuming that's what "request content" meant?)

If there are other (breaking) changes we want to make to the `callHttp` API, now is the time to call them out. For example:
* Do we want to follow `@azure/functions`'s lead and accept an `undici` `HttpHeaders` type for `headers` parameter instead of an object of strings?
* Do we want to rename the `asynchronousPatternEnabled` argument? It always felt like a confusing name to me, but this is the same the extension uses.

### DIFF
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -1,4 +1,3 @@
-import { TokenSource } from "./tokensource";
 import {
     EntityId,
     RetryOptions,
@@ -33,6 +32,7 @@ import {
 } from "./task";
 import moment = require("moment");
 import { ReplaySchema } from "./replaySchema";
+import { CallHttpOptions } from "./types";
 
 /**
  * Parameter data for orchestration bindings that can be used to schedule
@@ -301,28 +301,24 @@ export class DurableOrchestrationContext {
      *
      * @param req The durable HTTP request to schedule.
      */
-    public callHttp(
-        method: string,
-        uri: string,
-        content?: string | object,
-        headers?: { [key: string]: string },
-        tokenSource?: TokenSource,
-        asynchronousPatternEnabled = true
-    ): Task {
+    public callHttp(options: CallHttpOptions): Task {
+        let content = options.body;
         if (content && typeof content !== "string") {
             content = JSON.stringify(content);
         }
 
-        const req = new DurableHttpRequest(
-            method,
-            uri,
+        const request = new DurableHttpRequest(
+            options.method,
+            options.url,
             content as string,
-            headers,
-            tokenSource,
-            asynchronousPatternEnabled
+            options.headers,
+            options.tokenSource,
+            typeof options.asynchronousPatternEnabled === "undefined"
+                ? true
+                : options.asynchronousPatternEnabled
         );
-        const newAction = new CallHttpAction(req);
-        if (this.schemaVersion >= ReplaySchema.V3 && req.asynchronousPatternEnabled) {
+        const newAction = new CallHttpAction(request);
+        if (this.schemaVersion >= ReplaySchema.V3 && request.asynchronousPatternEnabled) {
             if (!this.defaultHttpAsyncRequestSleepTimeMillseconds) {
                 throw Error(
                     "A framework-internal error was detected: replay schema version >= V3 is being used, " +
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -313,7 +313,7 @@ export class DurableOrchestrationContext {
             content as string,
             options.headers,
             options.tokenSource,
-            typeof options.asynchronousPatternEnabled === "undefined"
+            options.asynchronousPatternEnabled === undefined
                 ? true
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -21,6 +22,40 @@ export interface OrchestrationTrigger extends FunctionTrigger {
...
+ */
+export interface CallHttpOptions {
+    /**
+     * The HTTP request method.
+     */
+    method: string;
+    /**
+     * The HTTP request URL.
+     */
+    url: string;
```
davidmrdavid, line 36: I'm not an expert on the terms here but I'm under the impression that we accept URI broadly speaking (which include URLs) and so I'd rather keep this as `uri` to denote that.

davidmrdavid: Not 100% sure if that means it would be better to keep the "new" param "body" as "content" as well. @hossam-nasr: can you check in with the wider DF team to confirm if URI is more appropiate?

davidmrdavid (commented):
tiny nitpicks, but mostly looks good :)

src/durableorchestrationcontext.ts
```diff
@@ -301,28 +301,24 @@ export class DurableOrchestrationContext {
...
+        const request = new DurableHttpRequest(
+            options.method,
+            options.url,
             content as string,
-            headers,
-            tokenSource,
-            asynchronousPatternEnabled
+            options.headers,
+            options.tokenSource,
+            typeof options.asynchronousPatternEnabled === "undefined"
```
ejizba:
```suggestion
           options.asynchronousPatternEnabled === undefined
```

src/types.ts
```diff
@@ -21,6 +22,40 @@ export interface OrchestrationTrigger extends FunctionTrigger {
...
+    /**
+     * The HTTP request body.
+     */
+    body?: string | object;
+    /**
+     * The HTTP request headers.
+     */
+    headers?: { [key: string]: string };
```
ejizba, line 44: > Do we want to follow @azure/functions's lead and accept an undici HttpHeaders type for headers parameter instead of an object of strings?

I would advocate for an "all or nothing" approach when it comes to undici. I don't want a mix of undici, axios, & our own stuff because it'll just get confusing. Unfortunately I don't think we have time for a full transition over to undici for durable, meaning I think we should leave the type as-is for now

Please keep this in mind when doing further breaking changes. If I had to choose between making a lot of breaking changes to just a few apis or applying a few breaking changes consistently to all of our apis - I would choose the latter

src/types.ts
ejizba: Node.js and our library package are consistently using url so I would support Hossam's change. If something makes durable unique, I'd be curious to know. It is call_http_, so I would think it would just be url

src/types.ts
hossam-nasr: > Please keep this in mind when doing further breaking changes. If I had to choose between making a lot of breaking changes to just a few apis or applying a few breaking changes consistently to all of our apis - I would choose the latter

I don't disagree with this, but do you mind clarifying what you mean? Are there specific breaking changes that you want to apply consistently on all our APIs that you'd rather I focus my time on?

src/types.ts
davidmrdavid: _tomato to-ma-to, url uri._
I won't block on this :)

davidmrdavid (approved):
LGTM

ejizba (approved):

src/types.ts
ejizba: For example in this PR you kind of have two breaking changes:
- rename stuff from uri to url and content to body
- Switch from lots of arguments to an options object

I'm saying I would prefer if you applied those changes to all public apis in this sdk before moving on to other breaking changes (Like the undici stuff or like removing "I" from the beginning of interfaces)

Issue #437, opened 2023-01-25
Entities do not support non-JSON serializable states


hossam-nasr:
If you try to set a durable entity state to a non-serializable object, for example a rich class, the methods of that class would not be available and your entity/orchestration would fail with an error. The simplest example of this is the `ShoppingEntity`/`ShoppingOrchestration` example in the `samples/` directory:

`ShoppingEntity/index.ts`:

```TS
import * as df from "durable-functions";
import { CartItem, ShoppingCart, Operations } from "../Shopping/data";

module.exports = df.entity<ShoppingCart>(function (context) {
  const cart = context.df.getState(() => new ShoppingCart());

  switch (context.df.operationName) {
    case Operations.ADD_ITEM:
      const cartItem = context.df.getInput<CartItem>();
      cart.addItem(cartItem);
      break;
    case Operations.REMOVE_ITEM:
      const itemToRemove = context.df.getInput<string>();
      cart.remoteItem(itemToRemove);
      break;
    case Operations.GET_VALUE:
      const value = cart.getCartValue();
      context.df.return(value);
      break;
  }

  context.df.setState(cart);
});
```

`ShoppingOrchestration/index.ts`:

```TS
import * as df from "durable-functions";
import { items, Operations } from "../Shopping/data";

module.exports = df.orchestrator(function* (context) {
  const entityId = new df.EntityId("ShoppingEntity", "shoppingCart");

  yield context.df.callEntity(entityId, Operations.ADD_ITEM, {
    itemId: items[0].itemId,
    quantity: 1,
  });
  yield context.df.callEntity(entityId, Operations.ADD_ITEM, {
    itemId: items[2].itemId,
    quantity: 2,
  });
  yield context.df.callEntity(entityId, Operations.ADD_ITEM, {
    itemId: items[1].itemId,
    quantity: 10,
  });

  const cartValue: number = yield context.df.callEntity(
    entityId,
    Operations.GET_VALUE
  );

  console.log(cartValue);
});
```
`ShoppingCart` class definition:

```TS
export class ShoppingCart {
    #items: CartItem[];

    constructor() {
        this.#items = [];
    }

    public addItem(item: CartItem): void {
        this.#items.push(item);
    }

    public remoteItem(itemId: string): void {
        const index = this.#items.findIndex((item) => item.itemId === itemId);
        this.#items = this.#items.slice(index, index + 1);
    }

    public getCartValue(): number {
        return this.#items.reduce(
            (value, item) =>
[...]

Issue #436, opened 2023-01-25
Double quotes added to input passed from suborchestrator to activity


hossam-nasr:
**Describe the bug**
If input is passed from an orchestration -> suborchestration -> activity function, there are extra double quotes added to the input. There seems to be some extra layer of serialization/deserialization happening here.

**Investigative information**

- Durable Functions extension version: 2.9.0
- durable-functions npm module version: 2.1.1
- Language (JavaScript/TypeScript) and version: TypeScript
- Node.js version: 16.13.0

**To Reproduce**

1. Create a simple orchestration that calls a suborchestration with some input, which in turn calls an activity function with the same input. Here, I am using the `SayHelloWithSubOrchestration` orchestration example in the samples:

`SayHelloWithSubOrchestration/index.js`:

```JS
const df = require("durable-functions");

module.exports = df.orchestrator(function* (context) {
    const input = context.df.getInput();

    const output = yield context.df.callSubOrchestrator("SayHelloWithActivity", input);
    return output;
});
```

`SayHelloWithActivity/index.js`:

```JS
const df = require("durable-functions");

module.exports = df.orchestrator(function* (context) {
    const input = context.df.getInput();

    const output = yield context.df.callActivity("Hello", input);
    return output;
});
```

`Hello/index.js`:

```JS
module.exports = function (context) {
    return `Hello ${context.bindings.name}!`;
};
```

2. Call the `SayHelloWithSubOrchestrator` orchestration with some input. If running locally, make the following HTTP request:

```http
POST http://localhost:7071/api/orchestrators/SayHelloWithSubOrchestrator

"Cairo"
```

**Expected behavior**
Output of the orchestration should be: `"Hello Cairo!"`

**Actual behavior**
Output of the orchestration instead is: `"Hello \"Cairo\"!"` with extra double quotes around the input.

davidmrdavid:
@hossam-nasr: Good catch. I think it would useful to see what the sub-orchestrator output looks like in the History. I suspect that sub-orchestrator results, in the History, are doubly serialized by default. Perhaps we can make a special case here to doubly de-serialize those outputs.

This would be a breaking change for V2 of the SDK, so I'd include it only in V3.

Finally, we can also investigate why the Extension leads to doubly-serialized results in sub-orchestrators and fix this there...but that's a larger change with a larger impact scope, so I wouldn't prioritize this just yet. Still worth investigating at that level to understand the root cause, and file it in the Extension repo once we understand it.

[...]

Issue #420, opened 2023-01-09
Improve user unit testing story in v3


hossam-nasr:
Related to discussion [here](https://github.com/Azure/azure-functions-durable-js/pull/399#discussion_r1020633590), and issues #257 and #160.

With v3, we have an opportunity to be more mindful about how we support users' unit testing story, and include APIs, samples, guides, documentation for unit testing

hossam-nasr:
Fixing this bug #378 should be an important piece of this work

davidmrdavid:
@hossam-nasr: Once this is addressed, we should document the unit testing strategy so that it can be a permanent fixture of the experience. I think putting it on `mslearn` would be best.

Issue #394, opened 2022-10-06
Add a smoke test for v3


hossam-nasr:
This issue is to track adding a smoke test for the build pipeline for v3, similar to #355 in v2. The smoke test step was removed in #389 as it's hard to have a finalized working function app until the new API for v3 is finalized

hossam-nasr:
Closed by #490

PR #438, opened 2023-01-26
Runnable new-and-enhanced JS samples directory for v3

### LINKED ISSUE #421: Add v3 samples/templates
Add JS and TS samples for using durable functions with the new programming model using SDK v3

hossam-nasr:
This PR tackles a few issues, mainly resolves #421, but it also addresses #357, #356, and partially #276. Also provides good scaffolding for #394. This PR migrates the old samples and adds a new, runnable, new programming model v3-compatible `samples-js` directory. This PR is only for the JS samples, but I will create another one for another, separate TS directory. I also will delete the existing samples directory in a separate PR. In addition, this PR:

* Renames functions to be simpler and more node idiomatic
* Updates/replaces outdated/maintenance/deprecated packages used in the samples, such as `moment` (replaced with `luxon`), `request` (replaced with `axios`), and the deprecated wunderground weather API (replaced with OpenWeatherMap API), and maybe others
* Leverages the ability to have multiple functions in the sample file to group related samples together

I tested all the samples manually and verified that they all work, with some notable exceptions:
* The `smsPhoneVerification` sample -- this revealed an issue on the framework level with activities having extra output bindings  -- issue here: https://github.com/Azure/azure-functions-nodejs-library/issues/53 to track
* The `sayHelloWithSubOrchestrator` sample -- there is an issue with added double quotes -- issue here #436 to track
* The `shoppingEntity`/`shoppingOrchestration` samples -- entities don't support non-JSON serializable states -- issue here #437 to track
* The unit testing sample -- I didn't migrate this one because there currently isn't a good unit testing story for the new prog model -- issue already here #420 to track
* The `listAzureSubscriptions` sample -- I couldn't test this locally

Ofc, any design/organization/naming choices made here are subject to discussion :)


samples-js/extensions.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <None Include="package.json" />
+    <None Include="local.settings.json" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.9.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.0.1" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Twilio" Version="3.0.2" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator" Version="4.0.1" />
+  </ItemGroup>
+</Project>
```
davidmrdavid, line 19: Can we remove this an just rely on extension bundles? Keeping this would mean that users need to have the .NET dependency toolchain in their machine :)

samples-js/package.json
```diff
@@ -0,0 +1,23 @@
...
+        "@azure/storage-blob": "^12.12.0",
+        "axios": "^1.2.3",
+        "durable-functions": "^3.0.0-alpha.1",
+        "luxon": "^3.2.1",
+        "readdirp": "^3.6.0"
+    },
+    "devDependencies": {
+        "chai": "^4.3.7",
+        "mocha": "^10.2.0",
+        "sinon": "^15.0.1"
```
davidmrdavid: do we really need these given that we don't have unit tests?

samples-js/functions/httpStart.js
```diff
@@ -0,0 +1,18 @@
...
+
+const clientInput = df.input.durableClient();
+
+app.http("httpStart", {
+    route: "orchestrators/{orchestratorName}",
+    extraInputs: [clientInput],
+    handler: async (request, context) => {
+        const client = df.getClient(context, clientInput);
+        const body = await request.json();
+        const instanceId = await client.startNew(request.params.orchestratorName, undefined, body);
```
davidmrdavid: We won't need this `undefined` parameter in the future, right?

samples-js/functions/httpSyncStart.js
```diff
@@ -0,0 +1,39 @@
...
+        return response;
+    },
+});
+
+function getTimeInSeconds(req, queryParameterName) {
+    const queryValue = req.query.get(queryParameterName);
+    return queryValue
+        ? queryValue * 1000 // expected to be in seconds
+        : undefined;
+}
```
davidmrdavid: Is it "expected" to be in seconds, or is it just "in seconds"?

samples-js/functions/smsPhoneVerification.js
```diff
@@ -0,0 +1,76 @@
...
+            if (winner === challengeResponseTask) {
+                // We got back a response! Compare it to the challenge code.
+                if (challengeResponseTask.result === challengeCode) {
+                    authorized = true;
+                    break;
+                }
+            } else {
+                // Timeout expired
+                break;
+            }
```
davidmrdavid: could you reverse this check so that you break if `(winner === timeoutTask)`? That would allow you to move that inner if statement one level out, which would help with readability

samples-js/functions/weatherMonitor.js
```diff
@@ -0,0 +1,129 @@
+const { output } = require("@azure/functions");
+const df = require("durable-functions");
+const { DateTime } = require("luxon");
+const axios = require("axios").default;
+
+const clearWeatherConditions = ["clear sky", "few clouds", "scattered clouds", "broken clouds"];
+
+const getIsClearActivity = "getIsClear";
+const sendGoodWeatherAlertActivity = "sendGoodWeatherAlert";
```
davidmrdavid, line 9: Just noting that I'm not a big fan of these constant strings across the file, which feel like a boilerplate. I would love it if we could have these templates benefit from any of the proposals [here](https://github.com/Azure/azure-functions-durable-js/issues/418) asap.

samples-js/functions/weatherMonitor.js
```diff
@@ -0,0 +1,129 @@
...
+                });
+
+                context.log(`Next check for ${locationString} at ${nextCheckpoint.toString()}`);
+
+                yield context.df.createTimer(nextCheckpoint.toJSDate()); // accomodate cancellation tokens
+            }
+        }
+
+        context.log("Monitor expiring.");
+    });
```
davidmrdavid: Can we sprinkle more comments here to help guide the user through this logic? Given the "height" of this method, I find it a bit difficult to read. I think comments describing logical blocks would help.

samples-js/functions/shopping.js
```diff
@@ -0,0 +1,47 @@
+const df = require("durable-functions");
+const { ShoppingCart, items } = require("./shoppingData");
+
+const entityName = "shoppingCartEntity";
+
+df.app.entity(entityName, function (context) {
+    const cart = context.df.getState(() => new ShoppingCart());
+
+    switch (context.df.operationName) {
+        case "ADD_ITEM":
```
davidmrdavid: I thought the `ShoppingCart` example wasn't working due to a serialization issue. Did it get resolved?

davidmrdavid (commented):
Leaving some preliminary comments. Thank you for this PR

samples-js/extensions.csproj
hossam-nasr: This is the only the new programming model can work right now because of #409. I want this directory to still be runnable until the fix is out, but I was planning on going back and reverting to extension bundles once #409 is fixed. I can file an issue for that too if it helps.

Also, what do you mean by users needing the dependency toolchain in their machine? Isn't this scenario covered by just `func extensions install`?

samples-js/package.json
hossam-nasr: No we don't lol, I added them when I was trying to add unit testing samples and forgot to remove them when I failed

hossam-nasr: Removed

samples-js/functions/httpStart.js
hossam-nasr: No we don't, pending #429. That being said, my general philosophy towards migrating these samples is that it is a pretty big task and a lot of things can break, so instead of waiting until _all_ the changes in the new prog model are ready, let me just get the bulk of the work done now, and have a working app with the current public release of the SDK, and then let's go back and update the samples once we have some of those updates in. Let me know if you have other thoughts or if this makes sense, but hopefully this addresses some of your other comments too

samples-js/functions/httpSyncStart.js
hossam-nasr: I think the "expected" here was referring to `queryValue` being in seconds, and it is "expected" because the user can pass in anything they want. That being said, I renamed the function name and changed the comments so hopefully it's more readable now

samples-js/functions/smsPhoneVerification.js
hossam-nasr: Done, thanks :)

samples-js/functions/weatherMonitor.js
hossam-nasr: I was planning on updating the samples once we have a conclusion and implementation on #418. Another issue I have with these string constants is that they don't work when the functions are registered in different files, unless you export and import them, which is more hassle than it might be worth. For the meantime, do you prefer keeping the constants or just having the plain strings?

samples-js/functions/weatherMonitor.js
hossam-nasr: I added some more comments, let me know if those help! Also, we don't _have_ to have all the functions in one file just because now we can. Let me know if it might make more sense/make it more readable to split each of the functions in this example into its own file!

samples-js/functions/shopping.js
hossam-nasr: No it still doesn't work. Removed now.

ejizba (approved):
Approving the overall structure of this PR. I skimmed the code, but don't feel the samples directory is worth a detailed review from me personally. As mentioned offline, my favorite thing about this PR is that it got you to test a lot of scenarios 🙂 Good work!

samples-js/functions/httpStart.js
hossam-nasr: I just updated this to the new public release `3.0.0-alpha.2` and updated this line too.

samples-js/functions/weatherMonitor.js
davidmrdavid: Let's keep the constants for now :)

samples-js/functions/weatherMonitor.js
davidmrdavid: I think it's fine as is. This is certainly a visually noisy sample, so we may want to rethink eventually, but it's fine for now

samples-js/extensions.csproj
davidmrdavid: Hmm, I'm not sure. I was under the impression that `func extensions install` might require some `dotnet` dependency on the given dev machine but now that I think about it, I have never seen it not work.

Let me put a pin on this, I realize bundles isn't viable just yet.

davidmrdavid (approved):
LGTM, although it seems you have some merge conflicts to resolve. Still, the contents of this PR seem good to me.

PR #440, opened 2023-01-27
Fix suborchestrator input double serialization issue

### LINKED ISSUE #436: Double quotes added to input passed from suborchestrator to activity

hossam-nasr:
Fixes #436

### DIFF
--- a/src/actions/callentityaction.ts
+++ b/src/actions/callentityaction.ts
@@ -10,7 +10,7 @@ export class CallEntityAction implements IAction {
         if (!entityId) {
             throw new Error("Must provide EntityId to CallEntityAction constructor");
         }
-        this.input = Utils.processInput(input);
+        this.input = input;
         Utils.throwIfEmpty(operation, "operation");
         this.instanceId = EntityId.getSchedulerIdFromEntityId(entityId);
     }
--- a/src/actions/callsuborchestratoraction.ts
+++ b/src/actions/callsuborchestratoraction.ts
@@ -10,7 +10,7 @@ export class CallSubOrchestratorAction implements IAction {
         public readonly instanceId?: string,
         input?: unknown
     ) {
-        this.input = Utils.processInput(input);
+        this.input = input;
         Utils.throwIfEmpty(functionName, "functionName");
 
         if (instanceId) {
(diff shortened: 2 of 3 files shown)


davidmrdavid:
@hossam-nasr: wouldn't this also affect the suborchestrator with-retry API? And what about entities?

hossam-nasr:
@davidmrdavid Good call out! I tested the issue with `context.df.callSubOrchestratorWithRetry()` and `context.df.callEntity()` and found that it does reproduce, so I introduced the fix to the `CallSubOrchestratorWithRetryAction` and `CallEntityAction` too

davidmrdavid (approved):
LGTM

@hossam-nasr: can you please test signalEntity as well and apply the same fix if appropiate? GIven that the fix is just a 1 liner, don't worry about re-requesting my approval if you add it to this PR

hossam-nasr:
@davidmrdavid SignalEntity doesn't seem to exist on `v3.x` 🤨 I think I branched `v3.x` out of `v2.x` before the `signalEntity()` addition was made. I will test this tho when I add it to this branch too, and I'll introduce the fix if it's required.

Issue #416, opened 2023-01-09
Improve TypeScript experience in SDK v3


hossam-nasr:
Based on discussions [here ](https://github.com/ejizba/func-nodejs-prototype/pull/14#discussion_r1014333823) and [here](https://github.com/ejizba/func-nodejs-prototype/pull/14#discussion_r1014338753), and related to issues #240 and #171.

The SDK has had some gaps in its support for TypeScript and exporting types (see #240, #171, and #352).

SDK v3 and support for the new programming model is a good chance to improve our types story and our TypeScript experience, which includes, but is not limited to:

1. Ensuring all necessary types (e.g., type for orchestration and entity contexts) are exported from the root of the package (#171)
2. Enrich and expand existing types (e.g., further define the `OrchestrationHandler` type)
3. Ensure no publicly exported types have the `I` prefix
4. Create TypeScript-specific samples and templates
5. Strongly type our TypeScript code samples and docs

hossam-nasr:
Discussed some offline. As part of this, we should also try to switch from autogeneration of `.d.ts` files, to manual authoring of them. This should would help us become more intentional about which types we export, and also address #352. To do this, we should:

1. Move all types/interfaces that we should expose publicly to the `types` directory (in normal `.ts` files)
2. Separate class implementations from declarations, and move declarations to the `types` directory (also in `.ts` files)
3. Switch to manual `.d.ts` file generation

Work for points 2 and 3 is already being done

hossam-nasr:
Closed by PRs: #446, #447, #450, #451, #470
Issues: #458, #455, #352

ejizba:
NOTE: We removed the "breaking-change" label for this particular issue because I wanted to track each breaking change individually instead of in a roll-up issue. See Hossam's links above

PR #449, opened 2023-02-01
Update OrchestrationHandler type

hossam-nasr:
Second bullet point in #416 . Updates `OrchestrationHandler` type from `(context: OrchestrationContext) => Generator<unknown, unknown, any>` to `OrchestrationHandler<T> = (context: OrchestrationContext) => Generator<Task, T, any>` (`T` being the return type of the generator)


src/types/orchestrationTypes.ts
```diff
@@ -1,13 +1,18 @@
...
-import { TokenSource } from ".";
+import { Task, TokenSource } from ".";
 import { DurableOrchestrationContext } from "../classes";
 
 /**
  * Type of a Generator that can be registered as an orchestration
+ * @param T the return value of the orchestration
  */
-export type OrchestrationHandler = (
+export type OrchestrationHandler<T = any> = (
```
ejizba: I've mentioned a few times how I'm not a fan of generic types and I'd rather remove them from durable instead of add them. Please give more examples/reasoning why this `T` is helpful if you would like to add it.

Also keeping in mind that generators have these three: `T = unknown, TReturn = any, TNext = unknown`, why is TReturn the one you are passing along but not `T` or `TNext`? This change just opens up a whole design discussion that I don't think is worth it at the moment

src/types/orchestrationTypes.ts
```diff
@@ -1,13 +1,18 @@
...
     context: OrchestrationContext
-) => Generator<unknown, unknown, any>;
+) => Generator<
+    Task, // orchestrations yield Task types
```
ejizba: By contrast to my above comment, changing `unknown` to `Task` feels more obviously beneficial and I would support getting this part of the PR merged

ejizba (changes requested):

src/types/orchestrationTypes.ts
hossam-nasr: I think your reasoning before was to only add generics if they would prevent/detect an otherwise undetectable error. I think this addition here satisfies this criteria. It allows the user to specify the expected return value of their generator, helping to detect cases where they've returned the wrong type, or for example, neglected a path that would return something unexpected. Here's an example:

```TS
// This API returns string in most cases but sometimes returns a number
const someApi = () => {
    if (Math.random() < 0.99) {
        return 'string';
    }
    return 1;
};

// Before
const orchestrator: OrchestrationHandler = function* (context: OrchestrationContext) {
    // Suppose I expect this orchestration to always return a string and call this API unwittingly
    // No error thrown here
    return someApi();
};

// After
const orchestrator: OrchestrationHandler<string> = function* (context: OrchestrationContext) {
    // Error thrown here because this is not always a string
    return someApi();
};
```

Basically, I consider specifying `T` the equivalent of writing `: T` at the end of a function declaration, except doing that is not valid syntax for Generator types:

```TS
// this is not valid syntax and will throw an error
const orchestrator: OrchestrationHandler = function* (context: OrchestrationContext): string
```

To answer your other question, `T` here is determined by the SDK and is always `Task` for valid `OrchestrationHandler`s, so that's why that change was also made. `TNext` is the result of the Durable APIs, which don't always have a consistent type, that's why I left it as `any` here. But again, it's determined by the SDK, not user code. Specifying `TNext` as, for example, `string`, means I will write an orchestration that only calls Durable APIs that return a string, which doesn't seem like a particularly common use-case, as opposed to specifying the return type.

src/types/orchestrationTypes.ts
```diff
@@ -1,13 +1,18 @@
...
+    T, // return type of the orchestration
+    any // what the SDK passes back to the orchestration
```
ejizba: &lt;continuing the TNext discussion on a separate thread&gt;

> `TNext` is the result of the Durable APIs, which don't always have a consistent type, that's why I left it as `any` here. But again, it's determined by the SDK, not user code. Specifying `TNext` as, for example, `string`, means I will write an orchestration that only calls Durable APIs that return a string, which doesn't seem like a particularly common use-case, as opposed to specifying the return type.

Can you give a few of the most common examples of durable apis and what the type of TNext would be?

src/types/orchestrationTypes.ts
ejizba: I appreciate your example - that's exactly the kind of thing I'm looking for thanks. Will we be able to specify OrchestrationHandler&lt;something&gt; in our templates? What about any of our samples? I'm trying to gauge how realistically helpful this is vs theoretically helpful

src/types/orchestrationTypes.ts
```diff
@@ -1,13 +1,18 @@
...
  * Type of a Generator that can be registered as an orchestration
+ * @param T the return value of the orchestration
  */
-export type OrchestrationHandler = (
+export type OrchestrationHandler<T = any> = (
     context: OrchestrationContext
-) => Generator<unknown, unknown, any>;
+) => Generator<
+    Task, // orchestrations yield Task types
+    T, // return type of the orchestration
```
ejizba: I feel like this should be `TReturn` to match the naming of `Generator`

src/types/orchestrationTypes.ts
hossam-nasr: I think this change could be applied to all of our templates/samples, yes. For example, here's what the simple hello orchestration can look like now:

```TS
const orchestrator: OrchestrationHandler<string[]> = function* (context: OrchestrationContext) {
    const outputs = [];
    outputs.push(yield context.df.callActivity(activityName, 'Tokyo'));
    outputs.push(yield context.df.callActivity(activityName, 'Seattle'));
    outputs.push(yield context.df.callActivity(activityName, 'Cairo'));

    return outputs;
};
```

If we merge this, I think we should update our TS templates/samples to follow this pattern too.

src/types/orchestrationTypes.ts
hossam-nasr: The simplest example would be the `context.df.callActivity()` API, in which case the `TNext` type would be whatever was the return type of the Activity, so that API alone could be anything, and could vary from one call to the next.

```TS
const orchestrator: OrchestrationHandler = function* (context: OrchestrationContext) {
    yield context.df.callActivity('returnsString'); // type of TNext here is string
    yield context.df.callActivity('returnsNumber'); // type of TNext here is number
    yield context.df.callHttp({ method: 'get', url: 'http://someApi' }); // type of TNext here is HttpResponse
};

df.app.activity('returnsString', {
    handler: (): string => {
        return 'string';
    },
});

df.app.activity('returnsNumber', {
    handler: (): number => {
        return 1;
    },
});
```

hossam-nasr: But I want to be clearer, that it's SDK code, not user code, that is affected by the type of `TNext`, because it's SDK code that calls `.next()` on the generator in this line here: https://github.com/Azure/azure-functions-durable-js/blob/4b4eae8a5dc9b1571be6cdfda229b7ae2d0fba4a/src/taskorchestrationexecutor.ts#L388-L390

So `TNext` specifies what type `taskValue` is allowed to be in the `this.generator.next(taskValue)` line, and it can really be anything.

src/types/orchestrationTypes.ts
hossam-nasr: Done

src/types/orchestrationTypes.ts
```diff
@@ -1,13 +1,18 @@
...
 /**
  * Type of a Generator that can be registered as an orchestration
+ * @param TReturn the type of the return value of the orchestration
  */
-export type OrchestrationHandler = (
+export type OrchestrationHandler<TReturn = any> = (
     context: OrchestrationContext
-) => Generator<unknown, unknown, any>;
+) => Generator<
+    Task, // orchestrations yield Task types
```
davidmrdavid: nitpick: can we add a "only" yields Task types?
```suggestion
    Task, // orchestrations yield only Task types
```

davidmrdavid (commented):
I left the world's smallest nitpick.
The PR actually looks good to me, but I see there's some discussion still in the comments so I'll refrain from approving for now. I'm available to help clarify the usage of these types as well. In general, I'd recommend merging this change.

src/types/orchestrationTypes.ts
davidmrdavid: Yep, that's right. For us `TNext` represents an external result (over the network/an external Function invocation), so it would be peculiar if a user restricted themselves to only receiving a specific type here. The type `any` makes sense here because we really can't control / type-check what comes over the network.

src/types/orchestrationTypes.ts
ejizba: Hossam and I discussed this offline and I'll summarize here for @davidmrdavid and other's sake:

I have two main roadblocks to approval at the moment
1. It bothers me that our generics are close but not quite the same as `Generator`. The type for `Generator` is coming straight from Node.js core types and I trust Node.js designers more than I trust us. If we decide generics are worth it, I would prefer we match all three from `Generator`. Something like this:
    ```TS
    export type OrchestrationHandler<T extends Task = Task, TReturn = any, TNext = any> ...
    ```
2. As mentioned by Hossam, I usually only like generics "if they would prevent/detect an otherwise undetectable error". Initially I found Hossam's example compelling because he said you couldn't specify the return type `: string[]` on a generator, but after messing around I realized you _can_ accomplish the same thing without generics. These two are equivalent:
    ```TS
    const orchestrator: OrchestrationHandler<string[]> = function* (context: OrchestrationContext) {
    const orchestrator = function* (context: OrchestrationContext): Generator<Task, string[], any> {
    ```

Generics have caused me a lot of pain in the past and once you add them you can't take them back without it being a breaking change. That's why I'm strict about it. We can change our mind a month from now if we get user feedback asking for generics. We can't change our mind a month from now if the generics start causing problems.

src/types/orchestrationTypes.ts
ejizba: For the sake of reference, here's the orchestration handlers as of today:
```TS
const orchestrator: OrchestrationHandler = function* (context: OrchestrationContext) {
    const outputs = [];
    outputs.push(yield context.df.callActivity(activityName, 'Tokyo'));
    outputs.push(yield context.df.callActivity(activityName, 'Seattle'));
    outputs.push(yield context.df.callActivity(activityName, 'Cairo'));

    return outputs;
};

const helloActivity: ActivityHandler = (input: string): string => {
    return `Hello, ${input}`;
};
```
Our own orchestration template is only yielding `callActivity` on `helloActivity` which is the user's own code and always returns a string. Doesn't that mean `string` would be a logical value for `TNext` in this case? The other example Hossam gave me was `callHttp`, but even in that case we know the type would always be `HttpResponse`

It's not just about `TNext`. Hossam suggested we specify `TReturn` as `string[]` for the template, but that doesn't do anything (in terms of build errors, preventing bugs, etc.) unless we also change the type of `outputs` which is currently `any[]`. And what does the type of `outputs` depend on? `TNext`. I think `TReturn` and `TNext` are inextricably linked and I think the Node.js designers probably saw that when designing the type.

src/types/orchestrationTypes.ts
hossam-nasr: Done, thanks :)

src/types/orchestrationTypes.ts
davidmrdavid: @ejizba:

For (1): our generator is only allowed to yield `Task` types, because we use `yield` as a proxy for `await`. So I think it would be a bit weird to ask users to always write `Generator<Task, ...>`in their type. Guiding users to `yield` only Tasks is the main benefit of this change, _imo_. I would be willing to compromise and allow users to parametrize `TNext` in addition to `TReturn`, but I don't think we should revert all the way to the default `Generator` type because this is a specialization of that type.

(2) I trust your experience here regarding breaking changes. Let's not rush this decision.

Given the discussion, I think there's a middle ground we could pivot to. See below.

@hossam-nasr / @ejizba : Can we consider limiting this change to simply refining `OrchestrationHandler` to be `Generator<Task, unknown, any>`? In other words, we remove the previous `unknown` with `Task` to help users guarantee that they only `yield` Tasks. Then, I believe we should be able to add generics via an overload. We can track the specific overload discussion as a post preview item. Does that seem reasonable?

ejizba: > Can we consider limiting this change to simply refining `OrchestrationHandler` to be `Generator<Task, unknown, any>`?

Yes, I am fully on board with doing this change by itself

> So I think it would be a bit weird to ask users to always write `Generator<Task, ...>`in their type.

Was your comment in regard to this suggestion?

`const orchestrator = function* (context: OrchestrationContext): Generator<Task, string[], any> {`

To me, specifying `Generator<Task, ...>` as the return type to `function*` is analogous to specifying `Promise<...>` as the return type to `async function`. Yes, the `Promise` is implied and you don't _have_ to specify a return type, but it's still best practice to do so.

hossam-nasr:
For the sake for getting this to move forward before preview with the most benefit possible, I've updated this so that the only change is adding the `Task` type as the only allowed yield type, while leaving `TReturn` as `unknown` and `TNext` as `any`. I've created an issue here though: #488 to track possibly adding the `TReturn` generic for GA

ejizba (approved):

Issue #352, opened 2022-04-26
Limit amount of type (`d.ts`) files exposed to users


davidmrdavid:
After a discussion with @ejizba, we've learned that our TS project exports type-declaration files even for our internal utility files. This is problematic because it increases the type-management burden of our end users, and makes it inconvenient for us to import and reference external types in our project.

We should look to limit our exported `*.d.ts` files to include only public types.

davidmrdavid:
FYI @hossam-nasr

hossam-nasr:
Closed by #470

PR #450, opened 2023-02-01
Add DurableClient class declaration

### LINKED ISSUE #154: Rename DurableOrchestrationClient to DurableClient
Now that `DurableOrchestrationClient` supports entity operations in Durable 2.x, we should rename it to `DurableClient`. This also makes it consistent with the binding type in function.json.

To make this a non-breaking change, we should continue to support `DurableOrchestrationClient`, but mark it as deprecated. I'm not sure what the official way to do this is in TypeScript, but apparently you do this in JS using the [@deprecated](https://jsdoc.app/tags-deprecated.html) JSDoc syntax.

hossam-nasr:
Resolves #154, part of #416. Adds a `DurableClient` class declaration that the `DurableOrchestrationClient` class implements, and exports it from the root of the package.

```TS
// before
const client = df.getClient(context)
// client. -> no intellisense

// after
const client: DurableClient = df.getClient(context);
// client. -> all the intellisense ✨
```

### DIFF
--- a/src/types/durableClientTypes.ts
+++ b/src/types/durableClientTypes.ts
@@ -20,6 +20,9 @@ export declare class DurableClient {
      */
     readonly taskHubName: string;
 
+    /** @hidden */
+    public readonly uniqueWebhookOrigins: string[];
+
     /**
      * Creates an HTTP response that is useful for checking the status of the
      * specified instance.
--- a/src/types/durableClientTypes.ts
+++ b/src/types/durableClientTypes.ts
@@ -1,7 +1,224 @@
-import { FunctionInput } from "@azure/functions";
+import { FunctionInput, HttpRequest, HttpResponse } from "@azure/functions";
+import { EntityId, EntityStateResponse, HttpManagementPayload, IHttpResponse } from "../classes";
+import { DurableOrchestrationStatus } from "../durableorchestrationstatus";
+import { IHttpRequest } from "../ihttprequest";
+import { OrchestrationRuntimeStatus } from "../orchestrationruntimestatus";
+import { PurgeHistoryResult } from "../purgehistoryresult";
 
 export interface DurableClientInput extends FunctionInput {
     type: "durableClient";
+}
+
+/**
+ * Client for starting, querying, terminating and raising events to
+ * orchestration instances.
+ */
+export interface DurableClient {
+    /**
+     * The name of the task hub configured on this orchestration client
+     * instance.
+     */
+    readonly taskHubName: string;
+
+    /**
+     * Creates an HTTP response that is useful for checking the status of the
+     * specified instance.
+     * @param request The HTTP request that triggered the current orchestration
+     *  instance.
+     * @param instanceId The ID of the orchestration instance to check.
+     * @returns An HTTP 202 response with a Location header and a payload
+     *  containing instance management URLs.
+     */
+    createCheckStatusResponse(
+        request: IHttpRequest | HttpRequest | undefined,
+        instanceId: string
+    ): HttpResponse;
+
+    /**
+     * Creates an [[HttpManagementPayload]] object that contains instance
+     * management HTTP endpoints.
+     * @param instanceId The ID of the orchestration instance to check.
+     */
+    createHttpManagementPayload(instanceId: string): HttpManagementPayload;
+
+    /**
+     * Gets the status of the specified orchestration instance.
+     * @param instanceId The ID of the orchestration instance to query.
+     * @param showHistory Boolean marker for including execution history in the
+     *  response.
+     * @param showHistoryOutput Boolean marker for including input and output
+     *  in the execution history response.
+     */
+    getStatus(
+        instanceId: string,
+        showHistory?: boolean,
+        showHistoryOutput?: boolean,
+        showInput?: boolean
+    ): Promise<DurableOrchestrationStatus>;
+
+    /**
+     * Gets the status of all orchestration instances.
+     */
+    getStatusAll(): Promise<DurableOrchestrationStatus[]>;
+
+    /**
+     * Gets the status of all orchestration instances that match the specified
+     * conditions.
+     * @param createdTimeFrom Return orchestration instances which were created
+     *  after this Date.
+     * @param createdTimeTo Return orchestration instances which were created
+     *  before this DateTime.
+     * @param runtimeStatus Return orchestration instances which match any of
+     *  the runtimeStatus values in this array.
+     */
+    getStatusBy(
+        createdTimeFrom: Date | undefined,
+        createdTimeTo: Date | undefined,
+        runtimeStatus: OrchestrationRuntimeStatus[]
+    ): Promise<DurableOrchestrationStatus[]>;
+
+    /**
+     * Purge the history for a specific orchestration instance.
+     * @param instanceId The ID of the orchestration instance to purge.
+     */
+    purgeInstanceHistory(instanceId: string): Promise<PurgeHistoryResult>;
+
+    /**
+     * Purge the orchestration history for instances that match the conditions.
+     * @param createdTimeFrom Start creation time for querying instances for
+     *  purging.
+     * @param createdTimeTo End creation time for querying instances for
+     *  purging.
+     * @param runtimeStatus List of runtime statuses for querying instances for
+     *  purging. Only Completed, Terminated or Failed will be processed.
+     */
+    purgeInstanceHistoryBy(
+        createdTimeFrom: Date,
+        createdTimeTo?: Date,
+        runtimeStatus?: OrchestrationRuntimeStatus[]
+    ): Promise<PurgeHistoryResult>;
+
+    /**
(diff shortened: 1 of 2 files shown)


src/types/durableClientTypes.ts
```diff
@@ -1,9 +1,226 @@
...
 
 export interface DurableClientInput extends FunctionInput {
     type: "durableClient";
 }
 
+/**
+ * Client for starting, querying, terminating and raising events to
+ * orchestration instances.
+ */
+export interface DurableClient {
```
ejizba: > ```
> // before
> const client = df.getClient(context)
> // client. -> no intellisense
> ```

This should still give intellisense. You don't need `: DurableOrchestrationClient` to get intellisense because the language service is smart enough to figure out the type of `client` based on the return type of `getClient`.

I think there's a few breaking changes here:
1. Changing the name from `DurableOrchestrationClient` to `DurableClient`
2. Changing the return of `getClient` from a class to an interface (affecting things like `instanceof`)

With all breaking changes that we're doing, could we get some reasoning on why these are worth it? If we're not sure, let's err on the side of maintaining the old behavior.

hossam-nasr: Does this still count as a breaking change when users couldn't access the `DurableOrchestrationClient` class from the root of the package? Before, to do something like `instance of`, you had to import it like this:

```TS
import { DurableOrchestrationClient } from 'durable-functions/lib/src/durableorchestrationclient';

if (client instanceof DurableOrchestrationClient)  {
     // do something
}
```

Regardless, here's how I would justify these changes:
1. The name "orchestration client" is outdated, since now durable clients can interact with both entities and orchestrations. This is why the name of the input binding type was also updated from `orchestrationClient` to `durableClient`. There's also an issue #154 to update the name of this class from `DurableOrchestrationClient` to `DurableClient`, so now is a good chance to do that :)
2. I feel less strongly about this one, but I just figured this offers cleaner separation between API declarations and implementation.

ejizba: > Does this still count as a breaking change when users couldn't access the DurableOrchestrationClient class from the root of the package? Before, to do something like instance of, you had to import it like this:

Yes. The code worked before and will break with these changes - that's about as simple as it gets. There are rare exceptions which we could discuss offline if you want 

1. I'm fine with the change given that reasoning, but we should handle it properly. I'm glad we already have an issue - please label it "v3"/"breaking", link it to this PR in the description, and make sure it gets closed when this PR is merged.
2. I'm fine with moving it to a different file to separate API declarations from implementation. You can keep it as a class though. Instead of `export interface` just do `export declare class`

davidmrdavid: My 2 cents:
On (1): Yes I agree we should support the name `DurableClient`. However, [this](https://github.com/Azure/azure-functions-durable-js/issues/154) ticket asks for it to be implemented in a non-breaking way so I will have to ask for the same. Can we use the good old trick of exposing the new name while keeping the old one with a `@deprecated` flag?

(2) +1 to keep it as a class.

hossam-nasr: On 1. @ejizba I've done as you requested, but I will also add that this issue was filed before the programming model and this new v3.x release was a thing, so I'm not sure if we should still keep the issue to consider doing this (in a non-breaking way) in v2.x too. Relatedly, @davidmrdavid I'd say maybe we should do this in a non-way breaking in v2.x if we want to, but I'd rather not keep the old name around for this release since it wasn't even exported from the root of the package and it wasn't in any of our docs. It's _technically_ a breaking change, but I don't personally feel bad about breaking this one.

On 2. done. I've converted this to a class declaration instead of an interface.

davidmrdavid: @hossam-nasr: Your reasoning for (1) makes sense to me. I'm ok with using the new name for v3 and having backwards-compatibility in v2, as you suggested.

src/types/durableClientTypes.ts
```diff
@@ -1,9 +1,229 @@
...
+import { OrchestrationRuntimeStatus } from "../orchestrationruntimestatus";
+import { PurgeHistoryResult } from "../purgehistoryresult";
 
 export interface DurableClientInput extends FunctionInput {
     type: "durableClient";
 }
 
+/**
+ * Client for starting, querying, terminating and raising events to
+ * orchestration instances.
```
davidmrdavid: nitpick: also for entities
```suggestion
 * orchestration and entities.
```

davidmrdavid (approved):
LGTM. From the discussion, it would appear we are ok with this change.

src/types/durableClientTypes.ts
```diff
@@ -1,9 +1,229 @@
...
+ */
+export declare class DurableClient {
+    /**
+     * The name of the task hub configured on this orchestration client
+     * instance.
+     */
+    readonly taskHubName: string;
+
+    /** @hidden */
+    public readonly uniqueWebhookOrigins: string[];
```
ejizba, line 24: Any particular reason you added this back? I guess I don't fully understand how the "hidden" tag works, but I assumed as a part of manually generating the types that we wouldn't see any hidden fields

src/types/durableClientTypes.ts
```diff
@@ -1,9 +1,229 @@
...
+
```
ejizba, line 25: I think this PR should keep the constructor in the types to match the auto-generated types. If we decide to remove the constructor, I think that should be a separate breaking change.

ejizba (changes requested):

src/types/durableClientTypes.ts
hossam-nasr: Yes, the reason was I found this referenced in the tests, so I figured there might be some use-case where this is needed for some reason. I'm not sure how the `@hidden` tags work either, but to not change the old behavior, I kept the `@hidden` tags.

hossam-nasr: Also, per your other suggestions, shouldn't we lean on just keeping them as-is for now to match the auto-generated declarations, and remove them in a separate change if we decide to?

src/types/durableClientTypes.ts
hossam-nasr: This is the type signature of the constructor:

```TS
constructor(
        state: HistoryEvent[],
        instanceId: string,
        currentUtcDateTime: Date,
        isReplaying: boolean,
        parentInstanceId: string | undefined,
        longRunningTimerIntervalDuration: string | undefined,
        maximumShortTimerDuration: string | undefined,
        defaultHttpAsyncRequestSleepTimeMillseconds: number | undefined,
        schemaVersion: ReplaySchema,
        input: unknown,
        private taskOrchestratorExecutor: TaskOrchestrationExecutor
    );
```

Exposing the `constructor()` would also force us to expose other types like `HistoryEvent` (which in turn exposes `HistoryEventType`), `ReplaySchema`, `TaskOrchestrationExecutor` (which in turn also exposes `TaskBase` and other internal classes), etc. In the interest of #352, and in the spirit of this change to let us more intentionally the public APIs, I think we should omit the constructor. Users shouldn't be using this constructor anyway.

src/types/durableClientTypes.ts
hossam-nasr: Done

src/types/durableClientTypes.ts
ejizba: I was lumping hidden fields together with private fields in my head. Both are technically accessible from JavaScript regardless, but the auto-generated types made it pretty clear they shouldn't be used. If you feel we should leave hidden tags in, I'm okay with that

src/types/durableClientTypes.ts
hossam-nasr: Sorry, this is actually the signature for the `DurableOrchestrationContext`. Let me add the constructor to the `DurableOrchestrationClient` (the subject of this PR).

ejizba: I see what you're saying, but I still think this is a breaking change and needs its own discussion. Can we add the constructor back for this PR without actually needing to expose all the other types yet? I feel like that should be possible

hossam-nasr: Fair enough. Yeah, it will be possible because this is all in `.ts` files for now anyway.

hossam-nasr: Added the constructor type without exposing all the necessary types too

src/types/durableClientTypes.ts
hossam-nasr: I think the difference is `@hidden` fields can be accessed by TypeScript code too. It's just hidden from documentation. I'm okay removing `@hidden` fields in general, but I'd also mark that as a breaking change. This particular field though I think we should keep because I found it used in the `getClient()` tests here: https://github.com/Azure/azure-functions-durable-js/blob/4b4eae8a5dc9b1571be6cdfda229b7ae2d0fba4a/test/unit/getclient-spec.ts#L101 

I'm not sure if this was supposed to be testing some internal thing, or if this in itself is a valid use-case, but either way, it feels breaking to remove

ejizba (approved):

Issue #422, opened 2023-01-12
Remove the `df` parameter from orchestration contexts


hossam-nasr:
Currently, all orchestration APIs are provided under a `df` parameter inside the `Context` object, e.g., `context.df.callActivity()`, `context.df.callSubOrchestration()`, `context.df.currentUtcDateTime`, etc.

Since all these durable APIs are a lot more relevant to the user than the other metadata (coming from Azure Functions directly) present in the context object, this issue is to explore whether we should provide the APIs directly in the Context object rather than nested inside a `df` property. We could also define our own type of orchestration context, picking only the relevant elements from the Functions context, instead of extending it.

hossam-nasr:
There are 3 options I can see here: 

1. Keep it as-is:

Durable APIs live under a `df` object, and `OrchestrationContext`/`EntityContext` types extend `InvocationContext`

2. Spread the Durable APIs in the main object:

Durable contexts (`OrchestrationContext`/`EntityContext`) still extend `InvocationContext`, but the Durable APIs live directly on the context object, rather than under one `df` object.

3. Flip the order:

Durable contexts contain the Durable APIs directly, but then add an `invocationContext` parameter that contains the `InvocationContext` object. In that way, Durable contexts would behave similarly to execution hook contexts.

davidmrdavid:
@hossam-nasr: From my understanding, we decided to keep this as-is for now. Is that correct?

hossam-nasr:
@davidmrdavid Yes, I believe we decided to skip this for preview. Removed the preview label

PR #451, opened 2023-02-01
Add orchestration and entity context type declarations

hossam-nasr:
Part of #416:

* Adds a `DurableOrchestrationContext` class declaration exported from the root of the package that the `DurableOrchestrationContext` class implements. This is the type of the `context.df` object in orchestrations.
* Converts `DurableEntityContext` type into an interface, and also exports it from the root of the package. This is the type of the `context.df` object in entity functions


src/durableorchestrationcontext.ts
```diff
@@ -30,13 +30,18 @@ import {
...
+    TimerTask,
+    DurableOrchestrationContext as DurableOrchestrationContextInterface,
+} from "./types";
 
 /**
  * Parameter data for orchestration bindings that can be used to schedule
  * function-based activities.
  */
-export class DurableOrchestrationContext {
+export class DurableOrchestrationContext implements DurableOrchestrationContextInterface {
```
davidmrdavid: Seems we removed the DurableEntityContext file. Why?

davidmrdavid (commented):
PR makes sense to me except for the deletion of the entity context:

src/durableorchestrationcontext.ts
hossam-nasr: It was all just moved to the `entityTypes.ts` file, to get us one step closer to #458. The plan is to have all the types exported from the `types` directory, then convert the `.ts` files there into `.d.ts` files, and that will now become the only source `.d.ts` files.

davidmrdavid (approved):
🚢 it

src/testingUtils.ts
```diff
@@ -8,8 +8,7 @@ import {
 } from "./classes";
 import { ReplaySchema } from "./replaySchema";
 import * as uuidv1 from "uuid/v1";
-import { DurableEntityContext } from "./durableentitycontext";
-import { EntityContext, OrchestrationContext } from "./types";
+import { DurableEntityContext, EntityContext, OrchestrationContext } from "./types";
```
davidmrdavid, line 11: I just realized that it's a bit confusing to have a `DurableEntityContext` and an `EntityContext` type. What is the difference between the two again?

src/durableorchestrationcontext.ts
```diff
@@ -31,13 +31,14 @@ import {
...
 import { CallHttpOptions, Task, TimerTask } from "./types";
+import * as types from "./types";
 import { SignalEntityAction } from "./actions/signalentityaction";
 
 /**
  * Parameter data for orchestration bindings that can be used to schedule
  * function-based activities.
  */
-export class DurableOrchestrationContext {
+export class DurableOrchestrationContext implements types.DurableOrchestrationContext {
```
davidmrdavid, line 41: so there's two `DurableOrchestrationContext` types now? I'm a bit confused by that.

src/types/entityTypes.ts
```diff
@@ -20,3 +20,82 @@ export interface EntityContext<T> extends InvocationContext {
      */
     df: DurableEntityContext<T>;
 }
+
+/**
+ * Provides functionality for application code implementing an entity operation.
+ * @param TState the JSON-serializable type of the Entity's state
+ */
+export interface DurableEntityContext<TState> {
```
davidmrdavid, line 46: I noticed this used to be an `export type` but now is `export interface`. Can you please clarify the difference?
Also, why does this export an `interface` where the orchestrator-context exports a `class`?

davidmrdavid (changes requested):
Actually, on closer inspection, I have a few more questions :)

src/testingUtils.ts
hossam-nasr: The `EntityContext` type is the type of the `context` argument passed to entity functions, and `DurableEntityContext` is the type of the `df` object on the `EntityContext`. If you have better suggestions for the naming, I'm open to it, but changing `DurableEntityContext` would be a breaking change. Ofc, this can be avoided if we implement #422 😛

src/durableorchestrationcontext.ts
hossam-nasr: No, this file hosts the implementation of the `DurableOrchestrationContext` class, which `implements` the `DurableOrchestrationContext` _declaration_ in the `types` directory. It's declaration/interface vs implementation. The plan is that the class declaration defined in the `types` directory will become our public "contract" for the API provided by the `DurableOrchestrationContext`, which _excludes_ things like private fields and methods, otherwise included in the automatically generated `.d.ts` files

src/types/entityTypes.ts
hossam-nasr: For this use-case, there are no relevant differences. I'm not sure what the subtle differences between `type` and `interface` in TypeScript (at least `>v2.1`) are, but for defining object/class structure, they're effectively the same. However, it feels to me at least that using `interface` for this use-case is more idiomatic, when defining the structure of an object. I think the same also applies to the earlier `DurableClient` and `DurableOrchestrationContext` classes, and if it was up to me, I'd rather change those to interfaces too. One argument is that you, as the user, _shouldn't_ be calling the constructor of those types. However, changing classes to interfaces _would_ be a breaking change, which is why I haven't done that here, but changing a `type` to an `interface` isn't.

src/testingUtils.ts
davidmrdavid: I see. Yeah the naming here is not ideal but I'm not ready to change this at this time :) . Let's put a pin on this, but thanks for clarifying!

davidmrdavid (approved):
LGTM after discussion

src/types/entityTypes.ts
ejizba: I agree with David's sentiment here and would prefer if we avoid changes like this during two scenarios:
1. When dealing with our public contract. If the risk and the value are equal (both low in this case), I err on the side of caution
2. During PRs that are mostly just moving code. These PRs are difficult to review, so even minor changes should be avoided

That being said, I don't have a specific example of when switching a type to an interface would cause a breaking change so I won't block on this PR. Please consider this feedback in future PRs, though

src/types/orchestrationTypes.ts
```diff
@@ -65,3 +65,224 @@ export interface CallHttpOptions {
      */
     asynchronousPatternEnabled?: boolean;
 }
+
+/**
+ * Provides functionality for application code implementing an orchestration operation.
+ */
+export declare class DurableOrchestrationContext {
```
ejizba, line 94: Same feedback as https://github.com/Azure/azure-functions-durable-js/pull/450#discussion_r1100574673

This comment is the only reason I'm requesting changes

src/types/orchestrationTypes.ts
```diff
@@ -65,3 +65,224 @@ export interface CallHttpOptions {
...
+     *
+     * @param input The JSON-serializable data to re-initialize the instance with.
+     */
+    continueAsNew(input: unknown): void;
+
+    /**
+     * Creates a durable timer that expires at a specified time.
+     *
+     * All durable timers created using this method must either expire or be
+     * cancelled using `TimerTask.cancel()` before the orchestrator
```
ejizba, line 267: I appreciate that you cleaned up the docs in several places, but in the future please avoid _any_ change that you possibly can when moving code. This is to speed up the review of these PRs and ensure things don't slip in on accident

ejizba (changes requested):

src/types/entityTypes.ts
hossam-nasr: Sounds good, I appreciate the feedback, thank you!

src/types/orchestrationTypes.ts
hossam-nasr: This is the type signature of the constructor:

```TS
constructor(
        state: HistoryEvent[],
        instanceId: string,
        currentUtcDateTime: Date,
        isReplaying: boolean,
        parentInstanceId: string | undefined,
        longRunningTimerIntervalDuration: string | undefined,
        maximumShortTimerDuration: string | undefined,
        defaultHttpAsyncRequestSleepTimeMillseconds: number | undefined,
        schemaVersion: ReplaySchema,
        input: unknown,
        private taskOrchestratorExecutor: TaskOrchestrationExecutor
    );
```

Exposing the `constructor()` would also force us to expose other types like `HistoryEvent` (which in turn exposes `HistoryEventType`), `ReplaySchema`, `TaskOrchestrationExecutor` (which in turn also exposes `TaskBase` and other internal classes), etc. In the interest of #352, and in the spirit of this change to let us more intentionally the public APIs, I think we should omit the constructor. Users shouldn't be using this constructor anyway.

src/types/orchestrationTypes.ts
hossam-nasr: Sounds good, thank you for the feedback.

ejizba (approved):

src/types/orchestrationTypes.ts
hossam-nasr: Added constructor, will file issue to remove

Issue #441, opened 2023-01-30
Switch APIs from using discrete optional arguments to using an options argument where it makes sense


hossam-nasr:
Based on discussion [here](https://github.com/Azure/azure-functions-durable-js/pull/435#discussion_r1088287211). This issue is a generalization of #415 and #435, to switch all of our APIs from using optional discrete arguments to accepting an options object argument, where applicable and where it makes sense.

hossam-nasr:
APIs affected/proposed:

**In DurableOrchestrationClient:**

<table>
<tr>
<th> Before </th>
<th> Proposed change </th>
</tr>
<tr>
<td>

```TS
getStatus(
    instanceId: string,
    showHistory?: boolean,
    showHistoryOutput?: boolean,
    showInput?: boolean
): Promise<DurableOrchestrationStatus>
```
</td>
<td>

```TS
getStatus(
    instanceId: string, 
    options?: GetStatusOptions
): Promise<DurableOrchestrationStatus>
```
</td>
</tr>
<tr>
<td>

```TS
getStatusBy(
    createdTimeFrom: Date | undefined,
    createdTimeTo: Date | undefined,
    runtimeStatus: OrchestrationRuntimeStatus[]
): Promise<DurableOrchestrationStatus[]>
```

</td>
<td>

```TS
getStatusBy(
    options: SelectionOptions
): Promise<DurableOrchestrationStatus[]>
```

</td>
</tr>
<tr>
<td>

```TS
purgeInstanceHistoryBy(
    createdTimeFrom: Date,
    createdTimeTo?: Date,
    runtimeStatus?: OrchestrationRuntimeStatus[]
): Promise<PurgeHistoryResult>
```

</td>
<td>

```TS
purgeInstanceHistoryBy(
    options: SelectionOptions
): Promise<PurgeHistoryResult>
```

</td>
</tr>
<tr>
<td>

```TS
raiseEvent(
    instanceId: string,
    eventName: string,
    eventData: unknown,
    taskHubName?: string,
    connectionName?: string
): Promise<void>
```

</td>
<td>

```TS
raiseEvent(options: RaiseEventOptions): Promise<void>
```

</td>
</tr>
<tr>
<td>

```TS
readEntityState<T>(
    entityId: EntityId,
    taskHubName?: string,
    connectionName?: string
): Promise<EntityStateResponse<T>>
```

</td>
<td>

```TS
readEntityState<T>(
    entityId: EntityId,
[...]

PR #452, opened 2023-02-01
Rename `GetStatusOptions` to `GetStatusInternalOptions`

hossam-nasr:
`GetStatusOptions` is an internal class and was only used to be passed to the private function `getStatusInternal()` on the `DurableOrchestrationClient` class. This change is mostly to unblock #441 and reserve the `GetStatusOptions` name for the new argument to the public `getStatus()` API


src/getstatusinternaloptions.ts
```diff
@@ -1,16 +1,16 @@
...
-    createdTimeTo?: Date;
-    runtimeStatus?: OrchestrationRuntimeStatus[];
-    showInput?: boolean;
-}
+// tslint:disable:member-access
+
+import { OrchestrationRuntimeStatus } from "./classes";
+
+/** @hidden */
+export class GetStatusInternalOptions {
```
ejizba: This one should def be an interface lol (especially since it's internal). And I'd move it directly into `durableorchestrationclient.ts`

ejizba (approved):

src/getstatusinternaloptions.ts
hossam-nasr: Done

PR #460, opened 2023-02-03
Add enablePolling property to CallHttpOptions

### LINKED ISSUE #439: Should we rename the `callHttp()` asynchronousPatternEnabled property?
This property defaults to true enables the polling behavior for `context.callHttp()` in orchestrations. The name sounds arcane though, but it's the name used by the durable extension. Should we use the v3 opportunity to rename it to a more descriptive name, e.g., `enablePolling`?

hossam-nasr:
Resolves #439. 

* Adds an `enablePolling` property to `CallHttpOptions` (accepted as an argument to `context.df.callHttp()`)
* Adds a `@deprecated` flag to the `asynchronousPatternEnabled` property, directing users to use the new name
* Updates the `callHttp()` implementation so that `enablePolling` takes precedence, followed by `asynchronousPatternEnabled`, while still defaulting to `true` if none are set.

### DIFF
--- a/src/types/orchestrationTypes.ts
+++ b/src/types/orchestrationTypes.ts
@@ -52,24 +52,16 @@ export interface CallHttpOptions {
      */
     tokenSource?: TokenSource;
     /**
-     * Specifies whether the Durable extension should continue
-     * polling the request after receiving a 202 response.
-     *
-     * This porperty name replaces the previous `asynchronousPatternEnabled` argument.
-     * If both are specified, this property takes precedence.
+     * Specifies whether to continue polling the request after receiving a 202 response.
+     * This replaces `asynchronousPatternEnabled`. If both are specified,
+     * `enablePolling` takes precedence.
      *
      * @default true
      */
     enablePolling?: boolean;
     /**
-     * Specifies whether the Durable extension should continue
-     * polling the request after receiving a 202 response.
-     *
-     * @deprecated this property name is deprecated.
-     * Please use the new `enablePolling` property instead.
-     * If both are specified, the `enablePolling` property takes precedence.
-     *
-     * @default true
+     * @deprecated use `enablePolling` instead. If both are specified,
+     * `enablePolling` takes precedence.
      */
     asynchronousPatternEnabled?: boolean;
 }
--- a/src/types/orchestrationTypes.ts
+++ b/src/types/orchestrationTypes.ts
@@ -52,7 +52,23 @@ export interface CallHttpOptions {
      */
     tokenSource?: TokenSource;
     /**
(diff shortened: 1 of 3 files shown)


src/types/orchestrationTypes.ts
```diff
@@ -52,7 +52,23 @@ export interface CallHttpOptions {
...
+    /**
+     * Specifies whether the Durable extension should continue
+     * polling the request after receiving a 202 response.
+     *
+     * @deprecated this property name is deprecated.
+     * Please use the new `enablePolling` property instead.
+     * If both are specified, the `enablePolling` property takes precedence.
+     *
      * @default true
      */
```
ejizba:
```suggestion
    /**
     * @deprecated use `enablePolling` instead
     */
```

src/types/orchestrationTypes.ts
```diff
@@ -52,7 +52,23 @@ export interface CallHttpOptions {
...
     tokenSource?: TokenSource;
     /**
-     * Specifies whether the DurableHttpRequest should handle the asynchronous pattern.
+     * Specifies whether the Durable extension should continue
+     * polling the request after receiving a 202 response.
+     *
+     * This porperty name replaces the previous `asynchronousPatternEnabled` argument.
+     * If both are specified, this property takes precedence.
+     *
+     * @default true
```
ejizba:
```suggestion
     * Specifies whether to continue polling the request after receiving a 202 response.
     * This replaces `asynchronousPatternEnabled`. If both are specified, `enablePolling` takes precedence.
     * @default true
```

src/types/orchestrationTypes.ts
```diff
@@ -52,7 +52,23 @@ export interface CallHttpOptions {
      */
     tokenSource?: TokenSource;
     /**
-     * Specifies whether the DurableHttpRequest should handle the asynchronous pattern.
+     * Specifies whether the Durable extension should continue
+     * polling the request after receiving a 202 response.
+     *
+     * This porperty name replaces the previous `asynchronousPatternEnabled` argument.
```
ejizba: porperty 👀

ejizba (approved):
Not blocking at this point in time, but let's try to avoid contributing to durable's overly lengthy (IMO) docs. Too much information can be just as bad as not enough

src/types/orchestrationTypes.ts
hossam-nasr: Done

src/types/orchestrationTypes.ts
hossam-nasr: Done

Issue #418, opened 2023-01-09
Can we improve the `callActivity()` and `callSubOrchestrator()` experiences in orchestrations?


hossam-nasr:
(see relevant discussion: https://github.com/ejizba/func-nodejs-prototype/pull/14#discussion_r1051046886)

Currently, if an orchestrator wants to call an activity function, it would have code like this:

```TS
context.df.callActivity('Hello', 'Cairo')
```

passing in the activity's name as a string. This provides no intellisense and no validation to confirm that the activity name is the same as the registered activity function. 

Similarly, suborchestrations are started by specifying the orchestration's name as a string:

```TS
context.df.callSubOrchestrator("DeviceProvisioningOrchestration", deviceId, child_id)
```

This issue is to explore whether the new programming model can provide us an opportunity to improve this `callActivity()` and `callSubOrchestrator()` experience, and whether such an improvement would be worth the engineering cost.

An example of this would be to have the API to register activities (`df.app.activity()`) and orchestrations (`df.app.orchestration()`) return some kind of object that can be passed to the `callActivity()`/`callSubOrchestration()` functions, as such:

```TS
const helloActivity = df.app.activity(activityName, { handler: (input: string) => {
       return `Hello, ${input}`;
   }
});

const orchestrator: OrchestrationHandler = function* (context) {
    const outputs = [];
    outputs.push(yield context.df.callActivity(helloActivity, 'Tokyo'));
    outputs.push(yield context.df.callActivity(helloActivity, 'Seattle'));
    outputs.push(yield context.df.callActivity(helloActivity, 'Cairo'));

    return outputs;
};
df.app.orchestration('durableOrchestrator1', orchestrator);
```

ejizba:
I'm wondering if it makes more sense to flip the order from this:
```typescript
context.df.callActivity(helloActivity, 'Tokyo')
```

To something like this:
```typescript
helloActivity.call(context, 'Tokyo')
```

hossam-nasr:
@ejizba I would be in favor of that solution, and I don't immediately see a reason we couldn't do this

davidmrdavid:
I think that's a really cool idea, @ejizba.

I would take it a step further and ask whether it would be possible to call it like this:

```TypeScript
yield helloActivity.call('Tokyo') // let's not forget the `yield` is necessary.
```

Or taking it to the logical extreme and just do

```Typescript
yield helloActivity('Tokyo') // let's not forget the `yield` is necessary.
```

[...]

PR #461, opened 2023-02-03
Add overload to improve the way to call orchestrations, activities, and entities

### LINKED ISSUE #445: Can we change the type of the argument to `callActivity` and `callSubOrchestrator` into something other than a string?
Same as #418 but narrowed down scope. This issue is particularly concerned with just changing the type of the argument to `context.df.callActivity()` and `context.df.callSubOrchestrator()` into something more reliable than a string. Ideally we should do this in a non-breaking way

hossam-nasr:
Resolves #445 in a non-breaking way:

* Updates `df.app.activity()` to return a `CallActivityInput`.
* Updates `df.app.orchestration()` to return a `CallSubOrchestratorInput`.
* Adds an overload to `context.df.callActivity()` and `context.df.callActivityWithRetry()` to accept a `CallActivitInput` as first argument instead of a `string`, while deprecating the old overload
* Adds an overload to `context.df.callSubOrchestrator()` and `context.df.callSubOrchestratorWithRetry()` to accept a `CallSubOrchestratorInput` as first argument instead of a `string`, while deprecating the old overload
* Adds improvement to entities too, by changing return type of `df.app.entity()` to `RegisterEntityResult`, which includes a `getEntityId()` method that takes a key and returns an `EntityId` instance (can be passed to `context.df.callEntity()` and other entity interaction APIs on clients and orchestration contexts.

**Before:**
Activities/Orchestrations:
```TS
const activityName = 'helloActivity';
const subOrchestratorName = 'sayHelloWithActivity';

const orchestrator: OrchestrationHandler = function* (context: OrchestrationContext) {
    const outputs = [];
    outputs.push(yield context.df.callSubOrchestrator(subOrchestratorName, 'Tokyo'));
    outputs.push(yield context.df.callSubOrchestrator(subOrchestratorName, 'Seattle'));
    outputs.push(yield context.df.callSubOrchestrator(subOrchestratorName, 'Cairo'));

    return outputs;
};
df.app.orchestration('durableOrchestrator1', orchestrator);

const sayHelloWithActivityHandler: OrchestrationHandler = function* (context: OrchestrationContext) {
    return yield context.df.callActivity(activityName, input);
};
 df.app.orchestration(subOrchestratorName, sayHelloWithActivityHandler);

const helloActivityHandler: ActivityHandler = (input: string): string => {
    return `Hello, ${input}`;
};
df.app.activity(activityName, { handler: helloActivityHandler });
```

Entities:
```TS
const entityName = 'counterEntity';

const counterEntityHandler: EntityHandler<number> = (context: EntityContext<number>) => {
    // entity implementation
};
df.app.entity(entityName, counterEntityHandler);

const httpStart: HttpHandler = async (req: HttpRequest, context: InvocationContext): Promise<HttpResponse> => {
    const id: string = req.params.id;
    const entityId: df.EntityId = new df.EntityId(entityName, id);
    // use entityId for management operations
    await client.signalEntity(entityId, 'add', 1);
    const stateResponse = await client.readEntityState(entityId);
    // ...etc.
};
```

**After:**

Orchestrations/Activities:

```TS
const orchestrator: OrchestrationHandler = function* (context: OrchestrationContext) {
    const outputs = [];
    outputs.push(yield context.df.callSubOrchestrator(sayHelloWithActivity, 'Tokyo'));
    outputs.push(yield context.df.callSubOrchestrator(sayHelloWithActivity, 'Seattle'));
    outputs.push(yield context.df.callSubOrchestrator(sayHelloWithActivity, 'Cairo'));

    return outputs;
};
df.app.orchestration('durableOrchestrator1', orchestrator);

const sayHelloWithActivityHandler: OrchestrationHandler = function* (context: OrchestrationContext) {
    return yield context.df.callActivity(helloActivity, input);
};
const sayHelloWithActivity: CallSubOrchestratorInput = df.app.orchestration(
    'sayHelloWithActivity',
    sayHelloWithActivityHandler
);

const helloActivityHandler: ActivityHandler = (input: string): string => {
    return `Hello, ${input}`;
};
const helloActivity: CallActivityInput = df.app.activity('hello', { handler: helloActivityHandler });
```

Entities:

```TS
const counterEntityHandler: EntityHandler<number> = (context: EntityContext<number>) => {
    // entity implementation
};
const counterEntity: RegisterEntityResult = df.app.entity('counterEntity', counterEntityHandler);

const httpStart: HttpHandler = async (req: HttpRequest, context: InvocationContext): Promise<HttpResponse> => {
    const id: string = req.params.id;
    const entityId: df.EntityId = counterEntity.getEntityId(id);
    // use entityId for management operations
    await client.signalEntity(entityId, 'add', 1);
    const stateResponse = await client.readEntityState(entityId);
    // ...etc.
};
```

### DIFF
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -30,13 +30,7 @@ import {
 } from "./task";
 import moment = require("moment");
 import { ReplaySchema } from "./replaySchema";
-import {
-    CallActivityInput,
-    CallHttpOptions,
-    CallSubOrchestratorInput,
-    Task,
-    TimerTask,
-} from "./types";
+import { Activity, CallHttpOptions, Orchestration, Task, TimerTask } from "./types";
 import { SignalEntityAction } from "./actions/signalentityaction";
 
 /**
@@ -224,9 +218,9 @@ export class DurableOrchestrationContext {
      * @returns A Durable Task that completes when the called activity
      * function completes or fails.
      */
-    public callActivity(activity: CallActivityInput, input?: unknown): Task;
+    public callActivity(activity: Activity, input?: unknown): Task;
 
-    public callActivity(nameOrOptions: string | CallActivityInput, input?: unknown): Task {
+    public callActivity(nameOrOptions: string | Activity, input?: unknown): Task {
         const name = typeof nameOrOptions === "string" ? nameOrOptions : nameOrOptions.name;
         const newAction = new CallActivityAction(name, input);
         const task = new AtomicTask(false, newAction);
@@ -256,13 +250,13 @@ export class DurableOrchestrationContext {
      * function.
      */
     public callActivityWithRetry(
-        activity: CallActivityInput,
+        activity: Activity,
         retryOptions: RetryOptions,
         input?: unknown
     ): Task;
 
     public callActivityWithRetry(
-        nameOrOptions: string | CallActivityInput,
+        nameOrOptions: string | Activity,
         retryOptions: RetryOptions,
         input?: unknown
     ): Task {
@@ -326,13 +320,13 @@ export class DurableOrchestrationContext {
      * the format `<calling orchestrator instance ID>:<#>`
      */
     public callSubOrchestrator(
-        orchestration: CallSubOrchestratorInput,
+        orchestration: Orchestration,
         input?: unknown,
         instanceId?: string
     ): Task;
 
     public callSubOrchestrator(
-        nameOrOptions: string | CallSubOrchestratorInput,
+        nameOrOptions: string | Orchestration,
         input?: unknown,
         instanceId?: string
     ): Task {
@@ -378,7 +372,7 @@ export class DurableOrchestrationContext {
      * @param instanceId A unique ID to use for the sub-orchestration instance.
      */
     public callSubOrchestratorWithRetry(
-        orchestration: CallSubOrchestratorInput,
+        orchestration: Orchestration,
         retryOptions: RetryOptions,
         input?: unknown,
         instanceId?: string
@@ -395,7 +389,7 @@ export class DurableOrchestrationContext {
      * @param instanceId A unique ID to use for the sub-orchestration instance.
      */
     public callSubOrchestratorWithRetry(
-        nameOrOptions: string | CallSubOrchestratorInput,
+        nameOrOptions: string | Orchestration,
         retryOptions: RetryOptions,
         input?: unknown,
         instanceId?: string
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -203,9 +203,6 @@ export class DurableOrchestrationContext {
      * function.
      * @returns A Durable Task that completes when the called activity
      * function completes or fails.
-     *
-     * @deprecated this version of `callActivity()` is deprecated. Pass an object
-     * returned from `df.app.activity()` as the first argument instead.
      */
     public callActivity(name: string, input?: unknown): Task;
 
@@ -235,9 +232,6 @@ export class DurableOrchestrationContext {
      * @param retryOptions The retry options for the activity function.
      * @param input The JSON-serializable input to pass to the activity
      * function.
-     *
-     *  @deprecated this version of `callActivity()` is deprecated. Pass an object
-     * returned from `df.app.activity()` as the first argument instead.
      */
     public callActivityWithRetry(name: string, retryOptions: RetryOptions, input?: unknown): Task;
 
@@ -303,9 +297,6 @@ export class DurableOrchestrationContext {
      * @param instanceId A unique ID to use for the sub-orchestration instance.
      * If `instanceId` is not specified, the extension will generate an id in
      * the format `<calling orchestrator instance ID>:<#>`
-     *
-     * @deprecated this version of `callSubOrchestrator()` is deprecated. Pass an object
-     * returned from `df.app.orchestration()` as the first argument instead.
      */
     public callSubOrchestrator(name: string, input?: unknown, instanceId?: string): Task;
 
@@ -351,9 +342,6 @@ export class DurableOrchestrationContext {
      * @param input The JSON-serializable input to pass to the orchestrator
      * function.
      * @param instanceId A unique ID to use for the sub-orchestration instance.
-     *
-     * @deprecated this version of `callSubOrchestratorWithRetry()` is deprecated. Pass an object
-     * returned from `df.app.orchestration()` as the first argument instead.
      */
     public callSubOrchestratorWithRetry(
         name: string,
--- a/src/durableorchestrationcontext.ts
+++ b/src/durableorchestrationcontext.ts
@@ -30,7 +30,13 @@ import {
 } from "./task";
 import moment = require("moment");
 import { ReplaySchema } from "./replaySchema";
-import { CallHttpOptions, Task, TimerTask } from "./types";
+import {
+    CallActivityInput,
+    CallHttpOptions,
+    CallSubOrchestratorInput,
+    Task,
+    TimerTask,
+} from "./types";
 import { SignalEntityAction } from "./actions/signalentityaction";
 
 /**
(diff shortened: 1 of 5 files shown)


src/durableorchestrationcontext.ts
```diff
@@ -202,8 +208,25 @@ export class DurableOrchestrationContext {
...
+    /**
+     * Schedules an activity function for execution
+     *
+     * @param activity The activity object returned from a `df.app.activity()` call.
+     * @param input The JSON-serializable input to pass to the activity
+     * function.
+     * @returns A Durable Task that completes when the called activity
+     * function completes or fails.
+     */
+    public callActivity(activity: CallActivityInput, input?: unknown): Task;
```
davidmrdavid: I find it a bit confusing to have an input of type `CallActivityInput` that isn't the input to the activity :) .
Can we have a different name that differentiates itself from the existing `input` parameter?

How about just exposing a type like `Activity`? In other words: dropping the `-Input` suffix from these types.

src/types/entityTypes.ts
```diff
@@ -20,3 +21,16 @@ export interface EntityContext<T> extends InvocationContext {
...
+ */
+export interface RegisterEntityResult {
+    /**
+     * Returns a new `EntityId` instance with the specified key. Can be passed to entity operation APIs.
+     *
+     * @param key the id to uniquely identity an entity instance among all instances of the same class.
+     * @returns an `EntityId` instance using the registered entity name and the specified key.
+     */
+    getEntityId: (key: string) => EntityId;
+}
```
davidmrdavid: Any reason why this has a uniquely different name?  The other types have a `-Input` suffix.

davidmrdavid (commented):
The PR makes sense to me, but first I have some questions on naming

src/durableorchestrationcontext.ts
```diff
@@ -202,8 +208,25 @@ export class DurableOrchestrationContext {
      * function.
      * @returns A Durable Task that completes when the called activity
      * function completes or fails.
+     *
+     * @deprecated this version of `callActivity()` is deprecated. Pass an object
```
ejizba: Let's be careful about deprecating stuff so quickly. From a migration perspective, I think it would be nice to give users time to stay on old overloads before we deprecate it (if we do). Some people take deprecated notices very seriously and would treat this as an error when upgrading, and we're trying to reduce migration pains

src/durableorchestrationcontext.ts
```diff
@@ -30,7 +30,13 @@ import {
 } from "./task";
 import moment = require("moment");
 import { ReplaySchema } from "./replaySchema";
```
ejizba, line 32: The benefits of this change are not as clear-cut as I initially thought they were. Some feedback on the description:

1. I don't think the "Before" examples are accurate. I think most people (including our templates) would put `'sayHelloWithActivity'` in a `const` which is much safer.
1. It bothers me in the "After" example that `sayHelloWithActivity` is used before it's defined. I know technically it will work as-is, but some linters will flag that as a problem and from a readability perspective I agree with the linters. With a string the order doesn't matter, but with the object it does.
1. I would like us to explore more complex projects where these aren't all in the same file. My gut is telling me that the imports could get complicated now that we depend on the output of registering the handler instead of using a static string. For example - have you ever run into circular import errors? Not sure if they're possible in durable, but they're a huge pain haha

davidmrdavid: +1 to (3). Can we do a manual check for how this works across multiple files?

As for whether circular import errors are possible in DF: You definitely could force one. Have 2 Orchestrators in different files and have them spawn each other as sub-orchestrators. I would be curious to see how the runtime handles this.

src/durableorchestrationcontext.ts
hossam-nasr: Like call it just `Activity` and `Orchestration`?

src/types/entityTypes.ts
hossam-nasr: Because this one isn't passed to one unique API, but can be used to create an `EntityId` that itself is used in multiple APIs (`signalEntity()`, `callEntity()`, `getEntityState()`, etc.). I figured it's easier to characterize this not by what's it's an input _to_, but rather what it's an output _of_. Admittedly, I didn't have particularly good names for these and I'm open for suggestions :)

src/durableorchestrationcontext.ts
davidmrdavid: Yep, same for `Entity` I'd say

src/types/entityTypes.ts
davidmrdavid: I see what you mean and realize that this is used differently. But it's also this type is "the same" as the other ones in that it is an output of the Function-registration APIs. So I'd recommend making their names uniform. I'd suggest just calling them `Activity`, `Orchestrator`, and `Entity` if possible.

src/durableorchestrationcontext.ts
hossam-nasr: I don't love that naming tbh, because it feels inaccurate, and also those are very special words in the SDK, we might need them later in more appropriate contexts. Besides, for `Entity`, it conflicts with the name of the class we already have. But since I have no better alternatives, I've made that change.

src/types/entityTypes.ts
hossam-nasr: Done

src/durableorchestrationcontext.ts
hossam-nasr: Ok, I've removed the deprecation flags

src/durableorchestrationcontext.ts
hossam-nasr: Re: 1. Fair enough, I've updated the description.
Re: 2. We could switch the order in the templates, but I agree it is inconvenient.
Re: 3. I've tested this with the following project, and it worked fine:

```TS
// helloActivity.ts:
import * as df from 'durable-functions';
import { ActivityHandler } from 'durable-functions';

const helloActivityHandler: ActivityHandler = (input: string): string => {
    return `Hello, ${input}`;
};

export default df.app.activity('helloActivity', { handler: helloActivityHandler });

//

// sayHelloWithActivity.ts:
import * as df from 'durable-functions';
import { OrchestrationContext, OrchestrationHandler } from 'durable-functions';
import helloActivity from './helloActivity';

const sayHelloWithActivityHandler: OrchestrationHandler = function* (context: OrchestrationContext) {
    return yield context.df.callActivity(helloActivity, input);
};

export default df.app.orchestration('sayHelloWithActivity', sayHelloWithActivityHandler);

// 

// helloSequence.ts:

import * as df from 'durable-functions';
import { OrchestrationContext, OrchestrationHandler } from 'durable-functions';
import sayHelloWithActivity from './sayHelloWithActivity';

const orchestrator: OrchestrationHandler = function* (context: OrchestrationContext) {
    const outputs = [];
    outputs.push(yield context.df.callSubOrchestrator(sayHelloWithActivity, 'Tokyo'));
    outputs.push(yield context.df.callSubOrchestrator(sayHelloWithActivity, 'Seattle'));
    outputs.push(yield context.df.callSubOrchestrator(sayHelloWithActivity, 'Cairo'));

    return outputs;
};
df.app.orchestration('helloSequence', orchestrator);
```

I'd argue I think it's cleaner to use this style when you have multiple orchestrations/activities across multiple files, and might be cleaner to use constant strings at the top if everything is in one file.

src/durableorchestrationcontext.ts
ejizba: > those are very special words in the SDK, we might need them later in more appropriate contexts

This concerns me as well. What about `RegisteredActivity` or `RegisteredActivityFunction`?

hossam-nasr: I like that, thanks Eric! Updated to `RegisteredActivity`, `RegisteredOrchestration` and `RegisteredEntity`.

src/durableorchestrationcontext.ts
ejizba: I feel these patterns are pretty close in functionality, and largely come down to personal preference. In that case, is the downside of overloads really worth the new option? If the new one was clearly better, I'd be fine with the overloads because the downside would be temporary. As it is, I feel like we'll be stuck supporting both overloads indefinitely because there will never be a compelling enough reason to remove the old pattern.

To remind y'all, this is why I try to avoid overloads - it turns a one line error into an ugly five line error:

|Without overloads|With overloads|
|---|---|
|Screenshot 2023-02-08 at 3 51 22 PM|Screenshot 2023-02-08 at 3 48 18 PM|

@davidmrdavid @hossam-nasr Are you both confident you like the new pattern better? If yes, I will defer to you guys and I'm fine getting it in. If no, I will remind you that we have plenty of time to tinker with this design since we're doing it in a non-breaking way. We can easily ship this feature any time in the future regardless of announcement dates

davidmrdavid: @ejizba: Thanks for the concrete example. I'm confident that, across all DF SDKs, we're looking to support safer ways of calling Activities/Sub-Orchestrators/Entities by providing alternatives to this string-based parameter. I know I've been asked to do something equivalent to this in Python's new progmodel, although overloads there a bit a nicer in error cases.

Yes, we have time to tinker with this design and it appears we would be ok to merge this at a later date, so I'm no rush. I think this would be something I'd love to demo, among other bits, to the DF team. That would help us be in the same page about the tradeoffs of the feature.

hossam-nasr: I don't have strong feelings about either merging this now (and including it in preview), or punting until later, after/during preview. I will defer to @davidmrdavid on this one. Let me know if you think we should have this PR in before preview and I'll happily rebase the branch so we can merge.

hossam-nasr:
Looks like we've decided to table this for preview, and we'll revisit this later. This is non-breaking so we can always add it after preview before GA, at GA, or even after GA. We already have two issues tracking this: #445 and #418, so I'm closing this PR

PR #462, opened 2023-02-03
Remove IHttpRequest and IHttpResponse interfaces

### LINKED ISSUE #454: Leverage the HttpRequest and HttpResponse classes from `@azure/functions`
Let's get rid of the `IHttpRequest` and `IHttpRespones` classes in this SDK, and instead rely on the `HttpRequest` and `HttpResponse` classes provided by `@azure/functions`. 

This means that `client.createCheckStatusResponse()` and `client.waitForCompletionOrCreateCheckStatusResponse()` should both accept an `HttpRequest` object, and return an `HttpResponse` object. 

While technically a breaking change, this is actually helpful in most cases, because `HttpRequest` is the type of trigger input received by HTTP-triggered functions anyway, and most functions just pass the result of those API back to the Functions library, which also now expects an `HttpResponse` object

hossam-nasr:
Resolves #454. This PR removes the `IHttpResponse` and `IHttpRequest` object, and instead relies on the `HttpResponse` and `HttpRequest` objects from `@azure/functions` (respectively). Since this affects the type signatures of public APIs, this _is_ a breaking change, but we discussed offline, and we decided this is a breaking change worth pursuing. In most cases, this should actually _help_ customers, since the expected `HttpRequest` argument to APIs is usually the request object given to the client functions (which now has `HttpRequest` type), and the `HttpResponse` type is usually passed as the return value of the function, which now is expected to be `HttpResponse`. I'm not sure why the `IHttpResponse` and `IHttpRequest` types were honored/respected before, but I've removed the checks/tests related to that as well.

### DIFF
--- a/test/unit/orchestrationclient-spec.ts
+++ b/test/unit/orchestrationclient-spec.ts
@@ -90,32 +90,6 @@ describe("Orchestration Client", () => {
             expect(response.headers.get("Retry-After")).to.equal("10");
         });
 
-        it(`returns a proper response object from request.http.url`, async () => {
-            const client = new DurableOrchestrationClient(defaultClientInputData);
-            const requestObj = {
-                http: {
-                    url: defaultRequestUrl,
-                    method: "GET",
-                },
-            };
-
-            const response = client.createCheckStatusResponse(requestObj, defaultInstanceId);
-            const responseBody = await response.json();
-
-            const expectedPayload = TestUtils.createHttpManagementPayload(
-                defaultInstanceId,
-                Constants.DefaultLocalOrigin,
-                defaultTaskHub,
-                defaultConnection
-            );
-
-            expect(response.status).to.equal(202);
-            expect(responseBody).to.deep.equal(expectedPayload);
-            expect(response.headers.get("Content-Type")).to.equal("application/json");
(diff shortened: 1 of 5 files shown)


test/unit/orchestrationclient-spec.ts
```diff
@@ -90,32 +90,6 @@ describe("Orchestration Client", () => {
...
-                defaultConnection
-            );
-
-            expect(response.status).to.equal(202);
-            expect(responseBody).to.deep.equal(expectedPayload);
-            expect(response.headers.get("Content-Type")).to.equal("application/json");
-            expect(response.headers.get("Location")).to.equal(expectedPayload.statusQueryGetUri);
-            expect(response.headers.get("Retry-After")).to.equal("10");
-        });
-
```
davidmrdavid, line 118: It seems the other tests got adapted to use the new type, but this one simply got removed. Why is that?

davidmrdavid (commented):
Just a question at this time.

test/unit/orchestrationclient-spec.ts
hossam-nasr: Because this test explicitly tests `returns a response object from request.http.url`, i.e., from the no-longer-supported `IHttpRequest` type.

davidmrdavid (approved):

ejizba (approved):

PR #463, opened 2023-02-03
Don't allow compound tasks with no children

### LINKED ISSUE #424: Should we continue to support compound tasks with no children?
Currently, the SDK supports creating compoung `Task`s (e.g.., `WhenAny`/`WhenAll` tasks) without any children. This support can lead to weird scenarios such as #423. Perhaps we should consider not supporting this scenario anymore in v3 (See discussion: https://github.com/Azure/azure-functions-durable-js/issues/423#issuecomment-1397694184)

hossam-nasr:
Resolves #424

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -166,9 +166,12 @@ export abstract class CompoundTask extends DFTask {
         children.map((c) => (c.parent = this));
         this.firstError = undefined;
 
-        // If the task has no children, then it's completed by definition.
+        // If the task has no children, throw an error
         if (children.length == 0) {
-            this.state = TaskState.Completed;
+            const message =
+                "Cannot initialize a CompoundTask with an empty childrens array. " +
+                "When constructing a CompoundTask (such as Task.all() or Task.any()), you must speciify at leat one Task.";
+            throw new Error(message);
         }
     }
 
--- a/src/task.ts
+++ b/src/task.ts
@@ -170,8 +170,7 @@ export abstract class CompoundTask extends DFTask {
         // See issue here for why this isn't allowed: https://github.com/Azure/azure-functions-durable-js/issues/424
         if (children.length == 0) {
             const message =
-                "Cannot initialize a CompoundTask with an empty childrens array. " +
-                "When constructing a CompoundTask (such as Task.all() or Task.any()), you must speciify at leat one Task.";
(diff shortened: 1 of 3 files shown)


src/task.ts
```diff
@@ -166,9 +166,12 @@ export abstract class CompoundTask extends DFTask {
         children.map((c) => (c.parent = this));
         this.firstError = undefined;
 
-        // If the task has no children, then it's completed by definition.
+        // If the task has no children, throw an error
```
ejizba, line 169: Can you link to this issue in the comment in the code:
`// https://github.com/Azure/azure-functions-durable-js/issues/424`

Otherwise it doesn't feel entirely clear to me _why_ this isn't allowed

src/task.ts
```diff
@@ -166,9 +166,12 @@ export abstract class CompoundTask extends DFTask {
...
         if (children.length == 0) {
-            this.state = TaskState.Completed;
+            const message =
+                "Cannot initialize a CompoundTask with an empty childrens array. " +
+                "When constructing a CompoundTask (such as Task.all() or Task.any()), you must speciify at leat one Task.";
```
ejizba:
```suggestion
                "When constructing a CompoundTask (such as Task.all() or Task.any()), you must specify at least one Task.";
```

src/task.ts
```diff
@@ -166,9 +166,12 @@ export abstract class CompoundTask extends DFTask {
         children.map((c) => (c.parent = this));
         this.firstError = undefined;
 
-        // If the task has no children, then it's completed by definition.
+        // If the task has no children, throw an error
         if (children.length == 0) {
-            this.state = TaskState.Completed;
+            const message =
+                "Cannot initialize a CompoundTask with an empty childrens array. " +
```
ejizba: Can we delete this line? I feel like both lines say the same thing and I like the bottom one better

src/task.ts
hossam-nasr: Done

src/task.ts
hossam-nasr: Oops, thanks for catching this!

src/task.ts
hossam-nasr: Done

ejizba (approved):

PR #465, opened 2023-02-03
Add new runnable TS Samples Directory

### LINKED ISSUE #421: Add v3 samples/templates

hossam-nasr:
TS version of #438. Resolves #421.


samples-ts/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "main": "dist/functions/*.js",
+    "scripts": {
+        "build": "tsc",
+        "watch": "tsc -w",
+        "prestart": "npm run build",
+        "start": "func start",
+        "test": "echo \"No tests yet...\""
+    },
+    "devDependencies": {
+        "@azure/functions": "file:../../../nodejs-worker/library",
```
ejizba: This should reference an actual version. And should probably be in dependencies (I know it's technically already a sub dependency of durable-functions, but still)

ejizba (approved):
Similar comment as other samples PR. Overall structure looks good, but didn't review every single line of code

samples-ts/package.json
hossam-nasr: Good call out, thanks! Done

ejizba: @hossam-nasr you fixed the version, but it should still be moved from `devDependencies` to `dependencies`

samples-ts/extensions.csproj
```diff
@@ -0,0 +1,19 @@
...
+  <WarningsAsErrors></WarningsAsErrors>
+  <DefaultItemExcludes>**</DefaultItemExcludes>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="host.json" />
+    <None Include="package.json" />
+    <None Include="local.settings.json" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.9.0" />
```
davidmrdavid: nitpick: why don't you make the DF version be `2.*` so that it always downloads the latest released version of DF Extension v2?

davidmrdavid (approved):
LGTM but left a comment about your `.csproj`.

samples-ts/package.json
hossam-nasr: Done, thanks

samples-ts/extensions.csproj
hossam-nasr: Good call out! Did that for the JS Samples directory too

PR #469, opened 2023-02-08
Switch from moment to luxon

### LINKED ISSUE #276: Replace `moment.js` usage with `Luxon`
The `moment.js` library is in [maintenance mode](https://momentjs.com/docs/#/-project-status/) and the maintainers suggest switching to a different library. We should update our samples to use [Luxon](https://moment.github.io/luxon/index.html) instead.

hossam-nasr:
As in title, resolves #276

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
         "axios": "^0.21.1",
         "debug": "~2.6.9",
         "lodash": "^4.17.15",
-        "moment": "^2.29.2",
+        "luxon": "^3.2.1",
         "uuid": "~3.3.2",
         "validator": "~13.7.0"
     },
@@ -56,6 +56,7 @@
         "@types/chai-string": "~1.4.1",
         "@types/debug": "0.0.29",
         "@types/lodash": "^4.14.119",
+        "@types/luxon": "^3.2.0",
         "@types/mocha": "^7.0.2",
         "@types/nock": "^9.3.0",
         "@types/node": "^18.0.0",
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,3 +1,5 @@
+import { Duration } from "luxon";
+
 /** @hidden */
 export class Utils {
     public static processInput<T>(input: string | T): string | T {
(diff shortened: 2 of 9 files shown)


src/utils.ts
```diff
@@ -118,4 +120,49 @@ export class Utils {
...
+
+    /**
+     * @hidden
+     * Converts a string in the format `[-][d.]hh:mm:ss[.fffffff]`
+     * to a `Duration` object
+     * (This is the format the default .NET serializer uses, and the
+     * format coming from the Durable extension)
+     * @param input the formatted string
+     */
+    public static durationFromString(input: string): Duration {
```
hossam-nasr, line 132: This was necessary because .NET serializes `TimeSpan` objects in this weird, non-ISO standard format: `[-][d.]hh:mm:ss[.fffffff]`. See [here](https://learn.microsoft.com/en-us/dotnet/api/system.timespan.tostring?view=net-7.0) for more details. This is the format of the durations the SDK receives in its payload from the Extension. `moment.js` seemed to handle this fine out of the gate, but I couldn't find a `luxon` method that either accepts this format, or even allows specifying a custom format. The same is also true for `dayjs`, the other contender for the replacement. I know it's not ideal, but I also added unit tests to make sure this logic is sound, and step-by-step comments. If anyone has better suggestions, I'm happy to hear them.  There is a GH issue here: https://github.com/moment/luxon/issues/296 on luxon to add a `Duration.fromFormat()` method, which would help us in this case, but until that's closed, this is the best I could come up with :)

test/unit/utils-spec.ts
```diff
@@ -154,6 +154,63 @@ describe("Utils", () => {
             }).to.not.throw();
         });
     });
+
+    describe("durationFromString()", () => {
```
hossam-nasr, line 158: These unit tests were mostly derived from the examples on the .NET serialization docs here: https://learn.microsoft.com/en-us/dotnet/api/system.timespan.tostring?view=net-7.0#system-timespan-tostring

package.json
```diff
@@ -46,7 +46,7 @@
         "axios": "^0.21.1",
         "debug": "~2.6.9",
         "lodash": "^4.17.15",
-        "moment": "^2.29.2",
+        "luxon": "^3.2.1",
```
ejizba, line 49: Unless there's a compelling reason to use Luxon, I think we should use dayjs. It's way more popular than Luxon and it specifically calls itself an "alternative to Moment.js"

https://npmtrends.com/dayjs-vs-luxon

src/utils.ts
ejizba: Can we punt this PR until after preview? I believe we discussed this during triage and determined the durable package doesn't export any moment types in its public API (right?), so it's more of an internal engineering change as opposed to something that will directly benefit the user.

Even after preview, I think the fact that we have to manually implement `durationFromString` is reason enough to stay on moment. It's not like the package is dead - it's still in maintenance mode.

package.json
hossam-nasr: I'm not sure why `Luxon` was picked as the alternative instead of `dayjs`. @davidmrdavid thoughts?

I will say, `Luxon` was also created by the creators of `moment.js`, and using `Luxon` for this migration, I've liked how more strict and type-safe it feels compared to `moment.js`, something that `dayjs` doesn't do from what I've seen.

src/utils.ts
hossam-nasr: The Durable SDK does list `moment` as one of its dependencies, and it _does_ export `moment` types in its public types, as the type of some private fields, which are included in the auto-generated `.d.ts` files by default.

package.json
davidmrdavid: I think `Luxon` was suggested by a third party user who notified us of `moment.js` now longer actively developed. I have no strong feelings about the use of library so whatever works for you both, works for me.

src/utils.ts
davidmrdavid: @hossam-nasr: just to make sure I understand the context. If we're moving to manually generated types, would we be able to avoid exporting `moment`?

hossam-nasr: @davidmrdavid Yes, I believe so. But it will stay in our dependencies, even if it's not in our public types (and therefore changing it in the future won't be a breaking change). We can def still do this after preview, but I thought we agreed during triage to just do it now if the fix is easy enough.

ejizba: > it's not in our public types (and therefore changing it in the future won't be a breaking change)

Given this, I think we should stay on moment until Luxon or dayjs add support for the format we need

ejizba (changes requested):

hossam-nasr:
> Given this, I think we should stay on moment until Luxon or dayjs add support for the format we need

Closing this PR for now

Issue #458, opened 2023-02-03
Switch to manual generation of DTS files


hossam-nasr:
Related to #352 and #416. Currently, the Durable SDK auto-generates its `.d.ts` from the source-code `.ts` files. We should switch to manually authoring our type declarations and our `.d.ts` files. This has multiple benefits:

* We can be intentional about which types we export, and what public APIs we support, helping us become more thoughtful about any breaking changes
* Helps us limit the number of unnecessary exports (see #352), avoiding issues such as #353.

hossam-nasr:
Discussed some offline. As part of this, we should also try to switch from autogeneration of `.d.ts` files, to manual authoring of them. This should would help us become more intentional about which types we export, and also address #352. To do this, we should:

1. Move all types/interfaces that we should expose publicly to the `types` directory (in normal `.ts` files)
2. Separate class implementations from declarations, and move declarations to the `types` directory (also in `.ts` files)
3. Switch to manual `.d.ts` file generation

Work for points 2 and 3 is already being done

_Originally posted by @hossam-nasr in https://github.com/Azure/azure-functions-durable-js/issues/416#issuecomment-1414169027_

hossam-nasr:
Closed by #473

PR #470, opened 2023-02-08
Export class declarations from the `types` directory

hossam-nasr:
Prelude to #458, part of #416. This PR adds class declarations exported from the root of the package (living in the `types` directory), for relevant public-facing classes. Specifically, it exports declarations for the following classes:

* `PurgeHistoryResult`
* `EntityStateResponse`
* `HttpManagementPayload`
* `EntityId`
* `DurableOrchestrationStatus`
* `RetryOptions`
* `ManagedIdentityTokenSource`
* `AggregatedError`
* `DurableError`
* `DummyOrchestrationContext`
* `DummyEntityContext`

Furthermore, it also moves the `OrchestrationRuntimeStatus` enum to the `orchestrationTypes` file.


src/aggregatederror.ts
```diff
@@ -1,22 +1,9 @@
...
- * Name: DurableError
- * Message: The activity function "ActivityA" failed.
- * StackTrace: <stacktrace>
- * -----------------------------------
- * Name: DurableError
- * Message: The activity function "ActivityB" failed.
- * StackTrace: <stacktrace>
- */
-export class AggregatedError extends Error {
+export class AggregatedError extends Error implements types.AggregatedError {
```
davidmrdavid, line 6: I notice this pattern of having a class "X" implement "types.X". Is this a common pattern in Node? I find it a bit confusing to have two components of the same name

davidmrdavid (commented):
PR mostly makes sense. Just a question at this time.

src/aggregatederror.ts
hossam-nasr: Yup, this is a common pattern. This is the distinction between class declaration and implementation. Since we want to move towards manual authoring of `d.ts` files, I'm manually writing out the declaration for the classes we would export here. This is taking us one step in the direction of #458. In reality, these two components already exist, except the declaration piece is currently automatically generated.

hossam-nasr: Also, you can see this pattern too in `@azure/functions` with classes like `HttpRequest` and `HttpResponse` that it exports.

src/types/index.ts
```diff
@@ -1,10 +1,88 @@
...
+ * A specfic error thrown when a scheduled activity or suborchestrator has failed.
+ * This error can be checked for via `instanceof` guards to catch only exceptions thrown
+ * by the DurableJS library.
+ */
+export declare class DurableError extends Error {
+    /**
+     * Constructs a `DurableError` instance.
+     * @param message error message.
+     */
+    constructor(message?: string);
```
ejizba: `message?: string` is slightly different than `message: string | undefined`, which is how the class is defined in `src`. I'd prefer to keep them consistent

src/types/orchestrationTypes.ts
```diff
@@ -65,3 +85,112 @@ export interface CallHttpOptions {
...
+     * default. It is populated only when explicitly requested in the call
+     * to `DurableOrchestrationClient.getStatus()`.
+     */
+    readonly history?: Array<unknown>;
+}
+
+/**
+ * The status of an orchestration instance.
+ */
+export enum OrchestrationRuntimeStatus {
```
ejizba: enums are more like classes than interfaces. enums & classes stick around when compiled to javascript, whereas types & interfaces are thrown away. That means you still need the real code in the src folder, and this file can only do `export declare enum`. It works now because this file is a `.ts` file, but it should break once this file is converted to a `.d.ts` file

In addition, all imports of `OrchestrationRuntimeStatus` will need to reference the src file, not the types file.

src/types/entityTypes.ts
```diff
@@ -20,3 +20,70 @@ export interface EntityContext<T> extends InvocationContext {
...
     df: DurableEntityContext<T>;
 }
+
+/**
+ * An entity context with dummy default values to facilitate mocking/stubbing the
+ * Durable Functions API.
+ */
+export declare class DummyEntityContext<T> extends InvocationContext implements EntityContext<T> {
+    /**
+     * Creates a new instance of a dummy orchestration context.
```
ejizba: Looks like you copied the `orchestration` comments, not the `entity` comments

src/types/durableClientTypes.ts
```diff
@@ -20,3 +20,52 @@ export interface StartNewOptions {
...
+     * The HTTP POST instance rewind endpoint URL.
+     */
+    readonly rewindPostUri: string;
+    /**
+     * The HTTP DELETE purge endpoint URL.
+     */
+    readonly purgeHistoryDeleteUri: string;
+
+    /** @hidden */
+    [key: string]: string;
```
ejizba: I'd prefer to remove hidden types

src/types/index.ts
```diff
@@ -1,10 +1,88 @@
...
+export declare class RetryOptions {
+    /**
+     * Creates a new instance of RetryOptions with the supplied first retry and
+     * max attempts.
+     * @param firstRetryIntervalInMilliseconds The first retry interval (ms). Must be greater than 0.
+     * @param maxNumberOfAttempts The maximum number of attempts.
+     */
+    constructor(firstRetryIntervalInMilliseconds: number, maxNumberOfAttempts: number);
+    /**
+     * The retry backoff coefficinet
```
ejizba:
```suggestion
     * The retry backoff coefficient
```

src/types/index.ts
```diff
@@ -1,10 +1,88 @@
...
+    constructor(resource: string);
+
+    /**
+     * The Azure Active Directory resource identifier of the web API being invoked.
+     * For example, `https://management.core.windows.net/` or `https://graph.microsoft.com/`.
+     */
+    readonly resource: string;
+
+    /** @hidden */
+    kind: string;
```
ejizba: hidden

src/types/index.ts
hossam-nasr: How is it different?

src/types/orchestrationTypes.ts
hossam-nasr: Done. Is there an equivalent of the `implements` here, or how is the `.d.ts` declaration mapped to the src file?

src/types/entityTypes.ts
hossam-nasr: Oops thanks, fixed!

src/types/index.ts
hossam-nasr: Removed

src/types/index.ts
hossam-nasr: Fixed

src/types/durableClientTypes.ts
hossam-nasr: Removed

src/types/index.ts
ejizba: With `message?: string`, its easier to forget to pass `message`. So for example if you want people to pass a message 99% of the time, you'd use `message: string | undefined` to remind people. If `message` is unimportant and used less often, you'd use `message?: string` for the sake of convenience.

I don't really care which one we choose in this case, but I think the types and the src file should be consistent

src/types/orchestrationTypes.ts
ejizba: I don't think there's an `implements`, no. Tbh I don't use enums very often as they can be kinda clunky

ejizba (approved):
Just the nit about the `message?` thing left, but otherwise LGTM

src/types/index.ts
hossam-nasr: Ok fair enough. Since it was `| undefined` in the `src`, I've changed it to be the same in the `types` too.

PR #473, opened 2023-02-09
Manually export .d.ts files

### LINKED ISSUE #458: Switch to manual generation of DTS files

hossam-nasr:
Resolves #458. To make reviewing easier, I've split this PR into smaller commits with descriptive names. The first commit does nothing but move files from the `src/types` to the `types` directory and updates imports.


types/orchestrationTypes.d.ts
```diff
@@ -1,6 +1,6 @@
 import { FunctionOptions, FunctionTrigger, InvocationContext, LogHandler } from "@azure/functions";
-import { Task, TimerTask, TokenSource } from ".";
-import { EntityId, RetryOptions } from "../classes";
+import { RetryOptions, Task, TimerTask, TokenSource } from ".";
+import { EntityId } from "./entityTypes";
```
ejizba: While we're renaming, I'd prefer if we get rid of `Types` from these file names. Now that its in the `types/` folder and has the `.d.ts` extension, I think `Types` is implied

ejizba (approved):
I like the smaller commits, thanks!

I'm still a bit nervous about the scope of this change, but I think it's worth the risk. I'm glad it's being done now instead of closer to the preview announcement, and I hope we do plenty of manual testing once this is merged. Tbh I expect some bumps, but I think we should be able to handle them

davidmrdavid (approved):
LGTM. Thanks for the well-split commits.
A follow-up action item would be to ensure that our autogenerated docs [here](https://learn.microsoft.com/en-us/javascript/api/durable-functions/?view=azure-node-latest) continue working after all these changes. Definitely not a preview blocker, but I'd like to finish this validation before GA.

types/orchestrationTypes.d.ts
hossam-nasr: Makes sense, done.

PR #476, opened 2023-02-10
Bump framework package

ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-library/issues/53

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "3.0.0-alpha.2",
             "license": "MIT",
             "dependencies": {
-                "@azure/functions": "^4.0.0-alpha.7",
+                "@azure/functions": "^4.0.0-alpha.8",
                 "axios": "^0.21.1",
                 "debug": "~2.6.9",
                 "lodash": "^4.17.15",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,4855 +1,4855 @@
...
-            "version": "3.1.1",
-            "dev": true
-        },
-        "yocto-queue": {
-            "version": "0.1.0",
-            "dev": true
-        }
-    }
-}
+{
```
hossam-nasr, line 1: Why is there such a big diff in the lock file?

hossam-nasr (approved):
LGTM

davidmrdavid (approved):

package-lock.json
ejizba: Well it's only whitespace changes. I've had these changes show up on my local branch recently every time I run `npm install` but didn't really think about it. Can one of you pull my branch, run `npm install` and see if your npm changes it back? I'm curious if we have competing `npm` formatting going on here. I'm on npm v8.10.0 btw

ejizba: tl;dr

If Hossam runs `npm install` from my branch the line endings stay as-is, so we decided I'm okay to merge this.

---

More details:

Tested with Hossam offline. It turns out my `npm` is converting `LF` line endings to `CRLF` line endings to match what's in the `package.json` file. The entire rest of the durable repo is using `CRLF` (probably because of [this](https://github.com/Azure/azure-functions-durable-js/blob/v2.x/.editorconfig#L7)), so in a way my `npm` is just making `package-lock.json` consistent with the rest of the repo. The other js repos are using `LF` as far as I checked, but I didn't feel like trying to change durable

PR #482, opened 2023-02-22
Switch to using an options object for all applicable APIs

### LINKED ISSUE #441: Switch APIs from using discrete optional arguments to using an options argument where it makes sense

hossam-nasr:
Resolves #441. Switch from using concrete optional arguments to an options object where applicable.

I've broken this PR into small steps in commits to aid with PR review.

APIs changed in this PR (notice this list is slightly different than the one in the issue):

**In DurableOrchestrationClient:**

<table>
<tr>
<th> Before </th>
<th> After </th>
</tr>
<tr>
<td>

```TS
getStatus(
    instanceId: string,
    showHistory?: boolean,
    showHistoryOutput?: boolean,
    showInput?: boolean
): Promise<DurableOrchestrationStatus>
```
</td>
<td>

```TS
getStatus(
    instanceId: string, 
    options?: GetStatusOptions
): Promise<DurableOrchestrationStatus>
```
</td>
</tr>
<tr>
<td>

```TS
getStatusBy(
    createdTimeFrom: Date | undefined,
    createdTimeTo: Date | undefined,
    runtimeStatus: OrchestrationRuntimeStatus[]
): Promise<DurableOrchestrationStatus[]>
```

</td>
<td>

```TS
getStatusBy(
    options: SelectionOptions
): Promise<DurableOrchestrationStatus[]>
```

</td>
</tr>
<tr>
<td>

```TS
purgeInstanceHistoryBy(
    createdTimeFrom: Date,
    createdTimeTo?: Date,
    runtimeStatus?: OrchestrationRuntimeStatus[]
): Promise<PurgeHistoryResult>
```

</td>
<td>

```TS
purgeInstanceHistoryBy(
    options: SelectionOptions
): Promise<PurgeHistoryResult>
```

</td>
</tr>
<tr>
<td>

```TS
raiseEvent(
    instanceId: string,
    eventName: string,
    eventData: unknown,
    taskHubName?: string,
    connectionName?: string
): Promise<void>
```

</td>
<td>

```TS
raiseEvent(options: RaiseEventOptions): Promise<void>
```

</td>
</tr>
<tr>
<td>

```TS
readEntityState<T>(
    entityId: EntityId,
    taskHubName?: string,
    connectionName?: string
): Promise<EntityStateResponse<T>>
```

</td>
<td>

```TS
readEntityState<T>(
    entityId: EntityId,
    options?: TaskHubOptions
): Promise<EntityStateResponse<T>>
```

</td>
</tr>
<tr>
<td>

```TS
rewind(
    instanceId: string,
    reason: string,
    taskHubName?: string,
    connectionName?: string
): Promise<void>`
```

</td>
<td>

```TS
rewind(
    instanceId: string, 
    reason: string, 
    options?: TaskHubOptions
): Promise<void>
```

</td>
</tr>
<tr>
<td>

```TS
signalEntity(
    entityId: EntityId,
    operationName?: string,
    operationContent?: unknown,
    taskHubName?: string,
    connectionName?: string
): Promise<void>
```
</td>
<td>

```TS
signalEntity(
    entityId: EntityId, 
    options?: SignalEntityContext
): Promise<void>
```
</td>
</tr>
<tr>
<td>

```TS
waitForCompletionOrCreateCheckStatusResponse(
    request: HttpRequest,
    instanceId: string,
    timeoutInMilliseconds?: number,
    retryIntervalInMilliseconds?: number
): Promise<HttpResponse>;
```

</td>
<td>

```TS
waitForCompletionOrCreateCheckStatusResponse(
    request: HttpRequest,
    instanceId: string,
    waitOptions?: WaitForCompletionOptions
): Promise<HttpResponse>;
```

</td>
</tr>
</table>

**In DurableOrchestrationContext:**

<table>
<tr>
<th> Before </th>
<th> After </th>
</tr>

<tr>
<td>

```TS
callSubOrchestrator(
    name: string,
    input?: unknown, 
    instanceId?: string
): Task
```

</td>
<td>

```TS
// This is effectively equivalent
// to the startNew() change
callSubOrchestrator(
    name: string, 
    options?: CallSubOrchestratorOptions
): Task
```

</td>
</tr>
<tr>
<td>

```TS
callSubOrchestratorWithRetry(
    name: string,
    retryOptions: RetryOptions,
    input?: unknown,
    instanceId?: string
): Task
```

</td>
<td>

```TS
callSubOrchestratorWithRetry(
    name: string,
    retryOptions: RetryOptions,
    options?: CallSubOrchestratorOptions
): Task
```

</td>
</tr>
</table>

### DIFF
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -222,11 +222,7 @@ export class DurableOrchestrationClient implements DurableClient {
     public async getStatusBy(
         selectionOptions: SelectionOptions
     ): Promise<DurableOrchestrationStatus[]> {
-        const internalOptions: GetStatusInternalOptions = {
-            ...selectionOptions,
-        };
-
-        const response = await this.getStatusInternal(internalOptions);
+        const response = await this.getStatusInternal(selectionOptions);
         switch (response.status) {
             case 200:
                 return response.data as DurableOrchestrationStatus[];
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -217,10 +217,8 @@ export class DurableOrchestrationClient implements DurableClient {
         }
     }
 
-    public async getStatusBy(
-        filterOptions: OrchestrationFilter
-    ): Promise<DurableOrchestrationStatus[]> {
-        const response = await this.getStatusInternal(filterOptions);
+    public async getStatusBy(filter: OrchestrationFilter): Promise<DurableOrchestrationStatus[]> {
+        const response = await this.getStatusInternal(filter);
         switch (response.status) {
             case 200:
                 return response.data as DurableOrchestrationStatus[];
@@ -252,11 +250,9 @@ export class DurableOrchestrationClient implements DurableClient {
         }
     }
 
-    public async purgeInstanceHistoryBy(
-        filterOptions: OrchestrationFilter
-    ): Promise<PurgeHistoryResult> {
+    public async purgeInstanceHistoryBy(filter: OrchestrationFilter): Promise<PurgeHistoryResult> {
         let requestUrl: string;
-        const { createdTimeFrom, createdTimeTo, runtimeStatus } = filterOptions;
+        const { createdTimeFrom, createdTimeTo, runtimeStatus } = filter;
         if (this.clientData.rpcBaseUrl) {
             // Fast local RPC path
             let path = new URL("instances/", this.clientData.rpcBaseUrl).href;
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -27,7 +27,6 @@ import {
     DurableClient,
     GetStatusOptions,
     SelectionOptions,
-    RaiseEventOptions,
     TaskHubOptions,
     SignalEntityOptions,
     WaitForCompletionOptions,
@@ -321,8 +320,13 @@ export class DurableOrchestrationClient implements DurableClient {
         }
     }
 
-    public async raiseEvent(options: RaiseEventOptions): Promise<void> {
-        const { eventName, instanceId, eventData, taskHubName, connectionName } = options;
+    public async raiseEvent(
+        instanceId: string,
+        eventName: string,
+        eventData: unknown,
+        options: TaskHubOptions = {}
+    ): Promise<void> {
+        const { taskHubName, connectionName } = options;
         if (!eventName) {
             throw new Error("eventName must be a valid string.");
         }
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -599,8 +599,14 @@ export class DurableOrchestrationClient implements DurableClient {
         instanceId: string,
         waitOptions: WaitForCompletionOptions = {}
     ): Promise<HttpResponse> {
-        const timeoutInMilliseconds: number = waitOptions.timeoutInMilliseconds || 10000;
-        const retryIntervalInMilliseconds: number = waitOptions.retryIntervalInMilliseconds || 1000;
+        const timeoutInMilliseconds: number =
+            waitOptions.timeoutInMilliseconds !== undefined
+                ? waitOptions.timeoutInMilliseconds
+                : 10000;
+        const retryIntervalInMilliseconds: number =
+            waitOptions.retryIntervalInMilliseconds !== undefined
+                ? waitOptions.retryIntervalInMilliseconds
+                : 1000;
 
         if (retryIntervalInMilliseconds > timeoutInMilliseconds) {
             throw new Error(
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -26,7 +26,7 @@ import {
     DurableClientInput,
     DurableClient,
     GetStatusOptions,
-    SelectionOptions,
+    OrchestrationFilter,
     TaskHubOptions,
     SignalEntityOptions,
     WaitForCompletionOptions,
@@ -219,9 +219,9 @@ export class DurableOrchestrationClient implements DurableClient {
     }
 
     public async getStatusBy(
-        selectionOptions: SelectionOptions
+        filterOptions: OrchestrationFilter
     ): Promise<DurableOrchestrationStatus[]> {
-        const response = await this.getStatusInternal(selectionOptions);
+        const response = await this.getStatusInternal(filterOptions);
         switch (response.status) {
             case 200:
                 return response.data as DurableOrchestrationStatus[];
@@ -254,10 +254,10 @@ export class DurableOrchestrationClient implements DurableClient {
     }
 
     public async purgeInstanceHistoryBy(
-        selectionOptions: SelectionOptions
+        filterOptions: OrchestrationFilter
     ): Promise<PurgeHistoryResult> {
         let requestUrl: string;
-        const { createdTimeFrom, createdTimeTo, runtimeStatus } = selectionOptions;
+        const { createdTimeFrom, createdTimeTo, runtimeStatus } = filterOptions;
         if (this.clientData.rpcBaseUrl) {
             // Fast local RPC path
             let path = new URL("instances/", this.clientData.rpcBaseUrl).href;
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -28,7 +28,6 @@ import {
     GetStatusOptions,
     OrchestrationFilter,
     TaskHubOptions,
-    SignalEntityOptions,
     WaitForCompletionOptions,
 } from "durable-functions";
 import { WebhookUtils } from "./webhookutils";
@@ -478,9 +477,11 @@ export class DurableOrchestrationClient implements DurableClient {
 
     public async signalEntity(
         entityId: EntityId,
-        options: SignalEntityOptions = {}
+        operationName?: string,
+        operationContent?: unknown,
+        options: TaskHubOptions = {}
     ): Promise<void> {
-        const { operationName, operationContent, taskHubName, connectionName } = options;
+        const { taskHubName, connectionName } = options;
         let requestUrl: string;
         if (this.clientData.rpcBaseUrl) {
             // Fast local RPC path
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -21,7 +21,17 @@ import {
     PurgeHistoryResult,
     Utils,
 } from "./classes";
-import { StartNewOptions, DurableClientInput, DurableClient } from "durable-functions";
+import {
+    StartNewOptions,
+    DurableClientInput,
+    DurableClient,
+    GetStatusOptions,
+    SelectionOptions,
+    RaiseEventOptions,
+    TaskHubOptions,
+    SignalEntityOptions,
+    WaitForCompletionOptions,
+} from "durable-functions";
 import { WebhookUtils } from "./webhookutils";
 
 /** @hidden */
@@ -179,17 +189,13 @@ export class DurableOrchestrationClient implements DurableClient {
 
     public async getStatus(
         instanceId: string,
-        showHistory?: boolean,
-        showHistoryOutput?: boolean,
-        showInput?: boolean
+        options: GetStatusOptions = {}
     ): Promise<DurableOrchestrationStatus> {
-        const options: GetStatusInternalOptions = {
+        const internalOptions: GetStatusInternalOptions = {
             instanceId,
-            showHistory,
-            showHistoryOutput,
-            showInput,
+            ...options,
         };
-        const response = await this.getStatusInternal(options);
+        const response = await this.getStatusInternal(internalOptions);
 
         switch (response.status) {
             case 200: // instance completed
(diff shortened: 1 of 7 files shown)


src/durableorchestrationclient.ts
```diff
@@ -214,17 +220,13 @@ export class DurableOrchestrationClient implements DurableClient {
...
-        createdTimeFrom: Date | undefined,
-        createdTimeTo: Date | undefined,
-        runtimeStatus: OrchestrationRuntimeStatus[]
+        selectionOptions: SelectionOptions
     ): Promise<DurableOrchestrationStatus[]> {
-        const options: GetStatusInternalOptions = {
-            createdTimeFrom,
-            createdTimeTo,
-            runtimeStatus,
+        const internalOptions: GetStatusInternalOptions = {
```
ejizba: nit: You can just pass `selectionOptions` directly

src/durableorchestrationclient.ts
```diff
@@ -324,13 +325,8 @@ export class DurableOrchestrationClient implements DurableClient {
         }
     }
 
-    public async raiseEvent(
```
ejizba: IMO the options object is mostly just for optional arguments. `instanceId`, `eventName`, and `eventData` seem pretty important to the method, so I don't think I'd put them on `options`. I would default to this:

```
raiseEvent(instanceId: string, eventName: string, eventData: unknown, options?: TaskHubOptions): Promise<void>;
```

But if we want to try to reduce the number of arguments I would consider something like this as well:
```
interface OrchestrationEvent {
    name: string;
    data: unknown;
}

raiseEvent(instanceId: string, event: OrchestrationEvent, options?: TaskHubOptions): Promise<void>;
```

src/durableorchestrationclient.ts
```diff
@@ -603,9 +597,11 @@ export class DurableOrchestrationClient implements DurableClient {
     public async waitForCompletionOrCreateCheckStatusResponse(
         request: HttpRequest,
         instanceId: string,
-        timeoutInMilliseconds = 10000,
-        retryIntervalInMilliseconds = 1000
+        waitOptions: WaitForCompletionOptions = {}
     ): Promise<HttpResponse> {
+        const timeoutInMilliseconds: number = waitOptions.timeoutInMilliseconds || 10000;
```
ejizba: For both of these: Is `0` valid? If so, the `||` won't work.

types/durableClient.d.ts
```diff
@@ -41,17 +41,9 @@ export declare class DurableClient {
     /**
      * Gets the status of the specified orchestration instance.
      * @param instanceId The ID of the orchestration instance to query.
-     * @param showHistory Boolean marker for including execution history in the
-     *  response.
-     * @param showHistoryOutput Boolean marker for including input and output
-     *  in the execution history response.
+     * @param options options object specifying optional extra configuration options
```
ejizba: https://www.youtube.com/watch?v=G9VKmrZMhUs

types/durableClient.d.ts
```diff
@@ -280,3 +235,109 @@ export declare class HttpManagementPayload {
...
+     */
+    connectionName?: string;
+}
+
+/**
+ * Options object passed to DurableClient APIs
+ * to select certain orchestrations on which to perform
+ * actions
+ */
+export interface SelectionOptions {
```
ejizba: This name isn't super clear to me. What if we rename it to `OrchestrationFilter`?

test/testobjects/TestOrchestrations.ts
```diff
@@ -316,10 +328,14 @@ export class TestOrchestrations {
         const retryOptions = new df.RetryOptions(100, 5);
         const output = [];
         output.push(
-            context.df.callSubOrchestratorWithRetry("SayHelloInline", retryOptions, "Tokyo")
+            context.df.callSubOrchestratorWithRetry("SayHelloInline", retryOptions, {
```
ejizba: Anyone else conflicted on `input`? Like yes, it's technically optional - but it also seems like it's an important argument to the function. How common is it to pass an input? I could see myself getting annoyed at doing `{ input: ...` all the time

types/orchestration.d.ts
```diff
@@ -205,29 +166,22 @@ export declare class DurableOrchestrationContext {
...
      * @param name The name of the orchestrator function to call.
      * @param retryOptions The retry options for the orchestrator function.
-     * @param input The JSON-serializable input to pass to the orchestrator
-     * function.
-     * @param instanceId A unique ID to use for the sub-orchestration instance.
+     * @param options optional object to control the scheduled orchestrator (e.g provide input, instanceID)
      */
     callSubOrchestratorWithRetry(
         name: string,
         retryOptions: RetryOptions,
```
ejizba: It's a bit weird to have `retryOptions` _and_ `options`. Any chance we can combine them? I know it's a bit complicated because `RetryOptions` is a class...

src/durableorchestrationclient.ts
davidmrdavid: I'm ok either way.

src/durableorchestrationclient.ts
davidmrdavid: I do think `0` is valid., although a silly choice. Can we check for a falsey 0 explicitly?

types/durableClient.d.ts
davidmrdavid: I believe this may actually be able to return Entity data so I would rather keep this a bit more general.
Selection might be a bit too broad though. Does `QueryOptions` seem better? After all, this is affecting the instance-querying APIs

types/orchestration.d.ts
davidmrdavid: I'd be ok just placing the retryOptions inside the new options object.

test/testobjects/TestOrchestrations.ts
davidmrdavid: Yes, @hossam-nasr and me were just talking about this.
I do think passing inputs to sub-orchestrations is common, especially because they're used to offload work that you do not want your orchestrator to compute itself, so they usually get passed in some work batch as input.

Just to ask the obvious question: how do you feel about an overload here? After all, it is both an optional argument, but also a common one, so I think it's a fair candidate.

davidmrdavid (commented):
Left some comments on Eric's thread to help find consensus on the API changes

src/durableorchestrationclient.ts
davidmrdavid: Not a big fan of this final option with a `OrchestrationEvent` type - that feels like 2 options objects instead of one.
I don't feel too strongly one way or another for the remaining options. @hossam-nasr - what do you think?

test/testobjects/TestOrchestrations.ts
ejizba: Well `input` is type `unknown` which makes it really difficult to do an overload. Both `input` and `options` could be an object, so we don't really know which one a user is passing us. Technically since `input` could be anything, an options object _is_ a valid input

src/durableorchestrationclient.ts
ejizba: Our goal here should be to create logical groupings of information, not "throw all arguments onto an options object". I don't know if `OrchestrationEvent` is the right grouping, but I think we should absolutely consider multiple objects as arguments for all APIs. If we're comparing to .NET - don't they pass multiple classes all the time? The only difference is that Node.js uses an `interface` as its default grouping instead of a `class`

Below are some scenarios where I'm re-using particular pieces of information. I'm hoping you can tell me which scenarios actually make sense in durable, to help us decide the best grouping

### A. Same task hub options, different everything else
```TS
const taskHubOptions = { taskHubName: 'example' };
await client.raiseEvent('instanceId1', { name: 'event1', data: 'Hello' }, taskHubOptions);
await client.raiseEvent('instanceId2', { name: 'event2', data: 'World' }, taskHubOptions);
```

### B. Same event, different instances
```TS
const event = { name: 'event1', data: 'Hello' };
await client.raiseEvent('instanceId1', event);
await client.raiseEvent('instanceId2', event);
```

### C. Same instance, different events
```TS
const instanceId = 'instanceId1';
await client.raiseEvent(instanceId, { name: 'event1', data: 'Hello' });
await client.raiseEvent(instanceId, { name: 'event2', data: 'World' });
```

### D. Same instance & event name, different event data

```TS
const instanceId = 'instanceId1';
const eventName = 'event1';
await client.raiseEvent(instanceId, eventName, 'Hello');
await client.raiseEvent(instanceId, eventName, 'World');
```

types/durableClient.d.ts
ejizba: I'm not a fan of calling this `Options` because it's a required argument to both `getStatusBy` and `purgeInstanceHistoryBy`. If you like `Query` better than `Filter`, I'd be okay with `OrchestrationQuery`

If we do an `Options` object for this method, I think it should be a second argument. Options arguments typically only have optional/static information that affects how the method works across all method calls, but nothing really unique to any particular method call. For example, you might you use the same `timeout` value for all calls, but the `createdTimeFrom` will change often based on each particular call.

> I believe this may actually be able to return Entity data so I would rather keep this a bit more general.

I was assuming it was orchestration-only because of `runtimeStatus?: OrchestrationRuntimeStatus[];`, plus all the comments mentioning `orchestrations`. We could name this just `Query` or `Filter`, but I guess that feels too generic to me

src/durableorchestrationclient.ts
davidmrdavid: +1 to your goal statement.

Regarding the scenarios - these are all valid in DF. In general, `raiseEvent` is just a generic "send and forget" API and so all these scenarios could make sense depending on the application, and the event receiver's design.

As all these scenarios are valid, I recommend keeping it simple and opting for your first suggestion:
`raiseEvent(instanceId: string, eventName: string, eventData: unknown, options?: TaskHubOptions): Promise<void>;`

I think that's the least assuming choice: it only bundles the optional parameters into an options object, which I think is the  simplest design.

types/durableClient.d.ts
davidmrdavid: I looked at the C# API and it appears this only returns orchestration data, so I take back my comment on Entities.

I'm good with `OrchestrationFilter`, so let's go for it.

test/testobjects/TestOrchestrations.ts
davidmrdavid: Good point on the `unknown` type.
At this time - I would recommend we make an exception for the `input` parameter have it stay as a positional parameter despite being optional. I agree it's an imperfect choice, but we have [other threads](https://github.com/Azure/azure-functions-durable-js/issues/418) where improvements to this API are being discussed where I think this imperfection can be better tackled.

src/durableorchestrationclient.ts
hossam-nasr: Done

src/durableorchestrationclient.ts
hossam-nasr: This all makes sense to me. This decision was one of those I was conflicted about, but saving the options object for just the optional arguments makes sense to me. I made the update agreed upon:

```TS
raiseEvent(instanceId: string, eventName: string, eventData: unknown, options?: TaskHubOptions): Promise<void>
```

src/durableorchestrationclient.ts
hossam-nasr: Done

src/durableorchestrationclient.ts
hossam-nasr: I updated the check to check for `undefined` explicitly and allow a value of 0

types/durableClient.d.ts
hossam-nasr: Simplified the wording here lol

types/durableClient.d.ts
hossam-nasr: Done, renamed to OrchestrationFilter

test/testobjects/TestOrchestrations.ts
hossam-nasr: Hmm I don't disagree that it's annoying to have to type `{ input: ... }` every time, especially if you're not specifying an `instanceId`, and I was conflicted about this decision too. But I saw this as just being the orchestration equivalent of `client.startNew()` which we did change. If we're going to leave this as-is, what should we do about `client.startNew()`? I think the old implementation of `client.startNew()` was especially problematic because the order of the arguments was different than here: it was the instanceId first, _then_ the input, meaning you always had to write something like: `client.startNew(name, undefined, input)` if you didn't want to specify the `instanceId`, which was recommended anyway.

If we're going to keep this as-is, should we also flip the order of arguments for `client.startNew()`?

types/orchestration.d.ts
hossam-nasr: I'd be okay with having the `retryOptions` under the `options` object too, but in that case, `callSubOrchestrator()` and `callSubOrchestratorWithRetry()` start to look awfully similar. How do we feel about combining these two APIs in this case? I discussed this with @davidmrdavid offline too. Ofc this can still be done in a separate PR.

types/durableClient.d.ts
```diff
@@ -61,18 +53,11 @@ export declare class DurableClient {
...
-    getStatusBy(
-        createdTimeFrom: Date | undefined,
-        createdTimeTo: Date | undefined,
-        runtimeStatus: OrchestrationRuntimeStatus[]
-    ): Promise<DurableOrchestrationStatus[]>;
+     *
+     * @param filterOptions the OrchestrationFilter object specifying which
+     * orchestrations to retrieve.
+     */
+    getStatusBy(filterOptions: OrchestrationFilter): Promise<DurableOrchestrationStatus[]>;
```
ejizba: nit: I would just call this `filter`, not `filterOptions`

ejizba (approved):

types/durableClient.d.ts
hossam-nasr: Done

PR #487, opened 2023-03-07
Incorrect ReturnType of `DurableOrchestrationClient.getStatus`

### LINKED ISSUE #486: Incorrect ReturnType of `DurableOrchestrationClient.getStatus`
**Describe the bug**
`DurableOrchestrationClient` interface states that `getStatus()` returns `Promise<DurableOrchestrationStatus>`. This is correct for instances that exist and weren't purged. But for purged instances `getStatus()` returns an empty string. This is highly unexpected and should be fixed.

**Investigative information**

- Durable Functions extension version: 4.0.5030
- durable-functions npm module version: 1.5.4
- Language (JavaScript/TypeScript) and version: 4.8.4 (Typescript)
- Node.js version: 16.15.1

**To Reproduce**

```javascript
const client = getClient(context);
const instanceId = await client.startNew("MY_ORCHESTRATOR");

await client.getStatus(instanceId) // -> returns DurableOrchestrationStatus object

await client.purgeInstanceHistory(instanceId)
await client.getStatus(instanceId) // -> returns ""
```

**Expected behavior**
Interface of `DurableOrchestrationClient.getStatus()` represents the actual ReturnType of the method.

christianemmert:
- Fixes the ReturnType of `DurableOrchestrationClient.getStatus` when no `response.data` is sent back by the client e.g. no instance found or instance is purged

Resolves #486

### DIFF
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -211,7 +211,7 @@ export class DurableOrchestrationClient {
         showHistory?: boolean,
         showHistoryOutput?: boolean,
         showInput?: boolean
-    ): Promise<DurableOrchestrationStatus> {
+    ): Promise<DurableOrchestrationStatus | undefined> {
         const options: GetStatusOptions = {
             instanceId,
             showHistory,
@@ -226,6 +226,9 @@ export class DurableOrchestrationClient {
             case 400: // instance failed or terminated
             case 404: // instance not found or pending
             case 500: // instance failed with unhandled exception
+                if (!response.data || !Object.keys(response.data).length) {
+                    return undefined;
+                }
                 return response.data as DurableOrchestrationStatus;
             default:
                 return Promise.reject(this.createGenericError(response));
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -211,7 +211,7 @@ export class DurableOrchestrationClient {
         showHistory?: boolean,
         showHistoryOutput?: boolean,
         showInput?: boolean
-    ): Promise<DurableOrchestrationStatus | undefined> {
+    ): Promise<DurableOrchestrationStatus> {
         const options: GetStatusOptions = {
             instanceId,
             showHistory,
@@ -223,13 +223,10 @@ export class DurableOrchestrationClient {
         switch (response.status) {
             case 200: // instance completed
             case 202: // instance in progress
+                return response.data as DurableOrchestrationStatus;
             case 400: // instance failed or terminated
             case 404: // instance not found or pending
             case 500: // instance failed with unhandled exception
-                if (!response.data || !Object.keys(response.data).length) {
-                    return undefined;
-                }
-                return response.data as DurableOrchestrationStatus;
             default:
                 return Promise.reject(this.createGenericError(response));
         }
--- a/samples/UnitTesting/DurableFunctionsHttpStart/index.ts
+++ b/samples/UnitTesting/DurableFunctionsHttpStart/index.ts
@@ -1,5 +1,5 @@
-﻿import * as df from "durable-functions"
-import { AzureFunction, Context, HttpRequest } from "@azure/functions"
+﻿import * as df from "durable-functions";
+import { AzureFunction, Context, HttpRequest } from "@azure/functions";
 
 const httpStart: AzureFunction = async function (context: Context, req: HttpRequest): Promise<any> {
     const client = df.getClient(context);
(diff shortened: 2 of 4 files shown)


christianemmert:
@microsoft-github-policy-service agree

src/durableorchestrationclient.ts
```diff
@@ -226,6 +226,9 @@ export class DurableOrchestrationClient {
             case 400: // instance failed or terminated
             case 404: // instance not found or pending
             case 500: // instance failed with unhandled exception
+                if (!response.data || !Object.keys(response.data).length) {
+                    return undefined;
+                }
                 return response.data as DurableOrchestrationStatus;
             default:
                 return Promise.reject(this.createGenericError(response));
```
davidmrdavid: Is there a reason why you would prefer to get `undefined` instead of throwing an exception? The latter seems, to me, more helpful as it forces the end user to respond / deal with the exception, whereas `undefined` variables can cause silent errors.

davidmrdavid (commented):
Thanks for your contribution. Just have a question about the design:

src/durableorchestrationclient.ts
christianemmert: The idea was to handle this issue more gracefully, which would not break existing implementations. But I agree that throwing an exception, in this case, would be a more idiomatic approach.

davidmrdavid: That makes sense. I understand the concern not to break existing implementations but, in this case, upgrading to a package does not happen automatically so I'm not personally too concerned about throwing this exception starting on a newer release.

@christianemmert: I would personally feel more comfortable merging a new exception-throwing case than introducing an explicit `undefined` here. If my reasons above seem sound to you, would you mind updating your PR here? Otherwise, I can think about this `undefined` alternative some more and make the final edits myself. Thanks!

christianemmert: Even though a package update does not happen automatically, a user of this package expects a package maintainer to follow semver rules. This means introducing a breaking change by throwing an additional exception should be done in a major release.

In my opinion, this is a nasty bug that should be fixed sooner rather than later, e.g. in a patch release.

ejizba: People might be handling the empty string right now, so changing to undefined would break them as well. I agree this fix should be done in a major release, and I think all the solutions discussed so far are equally breaking.

The good news is we are actually about to announce public preview of the next major version. @hossam-nasr @davidmrdavid can we fix this for v3 ASAP?

christianemmert: You are right people might have dealt with this issue in different ways as such either way it would break stuff for some or another.

Changed the fix to throw an error instead: [a95be58](https://github.com/Azure/azure-functions-durable-js/pull/487/commits/a95be58d4cbf5cfb612a58b6f8860c34fe727a49)

davidmrdavid: Thanks for the fix @christianemmert.
Based on @ejizba's suggestion, I think this is a good candidate for the v3 version that's, as of today, in public preview. Let me re-target this PR for that branch.

davidmrdavid: Hmm, seems that might not easy to do through the GitHub UI, due to the divergence between the two v2 and v3 branches. causing merge conflicts, so for now I'm keeping this branch as targeting v2. 

We might need to do some manual porting. @christianemmert - if you would like to continue driving this change, you're welcome to re-target this diff to V3, although it will require cherry-picking your commits. Otherwise, I'm happy to take over this PR and do that myself. Please let me know if you have a preference.

hossam-nasr:
Hi @christianemmert ,

Thank you so much for your contribution!

As others have pointed out, this would be a breaking change, so this change would have to be done on the new preview `v3.x` version. This is definitely a bugfix that we would like to see happen in `v3.x`, but I am uncomfortable accepting this change as-is without more manual testing and investigation. This PR would throw an error any time the client receives a 400, 404, or 500 response. But, there are still cases where those status codes contain valid `DurableOrchestrationStatus` objects that we should pass back to the user. We receive 400 when the instance failed or terminated, 404 when the instance is pending, and 500 when the instance failed with an unhandled exception. 

There are two parts I see here:

1. The SDK should throw an error when there's an empty response for whatever reason. This should guard us against potential unexpected behavior from the extension and still live up to the promise of always returning a `DurableOrchestrationStatus`
2. That alone may not be the bugfix we seek, since it's possible the extension should be returning a valid `DurableOrchestrationStatus`, but isn't for some reason. So we need to investigate the cases in which an empty response is returned, and identify if that's expected behavior or a bug that should be fixed on the extension side.

Based on this, I'll go ahead and close this PR. The good news is that with the preview announcement of `v3.x` today, we should have more time to investigate this further and apply a fix!

PR #502, opened 2023-03-27
Update 2.x build pipeline

hossam-nasr:
Update 2.x branch build pipeline to:

1. Automatically trigger builds on `v2.x` and `v3.x` branches, instead of `main` and `dev`
2. Add Node 18 to the testing matrix

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,11 +13,9 @@ pr:
     branches:
         include:
             - v2.x
-            - v3.x
 
 trigger:
     - v2.x
-    - v3.x
 
 jobs:
     - job: Test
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -1,16 +1,23 @@
...
+        NODE_18: "18.x",
+    }
 name: $(MODULE_VERSION)-$(Date:yyyyMMdd)$(Rev:.r)
 
 pr:
     branches:
         include:
-            - main
-            - dev
+            - v2.x
```
ejizba, line 15: You don't need `v3.x` on this branch

For `trigger`:
> The version of the pipeline in the pushed branch is used.

For `pr`:
> The version of the pipeline in the source branch for the pull request is used.

https://learn.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops#branch-consideration-for-triggers-in-yaml-pipelines

ejizba (approved):

azure-pipelines.yml
hossam-nasr: Removed 3.x

PR #512, opened 2023-05-09
Update `client.getStatus` to be more consistent with type contract and more type-safe

### LINKED ISSUE #486: Incorrect ReturnType of `DurableOrchestrationClient.getStatus`

hossam-nasr:
Resolves #486 , see issue for more details. In addition to the issue outlined in #486, this PR also updates the behavior of `client.getStatus()` as follows:

| **Scenario** | **Old behavior** | **New behavior** |
| -------------- | -------------------| ----------------------|
| Happy scenario: extension responds with 200 or 202, but the timestamps are strings |  Object is passed as-is to the customer, with timestamps as stings, breaking type contract | User is passed a valid `DurableOrchestrationStatus` object with timestamps as `Date` objects |
| Invalid or not found `instanceId`, extension responds with 404 and empty body | User is passed an empty string, breaking type contract (#486) | Exception is thrown |
| Extension encounters an unhandled exception, responds with 500 | User is passed the response body containing exception information as the status object, breaking type contract | Exception is thrown | 
| Extension responds with valid response code, but some fields are missing (e.g., missing `name` property) | Object is passed as-is to the user, breaking type contract | Exception is thrown |
| Extension responds with valid response code, but empty response body | User is passed an empty string, breaking type contract | Exception is thrown |

In addition to these behavior changes, this PR also:

* Updates the comments on the `getStatus()` method to accurately represent what each status code means:
    - 200: completed, failed, or terminated
    - 202: pending or running
    - 404: `instanceId` not found
    - 500: exception while the extension was processing the request
* Removes the `TODO` flag in `getStatus()` unit tests, and replaces it with unit tests to test every response code, as above
* Adds unit tests to test all scenarios in the table above

P.S.: This is a breaking change

### DIFF
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -206,8 +206,7 @@ export class DurableOrchestrationClient implements DurableClient {
                 }
                 return new DurableOrchestrationStatus(response.data);
             case 404: // instance not found
-                let msg =
-                    "The operation failed because the instanceId supplied was invalid or not found.";
+                let msg = `DurableClient error: Durable Functions extension replied with HTTP 404 response. This usually means we could not find any data associated with the instanceId provided: ${instanceId}.`;
                 if (response.data) {
                     msg += ` Details: ${JSON.stringify(response.data)}`;
                 }
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -212,7 +212,7 @@ export class DurableOrchestrationClient implements DurableClient {
                     msg += ` Details: ${JSON.stringify(response.data)}`;
                 }
                 throw new Error(msg);
-            case 500: // request failed with unhandled exception (data contains exception message)
+            case 500: // request failed with unhandled exception (response data contains exception details)
             default:
                 return Promise.reject(this.createGenericError(response));
         }
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -209,12 +209,14 @@ export class DurableOrchestrationClient implements DurableClient {
                     return new DurableOrchestrationStatus(response.data);
                 } catch (error) {
                     throw new Error(
-                        `DurableClient error: could not construct a DurableOrhcestrationStatus object using the data received from the Durable Functions extension: ${error.Message}`
+                        `DurableClient error: could not construct a DurableOrchestrationStatus object using the data received from the Durable Functions extension: ${error.message}`
                     );
                 }
 
             case 404: // instance not found
-                let msg = `DurableClient error: Durable Functions extension replied with HTTP 404 response. This usually means we could not find any data associated with the instanceId provided: ${instanceId}.`;
+                let msg =
+                    `DurableClient error: Durable Functions extension replied with HTTP 404 response. ` +
+                    `This usually means we could not find any data associated with the instanceId provided: ${instanceId}.`;
                 if (response.data) {
                     msg += ` Details: ${JSON.stringify(response.data)}`;
                 }
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -201,10 +201,18 @@ export class DurableOrchestrationClient implements DurableClient {
             case 400:
                 if (!response.data) {
                     throw new Error(
-                        "Operation failed. The received response contained empty response data."
+                        `DurableClient error: the Durable Functions extension replied with an empty HTTP ${response.status} response.`
                     );
                 }
-                return new DurableOrchestrationStatus(response.data);
+
+                try {
+                    return new DurableOrchestrationStatus(response.data);
+                } catch (error) {
+                    throw new Error(
+                        `DurableClient error: could not construct a DurableOrhcestrationStatus object using the data received from the Durable Functions extension: ${error.Message}`
+                    );
+                }
+
             case 404: // instance not found
                 let msg = `DurableClient error: Durable Functions extension replied with HTTP 404 response. This usually means we could not find any data associated with the instanceId provided: ${instanceId}.`;
                 if (response.data) {
--- a/src/durableorchestrationclient.ts
+++ b/src/durableorchestrationclient.ts
@@ -193,15 +193,26 @@ export class DurableOrchestrationClient implements DurableClient {
             instanceId,
             ...options,
         };
-        const response = await this.getStatusInternal(internalOptions);
+        const response: AxiosResponse = await this.getStatusInternal(internalOptions);
 
         switch (response.status) {
-            case 200: // instance completed
-            case 202: // instance in progress
-            case 400: // instance failed or terminated
-            case 404: // instance not found or pending
-            case 500: // instance failed with unhandled exception
-                return response.data as DurableOrchestrationStatus;
+            case 200: // instance completed, failed or terminated
+            case 202: // instance running or pending
+            case 400:
+                if (!response.data) {
+                    throw new Error(
+                        "Operation failed. The received response contained empty response data."
+                    );
+                }
+                return new DurableOrchestrationStatus(response.data);
+            case 404: // instance not found
+                let msg =
+                    "The operation failed because the instanceId supplied was invalid or not found.";
+                if (response.data) {
+                    msg += ` Details: ${JSON.stringify(response.data)}`;
+                }
+                throw new Error(msg);
+            case 500: // request failed with unhandled exception (data contains exception message)
             default:
                 return Promise.reject(this.createGenericError(response));
         }
--- a/src/durableorchestrationstatus.ts
+++ b/src/durableorchestrationstatus.ts
@@ -16,7 +16,7 @@ export class DurableOrchestrationStatus implements types.DurableOrchestrationSta
     constructor(init: unknown) {
         if (!this.isDurableOrchestrationStatusInit(init)) {
             throw new TypeError(
-                `Malformed data passed to DurableOrchestrationStatus constructor. Data received: ${JSON.stringify(
+                `Failed to construct a DurableOrchestrationStatus object because the initializer had invalid types or missing fields. Initializer received: ${JSON.stringify(
                     init
                 )}`
             );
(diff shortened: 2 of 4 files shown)


src/durableorchestrationclient.ts
```diff
@@ -193,15 +193,26 @@ export class DurableOrchestrationClient implements DurableClient {
...
-            case 400: // instance failed or terminated
-            case 404: // instance not found or pending
-            case 500: // instance failed with unhandled exception
-                return response.data as DurableOrchestrationStatus;
+            case 200: // instance completed, failed or terminated
+            case 202: // instance running or pending
+            case 400:
+                if (!response.data) {
+                    throw new Error(
+                        "Operation failed. The received response contained empty response data."
```
davidmrdavid: nit I think the following would make the error more concrete.
```suggestion
                        "DurableClient error: the Durable Functions extension replied with an empty HTTP 400 response."
```

davidmrdavid: Do we have a repro producing an HTTP 400 response? From the internet, HTTP 400 seems to correspond to a malformed client request so I doubt there's a "real" DurableOrchestrationStatus response in here. If this is just an error, maybe we should treat this the same as HTTP 500 and throw a generic error

src/durableorchestrationclient.ts
```diff
@@ -193,15 +193,26 @@ export class DurableOrchestrationClient implements DurableClient {
...
+                    );
+                }
+                return new DurableOrchestrationStatus(response.data);
+            case 404: // instance not found
+                let msg =
+                    "The operation failed because the instanceId supplied was invalid or not found.";
```
davidmrdavid:
```suggestion
                    "DurableClient error: Durable Functions extension replied with HTTP 404 response. This usually means we could not find any data associated with the instanceId provided.

```

davidmrdavid: Idea: can we log the instanceId provided as part of this error?

src/durableorchestrationstatus.ts
```diff
@@ -2,16 +2,61 @@ import { OrchestrationRuntimeStatus } from "./classes";
...
-        public readonly instanceId: string,
-        public readonly createdTime: Date,
-        public readonly lastUpdatedTime: Date,
-        public readonly input: unknown,
-        public readonly output: unknown,
-        public readonly runtimeStatus: OrchestrationRuntimeStatus,
-        public readonly customStatus?: unknown,
-        public readonly history?: Array<unknown>
-    ) {}
+    constructor(init: unknown) {
```
davidmrdavid, line 16: nit: what is `init` short for in this case? Is it initialzier? For readability, could we fully spell out this name?

src/durableorchestrationstatus.ts
```diff
@@ -2,16 +2,61 @@ import { OrchestrationRuntimeStatus } from "./classes";
...
+        this.createdTime = new Date(init.createdTime);
+        this.lastUpdatedTime = new Date(init.lastUpdatedTime);
+        this.input = init.input;
+        this.output = init.output;
+        this.runtimeStatus = init.runtimeStatus as OrchestrationRuntimeStatus;
+        this.customStatus = init.customStatus;
+        this.history = init.history;
+    }
+
+    private isDurableOrchestrationStatusInit(obj: unknown): obj is DurableOrchestrationStatusInit {
```
davidmrdavid, line 36: nit: same feedback here regarding the `Init` suffix.

src/durableorchestrationstatus.ts
```diff
@@ -2,16 +2,61 @@ import { OrchestrationRuntimeStatus } from "./classes";
...
+            (typeof objAsInit.lastUpdatedTime === "string" ||
+                objAsInit.lastUpdatedTime instanceof Date) &&
+            objAsInit.input !== undefined &&
+            objAsInit.output !== undefined &&
+            typeof objAsInit.runtimeStatus === "string"
+        );
+    }
+}
+
+interface DurableOrchestrationStatusInit {
```
davidmrdavid, line 52: nit: same feedback regarding `Init` suffix.

src/durableorchestrationstatus.ts
```diff
@@ -2,16 +2,61 @@ import { OrchestrationRuntimeStatus } from "./classes";
...
-        public readonly runtimeStatus: OrchestrationRuntimeStatus,
-        public readonly customStatus?: unknown,
-        public readonly history?: Array<unknown>
-    ) {}
+    constructor(init: unknown) {
+        if (!this.isDurableOrchestrationStatusInit(init)) {
+            throw new TypeError(
+                `Malformed data passed to DurableOrchestrationStatus constructor. Data received: ${JSON.stringify(
+                    init
+                )}`
```
davidmrdavid: Are we able to surface specifically what check failed in our validation check?

davidmrdavid (commented):
Thanks! Some initial feedback

davidmrdavid:
@hossam-nasr:

I'm curious to understand this scenario from your PR description a bit more:

"""
Extension responds with valid response code, but some fields are missing (e.g., missing name property)
"""

Is this a valid response you've _seen_ from the extension, or just a hypothetical scenario based on the previous implementation? If it's a response you've observed, what triggers it?

src/durableorchestrationclient.ts
hossam-nasr: This code path is for 200, 202 and 400 responses, so this error will throw if there's an empty 200, 202 or 400 response for some reason. I couldn't get 400 to repro (and I'm not sure when we _should_ expect a 400), but I left it in this path because previously, 400 was labeled for "instance failed or terminated." If you're confident that there is no valid response with 400 then I'm happy treating 400 the same as 500. I'd still leave this error for 200 and 202 though

src/durableorchestrationclient.ts
hossam-nasr: Accepted and added the instanceId

src/durableorchestrationstatus.ts
hossam-nasr: You want it to be called `DurableOrchestrationStatusInitializer`? That feels unnecessarily long tbh. What else could `Init` mean? 

P.S. I was following `@azure/function`'s example here. E.g., see [`HttpResponseInit`](https://github.com/Azure/azure-functions-nodejs-library/blob/5e160396d9c431170e37b76ae06d1cbfdf2d1de5/types/http.d.ts#L197). Also `BodyInit` from `unidici`

src/durableorchestrationstatus.ts
hossam-nasr: We could, but that means I would need to have a separate `if` statement for each field. Is it worth it?

src/durableorchestrationclient.ts
davidmrdavid: Oh I see, thanks for clarifying. For safety, it might be best to keep the 400 in there then.
Just curious: if you get the status of a terminated orchestrator, do you get a 400? It would be good to confirm that comment with a test for the termination case specifically.

Given that this code path is re-used for 200, 202 and 400 - then I'd change my edit suggestion to not to include the HTTP 400 part. Would it be possible to have something **like** this for the error message instead:

```JavaScript
"DurableClient error: the Durable Functions extension replied with an empty HTTP {response.status} response."
```

src/durableorchestrationstatus.ts
davidmrdavid: Well, if there's precedent for this in the `@azure/function` library, then I'm ok with it :)

src/durableorchestrationstatus.ts
davidmrdavid: resolving as per this: https://github.com/Azure/azure-functions-durable-js/pull/512#discussion_r1188923325

src/durableorchestrationstatus.ts
davidmrdavid: resolving as per: https://github.com/Azure/azure-functions-durable-js/pull/512#discussion_r1188923325

hossam-nasr:
> """ Extension responds with valid response code, but some fields are missing (e.g., missing name property) """
> 
> Is this a valid response you've _seen_ from the extension, or just a hypothetical scenario based on the previous implementation? If it's a response you've observed, what triggers it?

This isn't a valid response and shouldn't be expected in regular behavior, but this is a safeguard I've added to protect us from similar situations as #486 just in case

src/durableorchestrationstatus.ts
davidmrdavid: If it's too cumbersome, then it isn't given that we're logging the data received. I trust your judgement on that.
In that case, I would recommend rephrasing that error message to be a bit more literal. Please consider the following:

```JavaScript
                `Failed to construct a DurableOrchestrationStatus because the initializer did not type-check. Initializer received: ${JSON.stringify(

```

src/durableorchestrationclient.ts
hossam-nasr: Nope, I confirmed that `Terminated` and `Failed` both return a 200 response, and I've also added tests for both of them, and updated the comments accordingly. 

Done! I've also added a try/catch statement to catch and re-throw the error from the `DurableOrchestrationStatus` constructor to follow the same pattern: `DurableClient error: ...`

src/durableorchestrationstatus.ts
hossam-nasr: Done. I modified your message to explicitly state that the initializer had invalid types or missing fields.

davidmrdavid (approved):

PR #516, opened 2023-05-18
Remove support for node versions 10 and 12

hossam-nasr:
Those versions have reached EOL and are no longer supported by the Functions runtime

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 | Branch | Status                                                                                                                                                                                                                                    | Support level | Programming model | Node.js versions |
 | ------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------- | ----------------- | ---------------- |
-| v2.x   | [![Build Status](https://azfunc.visualstudio.com/Azure%20Functions/_apis/build/status/Azure.azure-functions-durable-js?branchName=v2.x)](https://azfunc.visualstudio.com/Azure%20Functions/_build/latest?definitionId=13&branchName=v2.x) | GA            | V1, V2, V3        | 14.x+            |
+| v2.x   | [![Build Status](https://azfunc.visualstudio.com/Azure%20Functions/_apis/build/status/Azure.azure-functions-durable-js?branchName=v2.x)](https://azfunc.visualstudio.com/Azure%20Functions/_build/latest?definitionId=13&branchName=v2.x) | GA            | V3                | 14.x+            |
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -1,13 +1,5 @@
...
-    {
-        MODULE_VERSION: "2.1.2",
-        NODE_10: "10.x",
-        NODE_12: "12.x",
-        NODE_14: "14.x",
-        NODE_16: "16.x",
-        NODE_18: "18.x",
-        NODE_20: "20.x",
-    }
+    { MODULE_VERSION: "2.1.2", NODE_14: "14.x", NODE_16: "16.x", NODE_18: "18.x", NODE_20: "20.x" }
```
ejizba: I would leave the tests as-is. Unless the tests are annoying (unreliable or take a super long time), I would only remove them in a new major version (which we already did in the v3.x branch). Keeping them in is an easy way to prevent us from breaking people on old versions and the branch itself will fade out as time goes on

For precedent, the old branch of the worker (v2.x) still tests on Node.js 8/10/12/14: https://github.com/Azure/azure-functions-nodejs-worker/blob/v2.x/azure-pipelines/test.yml#L1

README.md
```diff
@@ -1,6 +1,6 @@
...
 | ------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------- | ----------------- | ---------------- |
-| v2.x   | [![Build Status](https://azfunc.visualstudio.com/Azure%20Functions/_apis/build/status/Azure.azure-functions-durable-js?branchName=v2.x)](https://azfunc.visualstudio.com/Azure%20Functions/_build/latest?definitionId=13&branchName=v2.x) | GA            | V1, V2, V3        | 8.x+             |
+| v2.x   | [![Build Status](https://azfunc.visualstudio.com/Azure%20Functions/_apis/build/status/Azure.azure-functions-durable-js?branchName=v2.x)](https://azfunc.visualstudio.com/Azure%20Functions/_build/latest?definitionId=13&branchName=v2.x) | GA            | V1, V2, V3        | 14.x+            |
```
ejizba: The debate is whether documentation should be historical versions or current versions. I don't really care, but I want us to be consistent (at least within this file). So if you're doing "current" versions, you should change the programming model column to just `V3` and also update line 36 to be core tools "4.x"

azure-pipelines.yml
davidmrdavid: +1

hossam-nasr: Done, reverted this change

README.md
hossam-nasr: I don't have a strong opinion on this either. @davidmrdavid thoughts?

davidmrdavid: Let's have this table represent currently supported versions, @hossam-nasr

hossam-nasr: Done. updated to list v3 prog model only and v4 core tools

ejizba (approved):

PR #525, opened 2023-07-22
Refactor "shim" file into modules

hossam-nasr:
Similar to https://github.com/Azure/azure-functions-nodejs-library/pull/50 on the node library. This PR doesn't make any code or functionality changes, just reorganizes code. It removes the use of namespaces, and gets rid of the "shim.ts" file in favor of exporting modules for each "namespace"

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,9 +6,12 @@ import {
     DummyOrchestrationContext,
 } from "./classes";
 import { getClient } from "./durableorchestrationclient";
-import { app, input, trigger } from "./shim";
-import { DummyEntityContext } from "./util/testingUtils";
+import { DummyEntityContext, createEntityFunction, createOrchestrator } from "./util/testingUtils";
 import { ManagedIdentityTokenSource } from "./tokensource";
+
+export * as app from "./app";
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -19,7 +22,6 @@ export {
     RetryOptions,
     DummyOrchestrationContext,
     DummyEntityContext,
-    app,
-    input,
-    trigger,
+    createOrchestrator,
```
ejizba: I don't think we were exporting `createOrchestrator` and `createEntityFunction` from the package before this PR. Pretty sure they're just used for internal testing, right?

ejizba (commented):
would prefer to see a clean commit history for a refactor pr. The last two commits make sense, but not sure what's up with the first 3

src/index.ts
hossam-nasr: You're right, I removed them. They were exported from `shim` but not from `index`. Also, I don't know what was going with the git history, it looks like my local `v3.x` branch that I branched out of was ahead from remote by 3 commits for some reason. Anyway, I rebased and should be fixed now.

ejizba (approved):

PR #526, opened 2023-07-22
Refactor "task.ts" into individual files

hossam-nasr:
No functional changes, just moving code around. Refactoring the big "task.ts" file into individual files for each Task class


ejizba (approved):
Not a huge fan of the index.ts file - would rather we just import the task classes directly from their own file instead of re-exporting like that

But overall the PR is a net improvement so still approving

PR #534, opened 2023-07-28
Move the `taskOrchestrationExecutor` from RetryableTask constructor to `trySetValue()`

hossam-nasr:
This doesn't change any functionality of the `RetryableTask`. Rather, this means that a `TaskOrchestrationExecutor` is not required in the constructor for `RetryableTask`, but can be passed as in argument when it's really needed: in the `trySetValue()` function. This means that a `RetryableTask` can be constructed _outside_ the context of a `DurableOrchestrationContext` (where the `TaskOrchestrationExecutor` is available), unblocking #418 

The reason all these changes are required is because this is the new flow of the `TaskOrchestrationExecutor`:

`TaskOrchestrationExecutor.execute()` calls `TaskBase.setValue()` with a reference of itself -> `TaskBase.setValue()` calls `TaskBase.propagate()` -> calls `CompoundTask.handleCompletion()` -> calls `CompoundTask.trySetValue()`

### DIFF
--- a/src/task/CompoundTask.ts
+++ b/src/task/CompoundTask.ts
@@ -1,3 +1,4 @@
+import { TaskOrchestrationExecutor } from "src/orchestrations/TaskOrchestrationExecutor";
 import { IAction } from "../actions/IAction";
 import { DFTask } from "./DFTask";
 import { TaskBase } from "./TaskBase";
@@ -40,11 +41,11 @@ export abstract class CompoundTask extends DFTask {
      * @param child
      *  A sub-task of this task.
      */
-    public handleCompletion(child: TaskBase): void {
+    public handleCompletion(child: TaskBase, executor?: TaskOrchestrationExecutor): void {
         if (!this.isPlayed) {
             this.isPlayed = child.isPlayed;
         }
-        this.trySetValue(child);
+        this.trySetValue(child, executor);
     }
 
     /**
@@ -55,5 +56,5 @@ export abstract class CompoundTask extends DFTask {
      * @param child
      *  A sub-task
      */
-    abstract trySetValue(child: TaskBase): void;
+    abstract trySetValue(child: TaskBase, executor?: TaskOrchestrationExecutor): void;
 }
--- a/src/task/CompoundTask.ts
+++ b/src/task/CompoundTask.ts
@@ -39,6 +39,8 @@ export abstract class CompoundTask extends DFTask {
      * @hidden
      * Tries to set this task's result based on the completion of a sub-task
      * @param child
+     * @param executor The TaskOrchestrationExecutor instance that is managing the replay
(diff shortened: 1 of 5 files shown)


src/task/CompoundTask.ts
```diff
@@ -40,11 +41,11 @@ export abstract class CompoundTask extends DFTask {
      * @param child
      *  A sub-task of this task.
      */
-    public handleCompletion(child: TaskBase): void {
+    public handleCompletion(child: TaskBase, executor?: TaskOrchestrationExecutor): void {
```
davidmrdavid, line 46: Why is executor an optional argument?

src/task/RetryableTask.ts
```diff
@@ -48,7 +44,14 @@ export class RetryableTask extends WhenAllTask {
      * @param child
      *  The sub-task that just completed
      */
-    public trySetValue(child: TaskBase): void {
+    public trySetValue(child: TaskBase, executor?: TaskOrchestrationExecutor): void {
+        if (!executor) {
+            throw new Error(
+                "No executor passed to RetryableTask.trySetValue. " +
+                    "A TaskOrchestrationExecutor is required to schedule new tasks."
+            );
```
davidmrdavid: Can we phrase this in such a way that it's clear to users that this is a framework-internal error instead of a user-generated error? Ideally we'd also recommend that they open an issue on github to report this. Just trying to practice giving actionable error messages.

davidmrdavid (commented):
Love this refactoring. Just a few questions before approval

src/task/CompoundTask.ts
hossam-nasr: Because not all `CompoundTask`s would expect or utilize this argument. I could make the same change for all `CompoundTask`s, but I wanted to localize this change to be as small as possible. This requires the least amount of changes, and localizes the change to just the `RetryableTask`. It is not optional for the `RetryableTask` though, because an error will be thrown if it is not defined.

src/task/RetryableTask.ts
hossam-nasr: Sure, do you have a specific wording in mind?

davidmrdavid: Just start with "framework internal exception: <what the error is>." and end with "Consider reporting this issue at <github url>"

src/task/CompoundTask.ts
davidmrdavid: Oh I see, I remember the context now. Makes sense. Could you please add a comment noting this down? Maybe you can add it to the signature-level comment itself?

hossam-nasr: Done

src/task/RetryableTask.ts
hossam-nasr: Done

davidmrdavid (approved):

PR #535, opened 2023-07-31
Improve calling activities and suborchestrators

### LINKED ISSUE #418: Can we improve the `callActivity()` and `callSubOrchestrator()` experiences in orchestrations?

hossam-nasr:
Resolves #418. This makes it such that the registeration functions `app.orchestration()` and `app.activity()` return a `RegisteredOrchestration` and `RegisteredActivity`, respectively, which can be called to return a yieldable `Task`, without having to call `context.df.callSubOrchestrator()`, `context.df.callSubOrchestratorWithRetry()`, `context.df.callActivity()`, or `context.df.callActivityWithRetry()`. 

Instead, you can call:

 ```TS
yield mySubOrchestrator(input);
yield mySubOrchestrator(input).withRetry(retryOptions);
yield myActivity(input);
yield myActivity(input).withRetry(retryOptions);
```

with `mySubOrchestraor` and `myActivity` being the result of registering the orchestration/activity. Detailed examples below:

#### Before

```TS
const helloActivityHandler: ActivityHandler = function (input: unknown): string {
    return `Hello ${input}`;
};
df.app.activity(helloActivityName, {
    handler: helloActivityHandler,
});

const sayHelloWithActivityHandler: OrchestrationHandler = function* (context: OrchestrationContext) {
    const input: unknown = context.df.getInput();
    const output: string = yield context.df.callActivity(helloActivityName, input);
    return output;
};
df.app.orchestration("sayHelloWithActivity", sayHelloWithActivityHandler);

const sayHelloWithSubOrchestratorHandler: OrchestrationHandler = function* (
    context: OrchestrationContext
) {
    const input: unknown = context.df.getInput();
    const output: string = yield context.df.callSubOrchestrator("sayHelloWithActivity", input);
    return output;
};
df.app.orchestration("sayHelloWithSubOrchestrator", sayHelloWithSubOrchestratorHandler);
```

#### After

```TS
const helloActivityHandler: ActivityHandler = function (input: unknown): string {
    return `Hello ${input}`;
};
const helloActivity: RegisteredActivity = df.app.activity(helloActivityName, {
    handler: helloActivityHandler,
});

const sayHelloWithActivityHandler: OrchestrationHandler = function* (context: OrchestrationContext) {
    const input: unknown = context.df.getInput();
    const output: string = yield helloActivity(input);
    return output;
};
const sayHelloWithActivityOrchestrator: RegisteredOrchestration = df.app.orchestration("sayHelloWithActivity", sayHelloWithActivityHandler);

const sayHelloWithSubOrchestratorHandler: OrchestrationHandler = function* (
    context: OrchestrationContext
) {
    const input: unknown = context.df.getInput();
    const output: string = yield sayHelloWithActivityOrchestrator(input);
    return output;
};
df.app.orchestration("sayHelloWithSubOrchestrator", sayHelloWithSubOrchestratorHandler);

```

#### Before

```TS
const flakyFunction: ActivityHandler = function (_input: any, context: InvocationContext): void {
    context.log("Flaky Function Flaking!");
    throw new Error("FlakyFunction flaked");
};
df.app.activity("flakyFunction", {
    handler: flakyFunction,
});

const callActivityWithRetry: OrchestrationHandler = function* (context: OrchestrationContext) {
    const retryOptions: RetryOptions = new df.RetryOptions(1000, 2);

    let returnValue: any;
    try {
        returnValue = yield context.df.callActivityWithRetry("flakyFunction", retryOptions);
    } catch (e) {
        context.log("Orchestrator caught exception. Flaky function is extremely flaky.");
    }

    return returnValue;
};
df.app.orchestration("callActivityWithRetry", callActivityWithRetry);

```

#### After

```TS
const flakyFunction: ActivityHandler = function (_input: any, context: InvocationContext): void {
    context.log("Flaky Function Flaking!");
    throw new Error("FlakyFunction flaked");
};
const flakyActivity: RegisteredActivity = df.app.activity("flakyFunction", {
    handler: flakyFunction,
});

const callActivityWithRetry: OrchestrationHandler = function* (context: OrchestrationContext) {
    const retryOptions: RetryOptions = new df.RetryOptions(1000, 2);

    let returnValue: any;
    try {
        returnValue = yield flakyActivity().withRetry(retryOptions);
    } catch (e) {
        context.log("Orchestrator caught exception. Flaky function is extremely flaky.");
    }

    return returnValue;
};
df.app.orchestration("callActivityWithRetry", callActivityWithRetry);
```

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -4,15 +4,19 @@ import {
     EntityOptions,
     OrchestrationHandler,
     OrchestrationOptions,
+    RegisteredActivity,
+    RegisteredOrchestration,
 } from "durable-functions";
 import * as trigger from "./trigger";
 import { createOrchestrator, createEntityFunction } from "./util/testingUtils";
 import { app as azFuncApp } from "@azure/functions";
+import { RegisteredOrchestrationTask } from "./task/RegisteredOrchestrationTask";
+import { RegisteredActivityTask } from "./task/RegisteredActivityTask";
 
 export function orchestration(
     functionName: string,
     handlerOrOptions: OrchestrationHandler | OrchestrationOptions
-): void {
+): RegisteredOrchestration {
     const options: OrchestrationOptions =
         typeof handlerOrOptions === "function" ? { handler: handlerOrOptions } : handlerOrOptions;
 
@@ -21,6 +25,15 @@ export function orchestration(
         ...options,
         handler: createOrchestrator(options.handler),
     });
+
+    const result: RegisteredOrchestration = (
+        input?: unknown,
+        instanceId?: string
+    ): RegisteredOrchestrationTask => {
+        return new RegisteredOrchestrationTask(functionName, input, instanceId);
+    };
+
+    return result;
 }
 
 export function entity<T = unknown>(
@@ -37,9 +50,15 @@ export function entity<T = unknown>(
     });
 }
 
-export function activity(functionName: string, options: ActivityOptions): void {
+export function activity(functionName: string, options: ActivityOptions): RegisteredActivity {
     azFuncApp.generic(functionName, {
         trigger: trigger.activity(),
         ...options,
     });
+
+    const result: RegisteredActivity = (input?: unknown): RegisteredActivityTask => {
+        return new RegisteredActivityTask(functionName, input);
+    };
+
+    return result;
 }
--- a/src/task/RegisteredActivityTask.ts
+++ b/src/task/RegisteredActivityTask.ts
@@ -1,0 +1,24 @@
+import * as types from "durable-functions";
+import { AtomicTask } from "./AtomicTask";
+import { RetryOptions } from "../RetryOptions";
+import { CallActivityAction } from "../actions/CallActivityAction";
+import { CallActivityWithRetryAction } from "../actions/CallActivityWithRetryAction";
+import { RetryableTask } from "./RetryableTask";
+
+export class RegisteredActivityTask extends AtomicTask implements types.RegisteredActivityTask {
+    withRetry: (retryOptions: RetryOptions) => RetryableTask;
+
+    constructor(activityName: string, input?: unknown) {
+        super(false, new CallActivityAction(activityName, input));
+
+        this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
+            const callActivityWithRetryAction = new CallActivityWithRetryAction(
+                activityName,
+                retryOptions,
+                input
+            );
+            const backingTask = new AtomicTask(false, callActivityWithRetryAction);
+            return new RetryableTask(backingTask, retryOptions);
+        };
+    }
+}
--- a/src/task/RegisteredActivityTask.ts
+++ b/src/task/RegisteredActivityTask.ts
@@ -14,9 +14,9 @@ export class RegisteredActivityTask extends AtomicTask implements types.Register
         this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
             if (this.alreadyScheduled) {
                 throw new Error(
-                    "Invalid use of `.withRetry`: attempted to create a retry task from an already scheduled task. " +
+                    "Invalid use of `.withRetry`: attempted to create a retriable task from an already scheduled task. " +
                         `A task with ID ${this.id} to call activity ${activityName} has already been scheduled. ` +
-                        "Make sure you only call `.withRetry` on tasks that have not previously been yielded."
+                        "Make sure to only invoke `.withRetry` on tasks that have not previously been yielded."
                 );
             }
 
--- a/src/task/RegisteredActivityTask.ts
+++ b/src/task/RegisteredActivityTask.ts
@@ -15,7 +15,7 @@ export class RegisteredActivityTask extends AtomicTask implements types.Register
             if (this.alreadyScheduled) {
                 throw new Error(
                     "Invalid use of `.withRetry`: attempted to create a retry task from an already scheduled task. " +
-                        `A task with ID ${this.id} to call activity ${activityName} has already been scheduled.` +
+                        `A task with ID ${this.id} to call activity ${activityName} has already been scheduled. ` +
                         "Make sure you only call `.withRetry` on tasks that have not previously been yielded."
                 );
             }
--- a/src/task/RegisteredActivityTask.ts
+++ b/src/task/RegisteredActivityTask.ts
@@ -12,6 +12,12 @@ export class RegisteredActivityTask extends AtomicTask implements types.Register
         super(false, new CallActivityAction(activityName, input));
 
         this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
+            if (this.alreadyScheduled) {
+                throw new Error(
+                    "This activity has already been scheduled. You cannot call withRetry on an already scheduled task. Create a new task instead."
+                );
+            }
+
             const callActivityWithRetryAction = new CallActivityWithRetryAction(
                 activityName,
                 retryOptions,
--- a/src/task/RegisteredActivityTask.ts
+++ b/src/task/RegisteredActivityTask.ts
@@ -14,7 +14,9 @@ export class RegisteredActivityTask extends AtomicTask implements types.Register
         this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
             if (this.alreadyScheduled) {
                 throw new Error(
-                    "This activity has already been scheduled. You cannot call withRetry on an already scheduled task. Create a new task instead."
+                    "Invalid use of `.withRetry`: attempted to create a retry task from an already scheduled task. " +
+                        `A task with ID ${this.id} to call activity ${activityName} has already been scheduled.` +
+                        "Make sure you only call `.withRetry` on tasks that have not previously been yielded."
                 );
             }
 
--- a/src/task/RegisteredOrchestrationTask.ts
+++ b/src/task/RegisteredOrchestrationTask.ts
@@ -1,0 +1,26 @@
+import * as types from "durable-functions";
+import { CallSubOrchestratorAction } from "../actions/CallSubOrchestratorAction";
+import { CallSubOrchestratorWithRetryAction } from "../actions/CallSubOrchestratorWithRetryAction";
+import { AtomicTask } from "./AtomicTask";
+import { RetryableTask } from "./RetryableTask";
+import { RetryOptions } from "../RetryOptions";
+
+export class RegisteredOrchestrationTask extends AtomicTask
+    implements types.RegisteredOrchestrationTask {
+    withRetry: (retryOptions: RetryOptions) => RetryableTask;
+
+    constructor(orchestrationName: string, input?: unknown, instanceId?: string) {
+        super(false, new CallSubOrchestratorAction(orchestrationName, instanceId, input));
+
+        this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
+            const callSubOrchestratorWithRetryAction = new CallSubOrchestratorWithRetryAction(
+                orchestrationName,
+                retryOptions,
+                input,
+                instanceId
+            );
+            const backingTask = new AtomicTask(false, callSubOrchestratorWithRetryAction);
+            return new RetryableTask(backingTask, retryOptions);
+        };
+    }
+}
--- a/src/task/RegisteredOrchestrationTask.ts
+++ b/src/task/RegisteredOrchestrationTask.ts
@@ -13,6 +13,12 @@ export class RegisteredOrchestrationTask extends AtomicTask
         super(false, new CallSubOrchestratorAction(orchestrationName, instanceId, input));
 
         this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
+            if (this.alreadyScheduled) {
+                throw new Error(
+                    "This subOrchestrator has already been scheduled. You cannot call withRetry on an already scheduled task. Create a new task instead."
+                );
+            }
+
             const callSubOrchestratorWithRetryAction = new CallSubOrchestratorWithRetryAction(
                 orchestrationName,
                 retryOptions,
--- a/src/task/RegisteredOrchestrationTask.ts
+++ b/src/task/RegisteredOrchestrationTask.ts
@@ -15,9 +15,9 @@ export class RegisteredOrchestrationTask extends AtomicTask
         this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
             if (this.alreadyScheduled) {
                 throw new Error(
-                    "Invalid use of `.withRetry`: attempted to create a retry task from an already scheduled task. " +
+                    "Invalid use of `.withRetry`: attempted to create a retriable task from an already scheduled task. " +
                         `A task with ID ${this.id} to call subOrchestrator ${orchestrationName} has already been scheduled. ` +
-                        "Make sure you only call `.withRetry` on tasks that have not previously been yielded."
+                        "Make sure to only invoke `.withRetry` on tasks that have not previously been yielded."
                 );
             }
 
--- a/src/task/RegisteredOrchestrationTask.ts
+++ b/src/task/RegisteredOrchestrationTask.ts
@@ -15,7 +15,9 @@ export class RegisteredOrchestrationTask extends AtomicTask
         this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
             if (this.alreadyScheduled) {
                 throw new Error(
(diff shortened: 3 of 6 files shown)


davidmrdavid:
Just to confirm, in this example:

```JavaScript
const flakyFunction: ActivityHandler = function (_input: any, context: InvocationContext): void {
    context.log("Flaky Function Flaking!");
    throw new Error("FlakyFunction flaked");
};
const flakyActivity: RegisteredActivity = df.app.activity("flakyFunction", {
    handler: flakyFunction,
});

const callActivityWithRetry: OrchestrationHandler = function* (context: OrchestrationContext) {
    const retryOptions: RetryOptions = new df.RetryOptions(1000, 2);

    let returnValue: any;
    try {
        returnValue = yield (flakyActivity().withRetry(retryOptions));
    } catch (e) {
        context.log("Orchestrator caught exception. Flaky function is extremely flaky.");
    }

    return returnValue;
};
df.app.orchestration("callActivityWithRetry", callActivityWithRetry);
```

Does the line

```JavaScript
        returnValue = yield (flakyActivity().withRetry(retryOptions));
```

need the parentheses around the activity task, or could we have?

```JavaScript
        returnValue = yield flakyActivity().withRetry(retryOptions);
```

I'm just trying to understand if there's an order of evaluation problem here, which would explain the parenthesis. Just checking!

src/app.ts
```diff
@@ -4,15 +4,19 @@ import {
     EntityOptions,
     OrchestrationHandler,
     OrchestrationOptions,
+    RegisteredActivity,
+    RegisteredOrchestration,
```
davidmrdavid, line 8: non-blocking nit: I'm beginning to get a bit concerned about the similar sounding names we've been introducing. We have an `orchestration`, and `orchestrationHandler`, and now a `RegisteredOrchestration`. I realize how they're all _subtly_ different, but that subtlety adds complexity. Just something to think about, not strictly something we need to/ can tackle in this PR.

types/orchestration.d.ts
```diff
@@ -51,6 +51,15 @@ export interface OrchestrationTrigger extends FunctionTrigger {
...
 }
 
+export type RegisteredOrchestration = (
+    input?: unknown,
+    instanceId?: string
+) => RegisteredOrchestrationTask;
+
+export interface RegisteredOrchestrationTask extends Task {
+    withRetry: (retryOptions: RetryOptions) => Task;
+}
```
davidmrdavid, line 61: What happens if we do the following pseudocode:

```
var retryOptions = // something
var task = myRegisteredOrchestration();
var result1 = yield task;
var result2 = yield myRegisteredOrchestration().withRetry(retryOptions);
```

I think the final line, where we call `.withRetry` is weird edge case: the task has already completed (it's been yielded), and afterwards we've decided to add `.withRetry` options and yield again. From my understanding, we'll create a brand new task as a result and schedule it when we yield it. Is that correct? I don't think the behavior is immediately obvious, as I could also imagine that, since `task` already completed, the second `yield` will just fetch the pre-existing result from the history.

davidmrdavid (commented):
Looks good, but I have a question about the behavior. Please see my comment on the withRetry edge case.

hossam-nasr:
@davidmrdavid No it is not. I had that there first just because I was debugging something and thought that the precedence order might be causing it, but it turns out it wasn't. I just double checked and the parentheses are not required. I also update the PR description now.

src/app.ts
hossam-nasr: Yeah I hear you. I went with `RegisteredActivity` and `RegisteredOrchestration` because those were the names that were agreed upon on a different but similar PR. If you have different suggestions for these names, please let me know!

types/orchestration.d.ts
hossam-nasr: Calling `myRegisteredOrchestration()` creates a new `Task` each time, so subsequent calls to `myRegisteredOrchestration()` should _not_ affect each other. A more "interesting" scenario might be:

```TS
var retryOptions = // something
var task = myRegisteredOrchestration();
var result1 = yield task;
var result2 = yield task.withRetry(retryOptions);
```

But this should still be safe, since `.withRetry()` also creates a new `RetryableTask` each time, so the fact that `task` was already yielded (or even completed) doesn't make a difference. To double check, I tested the following orchestration:

```TS
df.app.orchestration("testScenario", function* (context: OrchestrationContext) {
    const retryOptions: RetryOptions = new df.RetryOptions(1000, 2);
    const task = helloActivity("David");
    const result1 = yield task;
    yield context.df.waitForExternalEvent("testEvent");
    const result2 = yield task.withRetry(retryOptions);
    return [result1, result2];
});
```

And waited to send the `testEvent` signal until after the first Hello Activity is run. And we see the result that we expect:

```JSON
{
    "name": "testScenario",
    "instanceId": "9257ed4396714de8978b62388e1992cc",
    "runtimeStatus": "Completed",
    "input": "",
    "customStatus": null,
    "output": [
        "Hello David",
        "Hello David"
    ],
    "createdTime": "2023-08-01T17:38:46Z",
    "lastUpdatedTime": "2023-08-01T17:40:17Z"
}
```

In the logs, I could also see that `sayHello` activity was called twice, not just the result from the first call reused.

src/app.ts
davidmrdavid: Not at this time, I like these names at this time because they're simple, but at some point I think we should take a global look at the names we've been introducing to see if we need to document their differences somewhere.

hossam-nasr: Sounds good :)

types/orchestration.d.ts
hossam-nasr: Discussed offline. Changing this behavior to throw an error instead, since it is counterintuitive to spawn a new task from an already scheduled task. Implemented the change so that this throws an error. The same scenario now fails with this exception message:

```JSON
{
    "name": "testScenario",
    "instanceId": "9fd3813184b74a6c8bdd5efd278a510a",
    "runtimeStatus": "Failed",
    "input": "",
    "customStatus": null,
    "output": "Orchestrator function 'testScenario' failed: This activity has already been scheduled. You cannot call withRetry on an already scheduled task. Create a new task instead.",
    "createdTime": "2023-08-01T19:22:25Z",
    "lastUpdatedTime": "2023-08-01T19:23:16Z"
}
```

Thankfully, the `alreadyScheduled` flag we added a while back made this change quite straightforward :)

src/app.ts
```diff
@@ -21,6 +25,15 @@ export function orchestration(
...
+        input?: unknown,
+        instanceId?: string
+    ): RegisteredOrchestrationTask => {
+        return new RegisteredOrchestrationTask(functionName, input, instanceId);
+    };
+
+    return result;
 }
 
 export function entity<T = unknown>(
```
davidmrdavid, line 39: Any thoughts on how this improvement would look like for entities?

src/task/RegisteredActivityTask.ts
```diff
@@ -0,0 +1,30 @@
...
+export class RegisteredActivityTask extends AtomicTask implements types.RegisteredActivityTask {
+    withRetry: (retryOptions: RetryOptions) => RetryableTask;
+
+    constructor(activityName: string, input?: unknown) {
+        super(false, new CallActivityAction(activityName, input));
+
+        this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
+            if (this.alreadyScheduled) {
+                throw new Error(
+                    "This activity has already been scheduled. You cannot call withRetry on an already scheduled task. Create a new task instead."
```
davidmrdavid: nitpicky nit -> How about something like this:

```suggestion
                    "Invalid use of `.withRetry`: attempted to add retry options to an already scheduled task. The affected task has ID={this.taskScheduledId} and name={this.activityName}. Make sure you only call `.withRetry` on tasks that have not previously been yielded.
```

Note, I don't think this suggestion compiles, but I think you get the idea of what I'm trying to do
(1) Summarize the error type: "Invalid use of .withRetry"
(2) Provide enough identifying information for the customer to find the culprit line of code: printing taskScheduledId and the activityName
(3) The fix: only call withRetry on unyielded tasks.

davidmrdavid: If you're good with this error message format, please also apply it to the orchestration case.

davidmrdavid (commented):
Almost ready to approve, just a few more nitpicks

src/app.ts
hossam-nasr: I do have thoughts, but it's not as straightforward as the activity and suborchestrator case, so I'm saving it for a different PR/discussion 😛

src/task/RegisteredActivityTask.ts
hossam-nasr: Sounds good to me. I reworded it a bit and added the right variables, but I think I got the gist of it. Let me know if the new error message works

src/task/RegisteredActivityTask.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    constructor(activityName: string, input?: unknown) {
+        super(false, new CallActivityAction(activityName, input));
+
+        this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
+            if (this.alreadyScheduled) {
+                throw new Error(
+                    "Invalid use of `.withRetry`: attempted to create a retry task from an already scheduled task. " +
+                        `A task with ID ${this.id} to call activity ${activityName} has already been scheduled. ` +
+                        "Make sure you only call `.withRetry` on tasks that have not previously been yielded."
```
davidmrdavid:
```suggestion
                    "Invalid use of `.withRetry`: attempted to create a retriable task from an already scheduled task. " +
                        `A task with ID ${this.id} to call activity ${activityName} has already been scheduled. ` +
                        "Make sure you only invoke `.withRetry` on tasks that have not previously been yielded."
```

src/task/RegisteredOrchestrationTask.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    constructor(orchestrationName: string, input?: unknown, instanceId?: string) {
+        super(false, new CallSubOrchestratorAction(orchestrationName, instanceId, input));
+
+        this.withRetry = (retryOptions: RetryOptions): RetryableTask => {
+            if (this.alreadyScheduled) {
+                throw new Error(
+                    "Invalid use of `.withRetry`: attempted to create a retry task from an already scheduled task. " +
+                        `A task with ID ${this.id} to call subOrchestrator ${orchestrationName} has already been scheduled. ` +
+                        "Make sure you only call `.withRetry` on tasks that have not previously been yielded."
```
davidmrdavid:
```suggestion
                    "Invalid use of `.withRetry`: attempted to create a retriable task from an already scheduled task. " +
                        `A task with ID ${this.id} to call subOrchestrator ${orchestrationName} has already been scheduled. ` +
                        "Make sure to only invoke `.withRetry` on tasks that have not previously been yielded."
```

davidmrdavid (approved):
Approved with suggestions. Excited to see this change. Let's please make sure to prioritize and track a similar change to Entities

src/app.ts
```diff
@@ -37,9 +50,15 @@ export function entity<T = unknown>(
...
 }
 
-export function activity(functionName: string, options: ActivityOptions): void {
+export function activity(functionName: string, options: ActivityOptions): RegisteredActivity {
     azFuncApp.generic(functionName, {
         trigger: trigger.activity(),
         ...options,
     });
+
+    const result: RegisteredActivity = (input?: unknown): RegisteredActivityTask => {
```
ejizba, line 59: I think this has been suggested before, but rather than this:
```
yield mySubOrchestrator(input);
yield mySubOrchestrator(input).withRetry(retryOptions);
```

What do we think of this?
```
yield mySubOrchestrator.call(input);
yield mySubOrchestrator.callWithRetry(input);
```

hossam-nasr: I believe this was one of the proposals that we discussed, and we agreed that the former was cleaner. Part of the reason is, for the second option, the `callWithRetry()` would actually need to look something like:

```TS
yield mySubOrchestrator.callWithRetry({ input: 'someInput', retryOptions });
```

Otherwise, we would end up with the null/undefined problem, in case a user wants to call an activity/subOrchestrator with retry, but with no input:

```TS
yield mySubOrchestrator.callWithRetry(null, retryOptions);
```

If we go with an options object for `.callWithRetry()`, that would also beg the question of whether we should also have an options object for `.call()`, for consistency's sake

ejizba: I think I was assuming the `retryOptions` argument was before `input`. Correct me if I'm wrong, but I think that's what we're [doing today](https://github.com/Azure/azure-functions-durable-js/blob/v3.x/types/orchestration.d.ts#L191). So in that case you'd never be forced to pass `null`

> If we go with an options object for .callWithRetry(), that would also beg the question of whether we should also have an options object for .call(), for consistency's sake

You bring up an interesting point. Won't we need an options object for `call` anyways? I believe this is what Jacob [said he was doing](https://github.com/Azure/azure-functions-durable-js/issues/418#issuecomment-1405387203) for C#, in part for things like `instanceId` and `fireAndForget` as well as the retry stuff. If we do an options object, we debatably don't need any `WithRetry` methods at all

hossam-nasr: > I think I was assuming the retryOptions argument was before input. Correct me if I'm wrong, but I think that's what we're [doing today](https://github.com/Azure/azure-functions-durable-js/blob/v3.x/types/orchestration.d.ts#L191). So in that case you'd never be forced to pass null

Yes, that is a good point. 

> Won't we need an options object for call anyways?

Given where the node SDK is now, I'm not convinced we would need to add a fire-and-forget operation to calling activities and sub-orchestrators any time soon. I would still also prefer to have to explicitly state that you'd like to call an activity/suborchestrator with retry, rather than that being implied by an options object parameter. I remember us having similar conversations too around changing `context.df.callActivity()` and `context.df.callSubOrchestrator()` to accepting options object, and that was the same conclusion

Would like to hear @davidmrdavid's thoughts here too

davidmrdavid: > Given where the node SDK is now, I'm not convinced we would need to add a fire-and-forget operation to calling activities and sub-orchestrators any time soon.

There are no fire-and-forget activities but the DF extension does support fire-and-forget sub-orchestrators. But that has not been a top ask for us at this time. Still, worth noting.

>  I remember us having similar conversations too around changing context.df.callActivity() and context.df.callSubOrchestrator() to accepting options object, and that was the same conclusion

In general, I think @hossam-nasr's design in this PR is giving the input to activities and sub-orchestrators special treatment so that:
(1) if a user has an input but not retryOptions, they don't have to construct a whole options object for it. I think this is a common scenario worth optimizing for.
(2) if a user has no input but they have `retryOptions`, they don't have to pass in a `null` input, which looks a bit clunky.

I like this design because I think inputs need to be handled separately from retry configuration as well as other knobs, as the use of the former is more common than the latter _afaik._

If we wanted to tweak the current design to allow for fire-and-forget sub-orchestrators, I think we could make the API look like this:

ejizba (approved):
As discussed offline, no major objections to this design and I think we've already spent enough time considering alternatives.

I still consider @davidmrdavid the primary approver on this PR, so as long as he's still okay with it, I am too

davidmrdavid (approved):
LGTM. Thanks to all who gave feedback for considering all the edge cases. I feel comfortable in this decision, and I appreciate the attention to detail here.

PR #536, opened 2023-08-01
Add convenience methods to register durable client functions

### LINKED ISSUE #414: Improve the experience of getting the Durable client in the new prog model
This issue is based off discussions [here](https://github.com/ejizba/func-nodejs-prototype/pull/14#discussion_r1018550932), [here](https://github.com/ejizba/func-nodejs-prototype/pull/14/files#r1044705193), and [here](https://github.com/ejizba/func-nodejs-prototype/pull/14#discussion_r1014348658).

In the current alpha release of the SDK (`3.0.0-alpha.1`) to support durable functions in the new programming model, this is the best experience for a user to register an HTTP-triggered durable client starter function:

```TS
const clientInput = df.input.durableClient();

const httpStart: HttpHandler = async (request: HttpRequest, context: InvocationContext) => {
    const client = df.getClient(context, clientInput);
    const instanceId = await client.startNew(request.params.orchestratorName, undefined, request.text());
    return client.createCheckStatusResponse(request, instanceId);
};

app.http('durableOrchestrationStart1', {
    route: 'orchestrators/{orchestratorName}',
    extraInputs: [clientInput],
    handler: httpStart,
});
```

As seen above, this requires the user to explicitly declare a durable client input binding, specify it in the `extraInputs` parameter when registering the function, and pass it to the `df.getClient()` method. This is clearly not an ideal user experience, and arguably a downgrade from the old model experience.

One suggestion/alternative would be to have the SDK provide an API that registers a "durable client handler," which takes in 3 arguments instead of the usual two (the trigger input, the context, and the durable client), while the SDK handles the input binding behind the scenes and provides the durable client directly to the handler. Such user code could look like this:

```TS
const clientHandler: DurableClientHandler = async (context: InvocationContext, request: HttpRequest, client) => {
    const instanceId = await client.startNew(request.params.orchestratorName, undefined, request.text());
    return client.createCheckStatusResponse(request, instanceId);
};

df.app.client('durableOrchestrationStart1', {
    trigger: trigger.http({ route: 'orchestrators/{orchestratorName}' }),
    return: output.http({}),
    handler: clientHandler,
});
```

However, per the discussion [here](https://github.com/ejizba/func-nodejs-prototype/pull/14/commits/df691cb359fcf160c17c4631ca6ce217144c4e5c#r1018550932), this "appears to come at a tradeoff with the trigger (you lose all the intellisense and defaults of doing `app.http`)".

Another alternative would be to provide some equivalent to `app.http()`, specifically for http-triggered durable client functions. As an example:

```TS
const clientHandler: DurableClientHandler = async (context: InvocationContext, request: HttpRequest, client) => {
    const instanceId = await client.startNew(request.query.get('orchestratorName'), undefined, request.text());
    return client.createCheckStatusResponse(request, instanceId);
};

df.app.httpClient('durableOrchestrationStart1', clientHandler);
```

However, since client functions could be triggered by any trigger, it begs the question of should we provide similar APIs for every possible trigger, which would be a lot of burden on the SDK.

This issue is to discuss alternatives to build upon or improve the current experience.

hossam-nasr:
Partially resolves #414. This PR adds a namespace `client`, under `df.app`, which provides methods to quickly register durable client functions. These functions receive three arguments instead of the regular two: the trigger, the durable client, and the `InvocationContext`, in this order. This allows users to quickly and easily register durable client functions, without having to manually configure the `durableClient` input binding and calling the `getClient` method. This PR starts out with three methods: `http` to register HTTP-triggered durable client functions, `timer`, and `generic`. These behave exactly the same as the equivalent from `@azure/functions`, except accepting the additional `DurableClient` argument.

Example usage:

### HTTP-triggered client function

```TS
const httpStart: HttpDurableClientHandler = async (
    request: HttpRequest,
    client: DurableClient,
    context: InvocationContext
): Promise<HttpResponse> => {
    const body: unknown = await request.json();
    const instanceId: string = await client.startNew(request.params.orchestratorName, {
        input: body,
    });

    return client.createCheckStatusResponse(request, instanceId);
};

df.app.client.http("httpStart", {
    route: "orchestrators/{orchestratorName}",
    handler: httpStart,
});
```

### Timer-triggered client function

```TS
const timerStart: TimerDurableClientHandler = async (
    _timer: Timer,
    client: DurableClient,
    context: InvocationContext
): Promise<void> => {
    const instanceId: string = await client.startNew("helloSequence");
};

df.app.client.timer("timerStart", {
    schedule: "0 */1 * * * *",
    handler: timerStart,
});
```

### Storage Queue-triggered client function (using generic)

```TS
const storageQueueStart: DurableClientHandler = async (
    message: string,
    client: DurableClient,
    context: InvocationContext
): Promise<void> => {
    const instanceId: string = await client.startNew(message);
};

df.app.client.generic("storageQueueStart", {
    trigger: trigger.storageQueue({ queueName: "orchestrations", connection: "StorageConnString" }),
    handler: storageQueueStart,  
})
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -7,6 +7,7 @@ import {
 } from "durable-functions";
 import {
     FunctionHandler,
+    FunctionInput,
     FunctionResult,
     InvocationContext,
     app as azFuncApp,
@@ -39,11 +40,15 @@ export function generic(functionName: string, options: DurableClientOptions): vo
 }
 
 function addClientInput(options: Partial<DurableClientOptions>): void {
-    if (options.extraInputs) {
+    if (options.extraInputs && !options.extraInputs.find(isDurableClientInput)) {
         options.extraInputs.push(input.durableClient());
     } else {
         options.extraInputs = [input.durableClient()];
     }
+}
+
+function isDurableClientInput(inputOptions: FunctionInput): boolean {
+    return inputOptions.type === "durableClient" || inputOptions.type === "orchestrationClient";
 }
 
 function convertToFunctionHandler(clientHandler: DurableClientHandler): FunctionHandler {
--- a/src/client.ts
+++ b/src/client.ts
@@ -7,13 +7,12 @@ import {
 } from "durable-functions";
 import {
     FunctionHandler,
-    FunctionInput,
     FunctionResult,
     InvocationContext,
     app as azFuncApp,
 } from "@azure/functions";
 import { DurableClient } from "./durableClient/DurableClient";
-import { getClient } from "./durableClient/getClient";
+import { getClient, isDurableClientInput } from "./durableClient/getClient";
 
 export function http(functionName: string, options: HttpDurableClientOptions): void {
     addClientInput(options);
@@ -45,10 +44,6 @@ function addClientInput(options: Partial<DurableClientOptions>): void {
     } else {
         options.extraInputs = [input.durableClient()];
     }
-}
-
-function isDurableClientInput(inputOptions: FunctionInput): boolean {
-    return inputOptions.type === "durableClient" || inputOptions.type === "orchestrationClient";
 }
 
 function convertToFunctionHandler(clientHandler: DurableClientHandler): FunctionHandler {
--- a/src/client.ts
+++ b/src/client.ts
@@ -39,10 +39,9 @@ export function generic(functionName: string, options: DurableClientOptions): vo
 }
 
 function addClientInput(options: Partial<DurableClientOptions>): void {
-    if (options.extraInputs && !options.extraInputs.find(isDurableClientInput)) {
+    options.extraInputs = options.extraInputs ?? [];
+    if (!options.extraInputs.find(isDurableClientInput)) {
         options.extraInputs.push(input.durableClient());
-    } else {
-        options.extraInputs = [input.durableClient()];
     }
 }
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,0 +1,54 @@
+import * as input from "./input";
+import {
+    DurableClientHandler,
+    DurableClientOptions,
+    HttpDurableClientOptions,
+    TimerDurableClientOptions,
+} from "durable-functions";
+import {
+    FunctionHandler,
+    FunctionResult,
+    InvocationContext,
+    app as azFuncApp,
+} from "@azure/functions";
+import { DurableClient } from "./durableClient/DurableClient";
+import { getClient } from "./durableClient/getClient";
+
+export function http(functionName: string, options: HttpDurableClientOptions): void {
+    addClientInput(options);
+    azFuncApp.http(functionName, {
+        ...options,
+        handler: convertToFunctionHandler(options.handler),
+    });
+}
+
+export function timer(functionName: string, options: TimerDurableClientOptions): void {
+    addClientInput(options);
+    azFuncApp.timer(functionName, {
+        ...options,
+        handler: convertToFunctionHandler(options.handler),
+    });
+}
+
+export function generic(functionName: string, options: DurableClientOptions): void {
+    addClientInput(options);
+    azFuncApp.generic(functionName, {
+        ...options,
+        handler: convertToFunctionHandler(options.handler),
+    });
+}
+
+function addClientInput(options: Partial<DurableClientOptions>): void {
+    if (options.extraInputs) {
+        options.extraInputs.push(input.durableClient());
+    } else {
+        options.extraInputs = [input.durableClient()];
+    }
+}
+
+function convertToFunctionHandler(clientHandler: DurableClientHandler): FunctionHandler {
+    return (trigger: unknown, context: InvocationContext): FunctionResult<any> => {
+        const client: DurableClient = getClient(context);
+        return clientHandler(trigger, client, context);
+    };
+}
--- a/types/app.d.ts
+++ b/types/app.d.ts
@@ -55,3 +55,5 @@ export function entity<T = unknown>(functionName: string, options: EntityOptions
  * @param options the configuration options for this activity, specifying the handler and the inputs and outputs
  */
 export function activity(functionName: string, options: ActivityOptions): RegisteredActivity;
+
+export * as client from "./app.client";
--- a/types/durableClient.d.ts
+++ b/types/durableClient.d.ts
@@ -1,9 +1,63 @@
-import { FunctionInput, HttpRequest, HttpResponse } from "@azure/functions";
+import {
+    FunctionInput,
+    FunctionOptions,
+    FunctionResult,
+    HttpFunctionOptions,
+    HttpRequest,
+    HttpResponse,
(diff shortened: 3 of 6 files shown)


src/client.ts
```diff
@@ -0,0 +1,54 @@
...
+export function generic(functionName: string, options: DurableClientOptions): void {
+    addClientInput(options);
+    azFuncApp.generic(functionName, {
+        ...options,
+        handler: convertToFunctionHandler(options.handler),
+    });
+}
+
+function addClientInput(options: Partial<DurableClientOptions>): void {
+    if (options.extraInputs) {
```
ejizba: I wonder if we should do any sort of special logic here to see if they already have a client input. I could see people doing that on accident

types/durableClient.d.ts
```diff
@@ -1,11 +1,65 @@
...
+    request: HttpRequest,
+    client: DurableClient,
+    context: InvocationContext
+) => FunctionResult<HttpResponseInit | HttpResponse>;
+
+/**
+ * Configures options for an HTTP-triggered Durable Client function.
+ */
+export interface HttpDurableClientOptions extends Omit<HttpFunctionOptions, "handler"> {
+    handler: DurableClientHandler;
```
ejizba: I want to make sure people still get intellisense for the `request` trigger input in JavaScript (where people don't have the types). I think we need this to make it happen, but in general can you double check intellisense in JavaScript with your changes?:

```suggestion
    handler: HttpDurableClientHandler;
```

types/durableClient.d.ts
```diff
@@ -1,11 +1,65 @@
...
+    myTimer: Timer,
+    client: DurableClient,
+    context: InvocationContext
+) => FunctionResult;
+
+/**
+ * Configures options for a timer-triggered Durable Client function.
+ */
+export interface TimerDurableClientOptions extends Omit<TimerFunctionOptions, "handler"> {
+    handler: DurableClientHandler;
```
ejizba:
```suggestion
    handler: TimerDurableClientHandler;
```

types/app.d.ts
```diff
@@ -45,3 +45,5 @@ export function entity<T = unknown>(functionName: string, options: EntityOptions
  * @param options the configuration options for this activity, specifying the handler and the inputs and outputs
  */
 export function activity(functionName: string, options: ActivityOptions): void;
+
+export * as client from "./app.client";
```
ejizba, line 59: I see a lot of possible organizations here. I'm not sure I have a strong preference, but have we considered all of these?

- df.app.client.http
- df.app.http
- df.app.httpClient
- df.app.httpWithClient

src/client.ts
hossam-nasr: We could do that for cleanliness sake, but my understanding is that it shouldn't make any difference. A durable client input binding is singleton, and there is no information specific to the input binding that would be required by `getClient()`, so having two would just be redundant. Correct me if I'm wrong @davidmrdavid

hossam-nasr: I made the change anyway

types/app.d.ts
hossam-nasr: Yeah these are all valid options. I'm not sure we had extensive discussion on how exactly to name/organize these. I just went with one suggestion here, but I also have no strong preferences.

ejizba: Perhaps `df.app.client.http` doesn't feel right to me because I'm used to the templates calling these starter functions. Even in your example above, the function name is "httpStart" or "timerStart" without "client" in the name. Should we do something like `df.app.start.http` or `df.app.httpStart` instead? Then again I see the docs categorize these as [client functions](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-types-features-overview#client-functions) so I'm a little confused here.

I don't have a strong preference, but I don't think this should be an arbitrary decision. All I ask is that you give a reason for your choice. Intellisense is a priority for the new model and the organization here directly impacts that experience.

hossam-nasr: Okay, makes sense. Here are a few reasons why I made this decision:

* I think client functions should be under their own namespace. I don't want a bunch of client functions polluting the intellisense for `df.app`, especially if we end up adding similar methods for all possible triggers, when at the end of the day, the "meat" of the durable app is in your orchestrations, activities, entities.
* As you said, the docs consistently refer to these functions as client functions. Sometimes we also call them client starter functions, but I think client function is a more ubiquitous name. I could also see this being called `df.app.starter` for example, but I think `client` is the more "official" name for them.
* We already have precedent for this name. To register a client input binding, we call `df.input.durableClient()`. In your client function, you call `df.getClient()` and you get a `DurableClient` instance. It didn't feel like a stretch to call these functions a `DurableClientFunction`, and by extension, to register them using `df.app.client`. It was a narrow decision for me to choose `client` over `durableClient`, but I eventually went for `client` because it felt like `df.app.durableClient.http()` would be a bit too lengthy, and if you're using `df.app`, it's already assumed to be a durable client anyway. I could also easily see this being `df.app.durableClient` though.

In conclusion, I think this should be a separate namespace. `starter` or `durableClient` are other proposals for what to call that namespace, that I don't have strong feelings against, but I went with `client` for the reasons mentioned above.

src/client.ts
```diff
@@ -0,0 +1,59 @@
...
+export function generic(functionName: string, options: DurableClientOptions): void {
+    addClientInput(options);
+    azFuncApp.generic(functionName, {
+        ...options,
+        handler: convertToFunctionHandler(options.handler),
+    });
+}
+
+function addClientInput(options: Partial<DurableClientOptions>): void {
+    if (options.extraInputs && !options.extraInputs.find(isDurableClientInput)) {
```
ejizba: This isn't quite right. If I have `[input.durableClient(), input.storageBlob()]`, I think it'll fall through to the else case and overwrite with just `[input.durableClient()]`

Here's how I'd do it:
```typescript
options.extraInputs = options.extraInputs ?? [];
if (!options.extraInputs.find(isDurableClientInput)) {
    options.extraInputs.push(input.durableClient());
}
```

src/client.ts
```diff
@@ -0,0 +1,59 @@
...
+        options.extraInputs.push(input.durableClient());
+    } else {
+        options.extraInputs = [input.durableClient()];
+    }
+}
+
+function isDurableClientInput(inputOptions: FunctionInput): boolean {
```
ejizba: nit: just use `isDurableClientInput` from `getClient.ts` instead of copying it here

ejizba (approved):
Fix the one bug, but otherwise LGTM

davidmrdavid (approved):

src/client.ts
hossam-nasr: Why would `[input.durableClient(), input.storageBlob()]` fall through to the else case? Wouldn't `!options.extraInputs.find(isDurableClientInput)` equal `false`?

hossam-nasr: Brainfart moment, fixed

PR #548, opened 2023-09-25
Update version to `v3.0.0-alpha.6`

hossam-nasr:
As in title, routine PR. Worth making a new release with the new syntactic improvements so I can reference them and update the docs, samples, templates, etc., in time for GA


hossam-nasr:
Closed in favor of #549

Issue #194, opened 2020-06-05
Update MODULE_VERSION in azure-pipelines.yml


davidmrdavid:
Inspired from the outstanding issue in: https://github.com/Azure/azure-functions-durable-js/pull/192

We have not been updating the MODULE_VERSION variable in `azure-pipelines.yml`, even though it's getting updated in the `package.json`. We should use variable groups, as  suggested by @aaronpowell, to fetch that variable as updated in the `package.json`.

This script should do the trick: `node -e 'console.log(require("./package.json").version)'`

We just need to embed it into the pipelines yaml.

davidmrdavid:
@hossam-nasr, I was doing some backlog cleaning and stumbled upon this old work item. This is by no means urgent, but I want to bring it to your attention.

Essentially, this "https://github.com/Azure/azure-functions-durable-js/blob/a572eacd407e6796d82a4006d4f87e07674940e0/azure-pipelines.yml#L2-L4" field is not getting updated to represent the actual version of our package. I don't think it really affects anything other than our CI reports, but it seems easy enough to fix.

Again, not urgent. Just putting this on your radar.

hossam-nasr:
@davidmrdavid Thanks for pointing this out! Perhaps adding a script to update/validate this automatically could avoid this kind of issue in the future, for example like @ejizba did for the node worker: https://github.com/Azure/azure-functions-nodejs-worker/pull/563

ejizba:
I think the script would be a bit overkill for this use case. Personally, I would just remove "MODULE_VERSION" and "name" from the azure pipelines yaml and be done with it. All it does is change the name in dev ops:
![Screen Shot 2022-05-25 at 3 16 50 PM](https://user-images.githubusercontent.com/11282622/170377515-52fad917-b08b-480d-b750-ed7cc4cc1c6a.png)

But the version is already listed in the artifact, which IMO is the more important place:
![Screen Shot 2022-05-25 at 3 17 10 PM](https://user-images.githubusercontent.com/11282622/170377613-f57f26be-2f16-45f8-bf78-01c7d9220597.png)

davidmrdavid:
You're right. This is probably overkill, and the usage of `MODULE_VERSION` predates me so I don't feel strongly about keeping it. 

_Closing this issue!_

PR #580, opened 2024-02-06
Complete updating v2.x to 2.1.4

castrodd:
The [release pipeline](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=161001&view=results) did not successfully complete with the previous PR: https://github.com/Azure/azure-functions-durable-js/pull/576.

By comparing the above PR with the [last PR to update the v2.x branch](https://github.com/Azure/azure-functions-durable-js/pull/522/files), it was clear that we forgot to update the `package-lock.json` and `azure-pipelines.yml`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,6 @@
 variables:
     {
-        MODULE_VERSION: "2.1.3",
+        MODULE_VERSION: "2.1.4",
         NODE_12: "12.x",
         NODE_14: "14.x",
         NODE_16: "16.x",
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -1,6 +1,6 @@
 variables:
     {
-        MODULE_VERSION: "2.1.3",
+        MODULE_VERSION: "2.1.4",
```
ejizba: Fyi, my personal preference is to remove MODULE_VERSION (and "name" from line 10) as stated here:

https://github.com/Azure/azure-functions-durable-js/issues/194#issuecomment-1137898533

ejizba (approved):

azure-pipelines.yml
castrodd: I prefer that as well. I removed both.

PR #592, opened 2024-03-18
Bump packages in sample apps

ejizba:
Primarily to resolve annoying axios alerts


davidmrdavid (approved):
Thanks for the maintenance work here!

PR #601, opened 2024-05-28
Add mirror pipeline for v3.x branch

castrodd:
This PR is part of the migration to 1ES Pipeline Templates.

### DIFF
--- a/code-mirror.yml
+++ b/code-mirror.yml
@@ -1,0 +1,19 @@
+trigger:
+    branches:
+        include:
+            - master
+            - v2.x
+            - v3.x
+
+resources:
+    repositories:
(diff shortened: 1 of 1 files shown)


code-mirror.yml
```diff
@@ -0,0 +1,19 @@
+trigger:
+    branches:
+        include:
+            - master
```
ejizba: Can we either remove master (doesn't seem to have been updated in 4 years) or rename it to v1.x (if that's why we're including it)

castrodd: Removed `master`

ejizba (approved):

davidmrdavid (approved):

PR #605, opened 2024-06-07
Update release pipeline

castrodd:
This PR updates the release pipeline to point to the new 1ES PT build pipeline. It also moves all the pipelines to the same directory location and deletes the old (non-1ES PT) build pipeline.


azure-pipelines/azure-pipelines-release.yml
```diff
@@ -14,8 +14,8 @@ pr: none
 resources:
```
ejizba: nit: Following other files in this folder, this should probably be named "release.yml" or "official-release.yml"

ejizba (approved):

azure-pipelines/azure-pipelines-release.yml
castrodd: Fixed!

PR #623, opened 2025-01-13
Replace lodash cloneDeep with structuredClone

wojtekmaj:
structuredClone is available in Node.js since version 17.0.0. As this package supports Node.js 18.0.0 and up, we're able to drop production dependency on lodash, making the package footprint smaller and install times faster.


wojtekmaj:
@torosent What was wrong?

PR #632, opened 2025-04-09
Add support to send trace information to the WebJobs extension

bachuv:
This PR adds support to send `traceparent` and `tracestate` information to the WebJobs extension. I added two new headers to the requestsent to the WebJobs extension , `traceparent` and `tracestate`, which are only added if there is an active "current span".

For the sample code below, I followed the instructions in https://learn.microsoft.com/en-us/azure/azure-functions/opentelemetry-howto?tabs=app-insights&pivots=programming-language-javascript.

Example HTTP trigger code

``` JavaScript
const { app } = require('@azure/functions');
﻿const df = require("durable-functions");
const { trace, propagation, context: otelContext, ROOT_CONTEXT, SpanKind } = require('@opentelemetry/api');

...

app.http('durableHelloHttpStart', {
    route: 'orchestrators/{orchestratorName}',
    extraInputs: [df.input.durableClient()],
    handler: async (request, context) => {
        // Create a new span that carries the incoming trace data.
        const traceparent = context.traceContext.traceParent;
        const tracestate = context.traceContext.traceState;
        const carrier = { traceparent, tracestate };

        // Extract the parent context
        // const parentContext = propagation.extract(otelContext.active(), carrier);
        const parentContext = propagation.extract(ROOT_CONTEXT, carrier);

        // Create and return a new span using the extracted context.
        const tracer = trace.getTracer('durableHelloHttpStart');
        const span = tracer.startSpan('myCustomSpan', {
            kind: SpanKind.CLIENT,
        }, parentContext);

        // Mark the span as active
        const activeCtx = trace.setSpan(otelContext.active(), span);
        return await otelContext.with(activeCtx, async () => {
            // Now any spans created inside this callback will automatically use "span"
            // as their parent.
            
            // Read the request body.
            const body = await request.text();
            const client = df.getClient(context);

            const instanceId = await client.startNew(request.params.orchestratorName, {
                input: body
            });
            
            context.log(`Started orchestration with ID = '${instanceId}'.`);

            span.end();

            return client.createCheckStatusResponse(request, instanceId);
        });
    },
});
```

src/index.js
``` JavaScript
const { app } = require('@azure/functions');
const { AzureFunctionsInstrumentation } = require('@azure/functions-opentelemetry-instrumentation');
const { AzureMonitorLogExporter, AzureMonitorTraceExporter } = require('@azure/monitor-opentelemetry-exporter');
[...]

PR #639, opened 2025-05-15
Distributed Tracing for Entities

sophiatev:
This PR adds additional information passing to enable distributed tracing for entities.

- `DurableClient.signalEntity` was modified to pass the current trace context in its HTTP headers (similar to how is done in the `startNew` call) such that the trace for signaling the entity created in WebJobs can correctly be linked as a child of the client trace that triggered the signal entity request.
- The start time for processing an entity invocation is also passed in the `OperationResult` such that the traces can accurately capture the time it takes for the request to be completed (using the `OperationResult.duration` field to calculate the end time).
- In the case of entity A signaling entity B, the id of the `RequestMessage` that led to entity A's invocation is now attached to the `Signal` object, as well as the `requestTime` of the signal request. The former is done so that the span for the signal can be correctly correlated as the child of the span for entity A's invocation (see further discussion in the WebJobs [PR](https://github.com/Azure/azure-functions-durable-extension/pull/3066)). 

Note that for a client signaling an entity via `DurableClient.signalEntity`, in order for this method to be able to obtain the current trace context, the user needs to actually start their own trace from their function app. An example of all the code changes necessary is provided in this [PR ](https://github.com/Azure/azure-functions-durable-js/pull/632) for a client starting an orchestration (the example code would be identical in this case, with `client.startNew` replaced by `client.signalEntity`):

### Client Signaling an Entity
```javascript
app.http('counterHttpAdd', {
    route: `add${entityName}/{id}`,
    extraInputs: [df.input.durableClient()],
    handler: async (req, context) => {
        // Create a new span that carries the incoming trace data.
        const traceparent = context.traceContext.traceParent;
        const tracestate = context.traceContext.traceState;
        const carrier = { traceparent: traceparent, tracestate: tracestate };

        // Extract the parent context
        const parentContext = propagation.extract(otelContext.active(), carrier);
        //const parentContext = propagation.extract(ROOT_CONTEXT, carrier);

        // Create and return a new span using the extracted context.
        const tracer = trace.getTracer('counterHttpAdd');
        const span = tracer.startSpan('myCustomSpan', {
            kind: SpanKind.CLIENT,
        }, parentContext);

        // Mark the span as active
        const activeCtx = trace.setSpan(otelContext.active(), span);
        return await otelContext.with(activeCtx, async () => {
            // Now any spans created inside this callback will automatically use "span"
            // as their parent.
            
            const id = req.params.id;
            const entityId = new df.EntityId(entityName, id);
    
            // increment value
            await client.signalEntity(entityId, 'add', 1);

            span.end();

            return new HttpResponse({
                jsonBody: 'done',
            });
        });
    },
});
```
where the counter entity looks like

```javascript
df.app.entity(entityName, (context) => {
    const currentValue = context.df.getState(() => 0);
    switch (context.df.operationName) {
        case 'add':
            const amount = context.df.getInput();
            const entityId = new df.EntityId("monitor", "myMonitor");
            context.df.signalEntity(entityId, "milestone-reached", context.df.instanceId);
            context.df.setState(currentValue + amount);
            break;
        case 'reset':
            context.df.setState(0);
            break;
        case 'get':
            context.df.return(currentValue);
            break;
    }
});
```

leads to the following trace, with the counter entity signaling the monitor entity in the `add` operation
image

### Orchestration Signaling an Entity
Whereas starting the following orchestration 
```javascript
app.http('durableHelloHttpStart', {
    route: 'orchestrators/{orchestratorName}',
    extraInputs: [df.input.durableClient()],
    handler: async (request, context) => {
        // Create a new span that carries the incoming trace data.
        const traceparent = context.traceContext.traceParent;
        const tracestate = context.traceContext.traceState;
        const carrier = { traceparent, tracestate };

        // Extract the parent context
        // const parentContext = propagation.extract(otelContext.active(), carrier);
        const parentContext = propagation.extract(ROOT_CONTEXT, carrier);

        // Create and return a new span using the extracted context.
        const tracer = trace.getTracer('durableHelloHttpStart');
        const span = tracer.startSpan('myCustomSpan', {
            kind: SpanKind.CLIENT,
        }, parentContext);

        // Mark the span as active
        const activeCtx = trace.setSpan(otelContext.active(), span);
        return await otelContext.with(activeCtx, async () => {
            // Now any spans created inside this callback will automatically use "span"
            // as their parent.
            
            // Read the request body.
            const body = await request.text();

            const instanceId = await client.startNew(request.params.orchestratorName, {
                input: body
            });
            
            span.end();

            return client.createCheckStatusResponse(request, instanceId);
        });
    },
});

df.app.orchestration('durableCounterOrchestrator', function* (context) {
    const entityId = new df.EntityId(entityName, "myCounter");

    yield context.df.callEntity(entityId, "get");
    context.df.signalEntity(entityId, "add", 1);
});
```

leads to the following trace 
image


src/entities/Entity.ts
```diff
@@ -138,12 +144,19 @@ export class Entity<T> {
...
         operationInput?: unknown
     ): void {
         returnState.signals.push(
-            new Signal(entity, operationName, operationInput ? JSON.stringify(operationInput) : "")
+            new Signal(
+                entity,
+                operationName,
+                operationInput ? JSON.stringify(operationInput) : "",
+                request.id,
+                new Date().getTime()
```
bachuv, line 158: Does this match how we are saving the time in the WebJobs extension (i.e. DateTimeOffset)?

sophiatev: Yep yep, the method description is "Returns the stored time value in milliseconds since midnight, January 1, 1970 UTC." which is exactly what the WebJobs side expects when it calls `DateTimeOffset.FromUnixTimeMilliseconds` on this field value

bachuv: Great!

bachuv (approved):

sophiatev:
@cgillum, can you merge this PR?

PR #641, opened 2025-07-11
Update various packages to latest

andystaples:
Updates various packages based on 'npm audit' and component governance
- Change test and samples to file reference dependencies instead of npm 
- Bump axios 1.6.1 -> 1.10.0

Also updates transitive packages in package-lock.json across the project


samples-js/package.json
```diff
@@ -9,8 +9,8 @@
     },
     "dependencies": {
         "@azure/functions": "^4.7.0",
-        "axios": "^1.6.0",
-        "durable-functions": "^3.0.0",
+        "axios": "^1.10.0",
+        "durable-functions": "file:../durable-functions",
```
andystaples, line 13: Note: these had to be updated to reference dependencies instead of packages from npm, due to auto-generated package-lock.json containing versions that will flag component governance until release
We can omit these changes, but we will have to release the SDK, update the durable-functions versions, and commit again

cgillum: I see. So, the tradeoff is that the samples can't be copy/pasted and used as-is when we reference the package this way but doing it this way ensures that all samples are using the latest dependencies. I think I'm okay with this.

cgillum (approved):

PR #661, opened 2025-11-07
Changing the Exception Handling for Rewind

sophiatev:
This [PR](https://github.com/Azure/azure-functions-durable-extension/pull/3227) in the WebJobs repo changes the HTTP Status codes that are returned when rewind fails, so this PR updates error handling in this repo accordingly.

### DIFF
--- a/src/durableClient/DurableClient.ts
+++ b/src/durableClient/DurableClient.ts
@@ -398,17 +398,21 @@ export class DurableClient implements types.DurableClient {
             case 202:
                 return;
             case 404:
-                return Promise.reject(new Error(`No instance with ID '${instanceId}' found.`));
+                return Promise.reject(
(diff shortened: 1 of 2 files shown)


src/durableClient/DurableClient.ts
```diff
@@ -399,12 +399,18 @@ export class DurableClient implements types.DurableClient {
                 return;
             case 404:
                 return Promise.reject(new Error(`No instance with ID '${instanceId}' found.`));
-            case 410:
+            case 412:
                 return Promise.reject(
                     new Error(
```
andystaples, line 406: Can we include/use the error message sent by the HTTP response? See my comment on the WebJobs PR, this way we can include information that WebJobs or the backend know, like the current status of the orchestration, as long as it is communicated in the response.
Applies to all the cases in this handling block

andystaples (approved):

Issue #667, opened 2025-12-10
Node 24 incompability


detresed:
**Describe the bug**
There is an issue with node 24 although it is marked to work with node 18+,
`yield context.df.createTimer(retryDate)` runs into `System.Private.CoreLib: Exception while executing function: Microsoft.Azure.WebJobs.Extensions.DurableTask: Orchestrator function 'example' failed: (0 , util_1.isDate) is not a function.`
This is caused by the utils.isDate which was deprecated by node 22 and removed in node 24 in favour of utils.types.isDate

**Investigative information**

- Durable Functions extension version: 3.5.0
- durable-functions npm module version: 3.3.0
- Language (JavaScript/TypeScript) and version: TS 4.9.5
- Node.js version: 24.10

**Example code**
```
import * as df from 'durable-functions';
import { OrchestrationContext } from 'durable-functions';

df.app.orchestration('example',  function* (context: OrchestrationContext) {
    const retryDate = new Date();
    retryDate.setMinutes(retryDate.getMilliseconds() + 500);
    yield context.df.createTimer(retryDate);
    context.log('orchestration finished');
});
```

**Expected behavior**
Durable functions to run and work with node 18+ and durable tasks to execute sucessfully.

**Actual behavior**
Durable task error.

**Known workarounds**
Downgrade to node 22

jasonvangundy:
We're trying to avoid the resource burn of going 20 -> 22 and would benefit greatly from going to node 24.  A fix here would be 🎉

camerongibbs-eic:
1 line PR fix has been up since Feb 14. can comfirm the pr fixes in production. please merge and release so I don't need to maintain my own enterprise fork and package version for a 6 letter fix.

cgillum:
Adding @torosent to find an owner to review/merge this.

torosent:
We will take care of it asap

bachuv:
The fix is released in [v3.3.1](https://github.com/Azure/azure-functions-durable-js/releases/tag/v3.3.1).

camerongibbs-eic:
champion, thanks for getting this sorted so quickly!

PR #675, opened 2026-02-13
fix: replace deprecate utils.isDate() with utils.types.isDate()

azerum:
utils.isDate() have been removed in Node.js v24


coolcalcacol:
Any progress on this? will fix #667

PR #704, opened 2026-06-10
Add critical sections support

bachuv:
## Summary

Brings the **critical sections** API to Durable Functions for JS/TS. Orchestrations can now acquire one or more entity locks via `context.df.lock(...)` and perform a sequence of operations on those entities atomically, with no other orchestration able to interleave operations against the same entities. This brings JS to parity with the `LockAsync` API that in .NET.

## What's new

- **`context.df.lock(...entities)`** — acquires locks on one or more entities atomically. Yield the returned task to receive a `DurableLock`.
- **`DurableLock`** — represents the held critical section. Exposes `release()`, `[Symbol.dispose]()`, and `ownedLocks`.
- **`context.df.isLocked()`** — returns a `LockState` indicating whether the orchestration currently holds locks and which entities are locked.
- **`LockingRulesViolationError`** — thrown when an orchestration violates the critical-section rules (nested lock, sub-orchestration call inside a section, calls to non-locked entities, parallel calls to the same locked entity, signaling a locked entity).

## Rules enforced inside a critical section

| Rule | Why |
|------|-----|
| Cannot acquire another lock (no nested sections) | Prevents deadlock; the lock set is fixed for the lifetime of a section |
| Cannot invoke sub-orchestrations | Sub-orchestrations cannot inherit the parent's locks |
| Can only call entities that are in the lock set | Prevents reaching outside the section's atomicity boundary |
| Cannot make two parallel calls to the same locked entity | The entity processes one operation at a time; parallel calls would serialize anyway |
| Cannot signal an entity that is in the lock set | Signals bypass the lock; would break the section's atomicity guarantee |

## Lock Release patterns

JavaScript needs multiple patterns because `using` only landed in TS 5.2 / Node 24 (so older toolchains can't use it), there is no reliable finalizer (so the extension itself plays the safety-net role at orchestration end), and `try/finally` fills the gap for the large segment of customers stuck between those two extremes.

| Pattern | Boilerplate | Exception-safe? | Hold time | Min Node | Min TS |
|---------|-------------|-----------------|-----------|----------|--------|
| `using` | None | ✅ Automatic | Block scope | 24+ | 5.2+ |
| `try/finally` | 3 lines | ✅ If written correctly | Block scope | 18+ | Any |
| Implicit | None | ✅ Always | **Entire orchestration** | 18+ | Any |
| `try/finally` + early `release()` | 4 lines | ✅ + minimal hold time | Until first release | 18+ | Any |

### `using` (TS 5.2+ / Node 24+)
```js
using lock = yield context.df.lock(src, dst);
yield context.df.callEntity(src, "debit", amount);
yield context.df.callEntity(dst, "credit", amount);
// lock auto-released at scope exit, even on throw
```

### `try / finally`
```js
const lock = yield context.df.lock(src, dst);
try {
    yield context.df.callEntity(src, "debit", amount);
    yield context.df.callEntity(dst, "credit", amount);
} finally {
    lock.release();
}
```

### Implicit (no release — extension cleans up at orchestration end)
```js
yield context.df.lock(src, dst);
yield context.df.callEntity(src, "debit", amount);
yield context.df.callEntity(dst, "credit", amount);
// locks held until the orchestration terminates
```

### `try / finally` + early `release()`
```js
const lock = yield context.df.lock(src, dst);
try {
    yield context.df.callEntity(src, "debit", amount);
    yield context.df.callEntity(dst, "credit", amount);
    lock.release();                 // release early — other orchestrations can proceed
    yield context.df.callActivity("sendReceipt", { src, dst, amount });
} finally {
    lock.release();                 // safety net; idempotent if already released
}
```
### Protocol

Requires negotiated OOProc schema **V4+**; older extensions get a clear upgrade error. Reuses the existing `callEntity` replay shape (`RequestMessage` keyed by lock-request GUID, resolved by matching `EventRaised`) — no new history event type.

### Rule enforcement

Inside a section, the worker throws `LockingRulesViolationError` for: nested `lock(...)`, `callSubOrchestrator(...)`, `callEntity(...)` to an unlocked entity, parallel `callEntity(...)` to the same locked entity, and `signalEntity(...)` to a locked entity.

### Release semantics

`release()` is idempotent. `[Symbol.dispose]()` aliases `release()`. Implicit release at orchestration end is handled by the extension.

### Changes

16 files, +1166 / −14:

- **Actions:** `LockEntitiesAction`, `ReleaseEntitiesAction`
- **Error:** `LockingRulesViolationError`
- **Lock primitive:** `DurableLock` (with `Symbol.dispose` polyfill for Node <20)
- **Context:** `lock`, `isLocked`, rule enforcement in `DurableOrchestrationContext`
- **Executor:** lock-action translation and response handling in `TaskOrchestrationExecutor`
- **Schema:** added `V4` to `ReplaySchema`
- **Public types** and `index.ts` exports
- **Samples:** JS try/finally, JS `using`, TS transfer-entities
- **Tests:** 25 new tests in `test/integration/critical-sections-spec.ts`

### Tests

`npm test` — **262 passing**, 2 pending (pre-existing), 0 failing. Covers input validation, schema gating, action emission, `isLocked()` state, rule enforcement, replay integration, and release-pattern parity (`release()`, `using`, `try/finally`, double-release, implicit-on-end).

### DIFF
--- a/samples-js/functions/transferTryFinally.js
+++ b/samples-js/functions/transferTryFinally.js
@@ -1,0 +1,41 @@
+// Critical sections sample (JavaScript, `try/finally` — primary JS pattern).
+//
+// Works on all supported Node versions (18+). Requires
+// azure-functions-durable-extension with OOProc schema V4 or newer.
+//
+// For Node >= 24 (or Node 22 with --harmony-explicit-resource-management),
+// see `transferUsing.js` for the more ergonomic `using` form.
+
+const df = require("durable-functions");
+
+df.app.entity("transferAccount", async function (context) {
+    let balance = context.df.getState(() => 0);
+    switch (context.df.operationName) {
+        case "add":
+            balance += context.df.getInput();
+            break;
+        case "get":
+            context.df.return(balance);
+            break;
+    }
+    context.df.setState(balance);
+});
+
+df.app.orchestration("transferTryFinally", function* (context) {
+    const { fromKey, toKey, amount } = context.df.getInput();
+    const src = new df.EntityId("transferAccount", fromKey);
+    const dst = new df.EntityId("transferAccount", toKey);
+
+    const lock = yield context.df.lock(src, dst);
+    try {
+        const balance = yield context.df.callEntity(src, "get");
+        if (balance >= amount) {
+            yield context.df.callEntity(src, "add", -amount);
+            yield context.df.callEntity(dst, "add", amount);
+            return { transferred: true };
+        }
+        return { transferred: false, reason: "insufficient funds" };
+    } finally {
+        lock.release();
+    }
+});
--- a/src/entities/DurableLock.ts
+++ b/src/entities/DurableLock.ts
@@ -1,16 +1,19 @@
 import { EntityId } from "./EntityId";
 
-// Define Symbol.dispose for runtimes (Node <20) that lack it natively.
-// TypeScript >=5.2 emits this same polyfill in user code that uses `using`,
-// but adding it here ensures `[Symbol.dispose]` is callable even from plain
-// JavaScript without TS down-level emit.
-if (typeof (Symbol as { dispose?: symbol }).dispose === "undefined") {
-    Object.defineProperty(Symbol, "dispose", {
-        value: Symbol("Symbol.dispose"),
-        writable: false,
-        enumerable: false,
-        configurable: false,
-    });
+// Polyfill Symbol.dispose on Node <20, where the `[Symbol.dispose]()` member below needs it
+// (`using` only reads this symbol). `configurable: true` and try/catch keep it non-destructive.
+if (typeof (Symbol as { dispose?: symbol }).dispose !== "symbol") {
+    try {
+        Object.defineProperty(Symbol, "dispose", {
+            value: Symbol("Symbol.dispose"),
+            writable: false,
+            enumerable: false,
+            configurable: true,
+        });
+    } catch {
+        // Another polyfill defined it first so we read the global
+        // `Symbol.dispose` when wiring up the `[Symbol.dispose]()` member below.
+    }
 }
 
 /**
--- a/src/entities/DurableLock.ts
+++ b/src/entities/DurableLock.ts
@@ -1,6 +1,78 @@
+import { EntityId } from "./EntityId";
+
+// Define Symbol.dispose for runtimes (Node <20) that lack it natively.
+// TypeScript >=5.2 emits this same polyfill in user code that uses `using`,
+// but adding it here ensures `[Symbol.dispose]` is callable even from plain
+// JavaScript without TS down-level emit.
+if (typeof (Symbol as { dispose?: symbol }).dispose === "undefined") {
+    Object.defineProperty(Symbol, "dispose", {
+        value: Symbol("Symbol.dispose"),
+        writable: false,
+        enumerable: false,
+        configurable: false,
+    });
+}
+
 /**
- * @hidden
- * For use with locking operations.
- * TODO: improve this
+ * Returned by yielding `DurableOrchestrationContext.lock(...)`. Represents
+ * the critical section over the locked entities.
+ *
+ * Release options (in recommended order):
+ *   1. `using` syntax (TS >=5.2 or plain JS on Node >=24). Invokes
+ *      `[Symbol.dispose]()` at block exit, which is an alias for `release()`.
+ *   2. `try/finally` calling `release()`.
+ *   3. Implicit release at orchestration completion (handled by the extension).
+ *
+ * `release()` is idempotent.
  */
-export class DurableLock {}
+export class DurableLock {
+    /** The locks held by this critical section, in deterministic (sorted) order. */
+    public readonly ownedLocks: ReadonlyArray<EntityId>;
+
+    private released = false;
+    private readonly onRelease: () => void;
+
+    /**
+     * @hidden
+     * Tracks scheduler-ids of entity calls scheduled inside the section but
+     * not yet resolved, used to enforce the "no parallel call to same locked
+     * entity" rule.
+     */
+    public readonly inFlightEntityCalls: Set<string> = new Set<string>();
+
+    /** @hidden */
+    constructor(ownedLocks: EntityId[], onRelease: () => void) {
+        // Defensive freeze so plain-JS callers can't mutate the held-lock list at runtime.
+        // (`ReadonlyArray<T>` is compile-time only.)
+        this.ownedLocks = Object.freeze([...ownedLocks]);
+        this.onRelease = onRelease;
+    }
+
+    /**
+     * Release the lock. Idempotent.
+     *
+     * Emits a `ReleaseEntities` action that the extension translates into
+     * `DurableOrchestrationContext.ReleaseLocks()`. If the lock has already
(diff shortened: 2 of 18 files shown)


samples-js/functions/transferTryFinally.js
```diff
@@ -0,0 +1,41 @@
+// Critical sections sample (JavaScript, `try/finally` — primary JS pattern).
+//
+// Works on all supported Node versions (18+). Requires
+// azure-functions-durable-extension with OOProc schema V4 or newer.
```
nytian, line 4: nit: We need to release a new durable extension version, right? Maybe add the pkg version number here for better clarity.

src/entities/DurableLock.ts
```diff
@@ -1,6 +1,78 @@
+import { EntityId } from "./EntityId";
+
+// Define Symbol.dispose for runtimes (Node <20) that lack it natively.
```
nytian: This is interesting. So with later Node version that supports dispose auto this won't cause issue? I mean if double dispose call will cause error

test/integration/critical-sections-spec.ts
```diff
@@ -0,0 +1,594 @@
...
+        "6.00:00:00",
+        30000,
+        schema,
+        undefined,
+        executor
+    );
+    return { ctx, executor };
+}
+
+describe("Critical Sections (lock / isLocked)", () => {
```
nytian, line 59: Can we add some comments here about what test we added for clarity? Also, it seems we test parallel calls to an entity which should throw error , and can we add a sequential call to a locked entity which should be allowed? Another nit is I wonder if we have test for re-lock after release, like 1. lock a 2. dosomething 3. release a. 4. lock a agagin succeeds? This confirms current lock is reset to undefined and  we begin a new critical section.

src/orchestrations/DurableOrchestrationContext.ts
```diff
@@ -25,12 +25,21 @@ import { CallSubOrchestratorWithRetryAction } from "../actions/CallSubOrchestrat
 import { ContinueAsNewAction } from "../actions/ContinueAsNewAction";
 import { CreateTimerAction } from "../actions/CreateTimerAction";
 import { ExternalEventType } from "../actions/ExternalEventType";
+import { LockEntitiesAction } from "../actions/LockEntitiesAction";
```
nytian, line 29: It seems like Context and Executor are sharing some info so make the lock work, what if there is something wrong and one of two didn't update this shared info correctly? Maybe it's better to have some additional check/validation first or using a well-defined shared interface? But this is minorand up to you about the design here :)

nytian (commented):
Thanks for the PR! Left a few minor comments. Also, a question regarding forward compatibility: for users who are using extensions that don’t support V4, will they have errors? Though I think this is not very common since they typically use bundles, which keep dependencies aligned, but wanted to confirm.

samples-js/functions/transferTryFinally.js
bachuv: Yes, but it's not released yet so I'll add this information after the release is complete.

src/entities/DurableLock.ts
bachuv: Symbol.dispose just calls release(), which is idempotent via the released flag so it won't cause an issue.

src/orchestrations/DurableOrchestrationContext.ts
bachuv: The Context→Executor handoff now uses a typed LockTask (readonly lockResult: DurableLock) instead of the ad-hoc __lockResult property

src/orchestrations/DurableOrchestrationContext.ts
```diff
@@ -341,4 +390,134 @@ export class DurableOrchestrationContext implements types.DurableOrchestrationCo
...
+
+        // The DurableLock the orchestrator generator receives on
+        // `yield ctx.df.lock(...)`. It is carried on the returned LockTask as a
+        // typed field (see LockTask), so the executor can hand it back on
+        // completion without a shared untyped property between the two.
+        const lock = new DurableLock(deduped, () => {
+            this.taskOrchestratorExecutor.recordFireAndForgetAction(new ReleaseEntitiesAction());
+            // Clear the active-section flag so subsequent code outside the
+            // section is no longer treated as locked.
+            if (this.currentLock === lock) {
```
sophiatev, line 496: It should be impossible that this if statement resolves to false, right? Should we throw an exception or something if that is ever the case? (genuine question)

test/integration/critical-sections-spec.ts
sophiatev, line 1: Same thought as from the WebJobs PR:

> maybe it would make sense to test that the locking functionality actually "works"? One way to do this would be to start two instances of the orchestration and add some sort of 5 second timer call inside the critical section. It should be the case then that one of the orchestrations (whichever loses the lock race) has to wait at least 5 seconds to finish, since it needs to wait for the first orchestration to release the locks before it can proceed. (Not sure what the best way to measure how long it takes an orchestration to finish is - do we expose this in our metadata?)

But I can't actually tell if we are exercising the full orchestrations locking/unlocking entities path in the "Replay-integration tests". It seems like we are manually replaying the orchestrations or something along those lines (this may just be my ignorance of JS coming into play here). So if what I suggested is impossible then feel free to ignore this comment

bachuv: I added the end to end test in the extension since we have the e2e framework there. I'm not sure if we can mirror that same setup in this repo right now.

src/orchestrations/DurableOrchestrationContext.ts
bachuv: Yes, but I don't think we should throw an exception because it could hide a user exception when release() is called. I think we should keep the if statement to avoid any unintentional bugs in the future.

sophiatev (approved):

src/orchestrations/DurableOrchestrationContext.ts
```diff
@@ -341,4 +390,134 @@ export class DurableOrchestrationContext implements types.DurableOrchestrationCo
...
+     * an `EventRaised` whose `Name` equals that same GUID. The executor's
+     * `EventSent` handler re-keys this task to that GUID, so the subsequent
+     * `EventRaised` matches and resolves the lock task — without requiring
+     * any new history event type.
+     */
+    public lock(first: types.EntityId | types.EntityId[], ...rest: types.EntityId[]): Task {
+        if (this.schemaVersion < ReplaySchema.V4) {
+            throw new Error(
+                `lock requires a Durable Functions extension that advertises OOProc schema V4 or higher (negotiated V${
+                    (this.schemaVersion as number) + 1
```
nytian: why +1 here?

nytian (approved):

src/orchestrations/DurableOrchestrationContext.ts
bachuv: It's because the schemaVersion starts at 0, so V1 is 0. I updated this message to be clearer.

``lock requires a Durable Functions extension that supports OOProc schema V4 or higher (this extension supports up to ${
                    ReplaySchema[this.schemaVersion]
                }). Please upgrade the Microsoft.Azure.WebJobs.Extensions.DurableTask package.``

nytian (approved):

PR #705, opened 2026-06-12
Set custom Exception Details

nytian:
This pull request introduces structured error handling for failed activities and orchestrations, enabling richer error information (including custom properties) to be surfaced from activity failures back to orchestrators and clients. The main changes include support for registering an `ExceptionPropertiesProvider`, wrapping activity handlers to serialize enriched error payloads, and updating the orchestration executor to consume and propagate these structured errors. This improves debuggability and extensibility for Durable Functions workflows.

**Structured error handling and propagation:**

- Added the ability to register an `ExceptionPropertiesProvider` via the new `setExceptionPropertiesProvider` API in `app.ts`, allowing custom properties to be attached to errors thrown from activities.
- Activity handlers are now wrapped so that, when an error is thrown, it is serialized with custom properties (if provided) into a JSON payload matching the Durable Task host's `TaskFailureDetails` shape. This payload is then surfaced to the Functions host for richer error reporting. 
- Introduced the `TaskFailedError` class, which is thrown into orchestrator generators when a scheduled activity or sub-orchestration fails and the host provides structured `FailureDetails`. This class exposes all error details, including custom properties. (`src/error/TaskFailedError.ts`, `src/index.ts`) 
**History event and DTO updates:**

- Extended history event types (`TaskFailedEvent`, `SubOrchestrationInstanceFailedEvent`) and their options to include a `FailureDetails` property, which carries the structured error payload. Added a wire-format DTO (`FailureDetailsDto`) to match the host's payload shape
- Updated the orchestration executor to extract and prefer structured `FailureDetails` from history events, falling back to legacy string properties when not available. (`src/orchestrations/TaskOrchestrationExecutor.ts`) 

These changes enable Durable Functions users to attach and propagate custom error metadata, greatly improving error transparency and diagnostics for orchestrations and activities.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -15,7 +15,7 @@ import { app as azFuncApp } from "@azure/functions";
 import { RegisteredOrchestrationTask } from "./task/RegisteredOrchestrationTask";
 import { RegisteredActivityTask } from "./task/RegisteredActivityTask";
 import {
-    appendExceptionPropertiesSuffix,
+    buildTaskFailureDetailsJson,
     setRegisteredExceptionPropertiesProvider,
 } from "./error/ExceptionPropertiesProvider";
 
@@ -77,29 +77,40 @@ export function setExceptionPropertiesProvider(
 }
 
 /**
- * Wraps a user-supplied activity handler so that thrown errors are augmented
- * with custom properties from the registered {@link ExceptionPropertiesProvider}
- * before propagating to the Functions host.
+ * Wraps a user-supplied activity handler so that thrown errors are reshaped
+ * into a TaskFailureDetails JSON payload (with custom properties from the
+ * registered {@link ExceptionPropertiesProvider}) before propagating to the
+ * Functions host. The host extension's OutOfProcMiddleware parses this payload
+ * and surfaces the structured properties on `FailureDetails.Properties`.
+ *
+ * If no provider is registered or it returns no properties, the error is
+ * re-thrown untouched so the legacy `ExceptionType: Message` wire format is
+ * preserved.
  */
 function wrapActivityHandler(handler: ActivityHandler): ActivityHandler {
     return async (triggerInput, context) => {
         try {
             return await handler(triggerInput, context);
         } catch (err) {
-            const original = err instanceof Error ? err : new Error(String(err));
-            const augmented = appendExceptionPropertiesSuffix(original.message, err);
-            if (augmented !== original.message) {
-                try {
-                    original.message = augmented;
-                } catch {
-                    // `message` is non-writable on some custom Error subclasses;
-                    // fall back to a fresh Error that preserves the original stack.
-                    const wrapped = new Error(augmented);
-                    wrapped.stack = original.stack;
-                    throw wrapped;
-                }
+            const serialized = buildTaskFailureDetailsJson(err);
+            if (!serialized) {
+                throw err;
             }
-            throw original;
+            const original = err instanceof Error ? err : new Error(String(err));
+            let assigned = true;
+            try {
+                original.message = serialized;
+            } catch {
+                assigned = false;
+            }
+            if (assigned) {
+                throw original;
+            }
+            // `message` is non-writable on some custom Error subclasses;
+            // fall back to a fresh Error that preserves the original stack.
+            const wrapped = new Error(serialized);
+            wrapped.stack = original.stack;
+            throw wrapped;
         }
     };
 }
--- a/src/app.ts
+++ b/src/app.ts
@@ -86,8 +86,10 @@ export function setExceptionPropertiesProvider(
  * If no provider is registered or it returns no properties, the error is
  * re-thrown untouched so the legacy `ExceptionType: Message` wire format is
  * preserved.
+ *
+ * @hidden
  */
-function wrapActivityHandler(handler: ActivityHandler): ActivityHandler {
+export function wrapActivityHandler(handler: ActivityHandler): ActivityHandler {
     return async (triggerInput, context) => {
         try {
             return await handler(triggerInput, context);
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,7 +1,9 @@
 import {
+    ActivityHandler,
     ActivityOptions,
     EntityHandler,
     EntityOptions,
+    ExceptionPropertiesProvider,
     OrchestrationHandler,
     OrchestrationOptions,
     RegisteredActivity,
@@ -12,6 +14,10 @@ import { createOrchestrator, createEntityFunction } from "./util/testingUtils";
 import { app as azFuncApp } from "@azure/functions";
 import { RegisteredOrchestrationTask } from "./task/RegisteredOrchestrationTask";
 import { RegisteredActivityTask } from "./task/RegisteredActivityTask";
+import {
+    appendExceptionPropertiesSuffix,
+    setRegisteredExceptionPropertiesProvider,
+} from "./error/ExceptionPropertiesProvider";
 
 export function orchestration(
     functionName: string,
@@ -54,6 +60,7 @@ export function activity(functionName: string, options: ActivityOptions): Regist
     azFuncApp.generic(functionName, {
         trigger: trigger.activity(),
         ...options,
+        handler: wrapActivityHandler(options.handler),
     });
 
     const result: RegisteredActivity = (input?: unknown): RegisteredActivityTask => {
@@ -61,6 +68,40 @@ export function activity(functionName: string, options: ActivityOptions): Regist
     };
 
     return result;
+}
+
+export function setExceptionPropertiesProvider(
+    provider: ExceptionPropertiesProvider | undefined
+): void {
+    setRegisteredExceptionPropertiesProvider(provider);
+}
+
+/**
+ * Wraps a user-supplied activity handler so that thrown errors are augmented
+ * with custom properties from the registered {@link ExceptionPropertiesProvider}
+ * before propagating to the Functions host.
+ */
+function wrapActivityHandler(handler: ActivityHandler): ActivityHandler {
+    return async (triggerInput, context) => {
+        try {
+            return await handler(triggerInput, context);
+        } catch (err) {
+            const original = err instanceof Error ? err : new Error(String(err));
+            const augmented = appendExceptionPropertiesSuffix(original.message, err);
+            if (augmented !== original.message) {
+                try {
+                    original.message = augmented;
+                } catch {
+                    // `message` is non-writable on some custom Error subclasses;
+                    // fall back to a fresh Error that preserves the original stack.
+                    const wrapped = new Error(augmented);
+                    wrapped.stack = original.stack;
+                    throw wrapped;
+                }
+            }
+            throw original;
+        }
+    };
 }
 
 export * as client from "./client";
--- a/src/error/ExceptionPropertiesProvider.ts
+++ b/src/error/ExceptionPropertiesProvider.ts
@@ -59,22 +59,22 @@ export function buildTaskFailureDetailsJson(error: unknown): string | undefined 
         return undefined;
     }
 
-    const errorObj = error instanceof Error ? error : undefined;
-    const errorType = errorObj?.constructor?.name ?? "Error";
-    const errorMessage =
-        errorObj?.message ?? (typeof error === "string" ? error : JSON.stringify(error));
-
-    const payload: Record<string, unknown> = {
-        errorType,
-        errorMessage,
-        isNonRetriable: false,
-        properties,
-    };
-    if (errorObj?.stack) {
-        payload.stackTrace = errorObj.stack;
-    }
-
     try {
+        const errorObj = error instanceof Error ? error : undefined;
+        const errorType = errorObj?.constructor?.name ?? "Error";
+        const errorMessage =
+            errorObj?.message ?? (typeof error === "string" ? error : JSON.stringify(error));
+
+        const payload: Record<string, unknown> = {
+            errorType,
+            errorMessage,
+            isNonRetriable: false,
+            properties,
+        };
+        if (errorObj?.stack) {
+            payload.stackTrace = errorObj.stack;
+        }
+
         return JSON.stringify(payload);
     } catch {
         return undefined;
--- a/src/error/ExceptionPropertiesProvider.ts
+++ b/src/error/ExceptionPropertiesProvider.ts
@@ -10,13 +10,6 @@ export function setRegisteredExceptionPropertiesProvider(
     registeredProvider = provider;
 }
 
-/** @hidden */
-export function getRegisteredExceptionPropertiesProvider():
-    | types.ExceptionPropertiesProvider
-    | undefined {
-    return registeredProvider;
-}
-
 /**
  * @hidden
  * Safely invokes the registered provider. Returns `undefined` when no provider
@@ -59,22 +52,22 @@ export function buildTaskFailureDetailsJson(error: unknown): string | undefined 
         return undefined;
     }
 
+    const errorObj = error instanceof Error ? error : undefined;
+    const errorType = errorObj?.name ?? errorObj?.constructor?.name ?? "Error";
+    const errorMessage =
+        errorObj?.message ?? (typeof error === "string" ? error : JSON.stringify(error));
+
+    const payload: Record<string, unknown> = {
+        errorType,
+        errorMessage,
+        isNonRetriable: false,
+        properties,
+    };
+    if (errorObj?.stack) {
+        payload.stackTrace = errorObj.stack;
+    }
+
     try {
-        const errorObj = error instanceof Error ? error : undefined;
-        const errorType = errorObj?.constructor?.name ?? "Error";
-        const errorMessage =
-            errorObj?.message ?? (typeof error === "string" ? error : JSON.stringify(error));
-
-        const payload: Record<string, unknown> = {
-            errorType,
-            errorMessage,
-            isNonRetriable: false,
-            properties,
-        };
-        if (errorObj?.stack) {
-            payload.stackTrace = errorObj.stack;
-        }
-
         return JSON.stringify(payload);
     } catch {
         return undefined;
--- a/src/error/ExceptionPropertiesProvider.ts
+++ b/src/error/ExceptionPropertiesProvider.ts
@@ -39,25 +39,34 @@ export function extractExceptionProperties(error: unknown): Record<string, unkno
 
 /**
  * @hidden
- * Sentinel label appended to an error message to carry custom exception
- * properties back to the Durable Task host extension. Mirrors the existing
- * `$OutOfProcData$` convention so the extension can locate and strip it.
+ * Builds the single-line JSON payload that mirrors the protobuf
+ * `TaskFailureDetails` shape consumed by the Durable Task host extension's
+ * `OutOfProcMiddleware.TryExtractSerializedFailureDetailsFromException`.
  *
- * NOTE: Single source of truth for the wire format. If the host extension
- * expects a different sentinel or encoding, change it here only.
+ * Returns `undefined` when the registered provider yields no properties for
+ * the given error — callers should then leave the original error untouched so
+ * existing JS wire-format behaviour is preserved for users who haven't opted in.
  */
-export const exceptionPropertiesLabel = "\n\n$FailureProperties$:";
-
-/**
- * @hidden
- * Returns `message` with a serialized exception-properties suffix appended
- * when the registered provider yields properties for `error`. Otherwise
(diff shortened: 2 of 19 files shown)


src/error/ExceptionPropertiesProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+    const payload: Record<string, unknown> = {
+        errorType,
+        errorMessage,
+        isNonRetriable: false,
+        properties,
+    };
+    if (errorObj?.stack) {
+        payload.stackTrace = errorObj.stack;
+    }
+    return JSON.stringify(payload);
```
bachuv: Can you add a try/catch around this line? If this throws an exception then the user will see this exception instead of the original exception.


```suggestion
    try {
        return JSON.stringify(payload);
    } catch {
        return undefined;
    }
```

src/error/ExceptionPropertiesProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+    const errorType =
+        errorObj?.constructor?.name ?? (typeof error === "string" ? "Error" : "Error");
+    const errorMessage =
+        errorObj?.message ?? (typeof error === "string" ? error : JSON.stringify(error));
+
+    const payload: Record<string, unknown> = {
+        errorType,
+        errorMessage,
+        isNonRetriable: false,
+        properties,
```
bachuv, line 74: Should this payload also include the `innerFailure`?

src/app.ts
```diff
@@ -63,4 +70,49 @@ export function activity(functionName: string, options: ActivityOptions): Regist
...
+ * into a TaskFailureDetails JSON payload (with custom properties from the
+ * registered {@link ExceptionPropertiesProvider}) before propagating to the
+ * Functions host. The host extension's OutOfProcMiddleware parses this payload
+ * and surfaces the structured properties on `FailureDetails.Properties`.
+ *
+ * If no provider is registered or it returns no properties, the error is
+ * re-thrown untouched so the legacy `ExceptionType: Message` wire format is
+ * preserved.
+ */
+function wrapActivityHandler(handler: ActivityHandler): ActivityHandler {
```
bachuv: Do we also need handlers for orchestrations and entities?

src/error/TaskFailedError.ts
```diff
@@ -0,0 +1,39 @@
+import * as types from "durable-functions";
+import { FailureDetailsDto } from "../history/FailureDetailsDto";
+
+/**
+ * Error subclass thrown into orchestrator generators when a scheduled activity
+ * or sub-orchestration fails and the host has provided structured
```
bachuv, line 6: Should there be a `SubOrchestrationFailedError` class for suborchestrations instead of reusing the same TaskFailedError for both? What does .NET do? Let's match .NET.

src/error/ExceptionPropertiesProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export function buildTaskFailureDetailsJson(error: unknown): string | undefined {
+    const properties = extractExceptionProperties(error);
+    if (!properties) {
+        return undefined;
+    }
+
+    const errorObj = error instanceof Error ? error : undefined;
+    const errorType =
+        errorObj?.constructor?.name ?? (typeof error === "string" ? "Error" : "Error");
```
bachuv: Looks like both branches evaluate to "Error". What is the intended behavior?

src/error/TaskFailedError.ts
nytian: yes it's matching .NET use TaskFailedException for both sub-orchestrations and activity functions.

src/error/ExceptionPropertiesProvider.ts
nytian: udpated

src/error/ExceptionPropertiesProvider.ts
bachuv, line 1: I wasn't able to get the feature working end to end in a sample app locally. The orchestrator didn't get a TaskFailedError and the custom properties JSON just showed up as plain text in err.message. Can you try a local E2E test and confirm the expected behavior?

I asked Copilot to investigate and it said the following:

> What's broken: For Node.js apps the host extension routes activity failures through TaskActivityShim, which discards the worker's serialized TaskFailureDetails JSON into a plain TaskFailureException.Reason/Details pair, so the JS orchestrator never sees a populated FailureDetails field on the TaskFailed history event and TaskFailedError (with failureDetails.properties) is never thrown.
> 
> Fix: Update TaskActivityShim.RunAsync to call the existing TryExtractSerializedFailureDetailsFromException helper on the activity's exception message and attach the parsed FailureDetails to the TaskFailedEvent it returns (matching what OutOfProcMiddleware.CallActivityAsync already does for the gRPC path).

src/error/ExceptionPropertiesProvider.ts
nytian: We don't have innerFailure at .NET isolated so I guess we should match the behavior? We only have this at.net-inprocess.

src/error/ExceptionPropertiesProvider.ts
nytian: yeah we need a change at extension repo too to make this work. I have a draft PR there

bachuv: Great! Can you share a link to that PR?

src/app.ts
nytian: Orchestrations: Yes. I just added it. If a sub-orch fails, the parent needs the structured FailureDetails inner chain + properties to see the original activity error. The activity path alone doesn't cover that.

Entities: No. Entity failures come back as the operation result, not a propagating task failure, no nested chain to preserve, so nothing to add there. I think at . net isolated we didn' support entity either

src/error/ExceptionPropertiesProvider.ts
nytian: yes we need a extension repo change too, I have a draft PR there

src/error/ExceptionPropertiesProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+export function buildTaskFailureDetailsJson(error: unknown): string | undefined {
+    const properties = extractExceptionProperties(error);
+    if (!properties) {
+        return undefined;
+    }
+
+    const errorObj = error instanceof Error ? error : undefined;
+    const errorType = errorObj?.constructor?.name ?? "Error";
+    const errorMessage =
+        errorObj?.message ?? (typeof error === "string" ? error : JSON.stringify(error));
```
bachuv: Should `JSON.stringify()` be in a try/catch?

src/orchestrations/TaskOrchestrationExecutor.ts
```diff
@@ -132,6 +152,15 @@ export class TaskOrchestrationExecutor {
...
             schemaVersion: this.schemaVersion,
+            // If the orchestration failed because an uncaught sub-orchestration or
+            // activity failure propagated out, relay its structured FailureDetails
+            // (including the InnerFailure chain and any custom Properties) so the
+            // host can forward it to a calling parent orchestration. Only set on
+            // failure; omitted otherwise, leaving existing behavior unchanged.
+            failureDetails:
+                this.exception instanceof TaskFailedError
+                    ? toFailureDetailsDto(this.exception.failureDetails)
+                    : undefined,
```
bachuv, line 167: From what I can see, we populate the failure details for any exception from a sub-orchestration or activity. Do we do the equivalent for exceptions thrown in the orchestrator?

types/taskFailedError.d.ts
```diff
@@ -0,0 +1,50 @@
...
+    /** Stack trace, when available. */
+    stackTrace?: string;
+    /** Whether the failure was flagged non-retriable by the worker. */
+    isNonRetriable: boolean;
+    /**
+     * Custom key/value properties attached by the failing worker via its
+     * `ExceptionPropertiesProvider`. Undefined when the worker attached no
+     * properties or when the host did not propagate them.
+     */
+    properties?: Record<string, unknown> | null;
```
bachuv: Can this be `null`? Seems like it would either have a value or be undefined.

src/error/TaskFailedError.ts
```diff
@@ -0,0 +1,59 @@
...
+ * `FailureDetails` (including any custom `Properties` attached via an
+ * {@link types.ExceptionPropertiesProvider} on the failing worker).
+ *
+ * Falls back to a plain `Error` for legacy host versions that do not include
+ * `FailureDetails` on the history event.
+ */
+export class TaskFailedError extends Error implements types.TaskFailedError {
+    public readonly failureDetails: types.FailureDetails;
+
+    constructor(failureDetails: types.FailureDetails) {
```
bachuv: Should we also set the `TaskName` and `TaskId` like in .NET - https://github.com/microsoft/durabletask-dotnet/blob/3e92eb6732645973de7d1598386b3b33188dba7b/src/Abstractions/TaskFailedException.cs#L22?

types/taskFailedError.d.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Structured details about a failed activity, orchestration, or entity
+ * operation. Mirrors `DurableTask.Core.FailureDetails`.
+ */
+export interface FailureDetails {
```
bachuv, line 5: Should we have the `isCausedBy` helper like in .NET?

src/error/TaskFailedError.ts
nytian: these properties are not passed here at JS, that requires more changes and do we have to add them? I would prefer no.

src/orchestrations/TaskOrchestrationExecutor.ts
nytian: no this only applies to TaskFailedException surfaced from activity functions or sub orchestrations,  for general error we don't catch, which will just be saved to state's error. Also we don't know the character of general error so I dont think its needs to be caught here...

src/error/TaskFailedError.ts
bachuv: I was suggesting to add them for .NET parity. Does it not change the behavior to be different from .NET?

types/taskFailedError.d.ts
nytian: added

src/history/FailureDetailsDto.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * @hidden
+ * Wire-format DTO matching the JSON shape the Durable Task host extension
+ * serializes for `DurableTask.Core.FailureDetails` (PascalCase via
+ * `[DataMember]`). Includes the `Properties` map added by
+ * https://github.com/Azure/azure-functions-durable-extension/pull/3215.
+ */
+export interface FailureDetailsDto {
```
bachuv: non-blocking, but I'm wondering if `FailureDetailsPayload` is a better name.

bachuv (approved):

src/error/TaskFailedError.ts
```diff
@@ -12,48 +12,78 @@ import { FailureDetailsDto } from "../history/FailureDetailsDto";
...
+        this.taskName = taskName;
+        this.taskId = taskId;
         if (failureDetails.stackTrace) {
             this.stack = failureDetails.stackTrace;
         }
     }
 
-    public static fromWireDto(dto: FailureDetailsDto): TaskFailedError {
-        return new TaskFailedError(toFailureDetails(dto));
+    public static fromWireDto(
```
bachuv: Can this method be private?

nytian: udpated

andystaples (commented):
## Review summary

Structured error propagation — `ExceptionPropertiesProvider` + `TaskFailedError` + `FailureDetails` on history events and the OOP orchestrator state. I traced the wire contract end-to-end and it lines up: inbound `FailureDetails` (PascalCase) on `TaskFailed` / `SubOrchestrationInstanceFailed` events, the activity `{...}`-prefixed payload consumed by the host's `TryExtractSerializedFailureDetailsFromException`, and the outbound `failureDetails` on the OOP state matching the extension's `OutOfProcOrchestratorState`. Backward-compat fallbacks (legacy `Reason`/`Details`; omitting `failureDetails` for non-`TaskFailedError`) are correct, and build/lint/tests all pass locally (160 unit tests, incl. the 23 new ones).

The comments below are the issues and considerations worth a look before merge. Nothing here is a hard blocker on the JS code itself, but the first two — the orchestrator doc/impl mismatch and the release-coupling with the extension parser — are worth resolving or explicitly acknowledging.

types/app.d.ts
```diff
@@ -56,4 +57,21 @@ export function entity<T = unknown>(functionName: string, options: EntityOptions
  */
 export function activity(functionName: string, options: ActivityOptions): RegisteredActivity;
 
+/**
+ * Registers a global {@link ExceptionPropertiesProvider} for the function app.
+ *
+ * When an activity or orchestrator function throws, the provider is consulted
```
andystaples: **Docs promise orchestrator coverage that isn't implemented.** Only activity handlers are wrapped (`wrapActivityHandler` in `app.ts`); orchestrators are registered via `createOrchestrator` unwrapped, so the provider is never consulted for an error an orchestrator raises directly. Either wrap orchestrator handlers too, or narrow this doc (and the matching one in `exceptionPropertiesProvider.d.ts`) to activities only. FWIW, in .NET the provider covers activities + *unexpected* orchestrator-dispatch errors; normal orchestrator user-code failures are handled by the executor.

src/app.ts
```diff
@@ -63,4 +70,49 @@ export function activity(functionName: string, options: ActivityOptions): Regist
...
+            return await handler(triggerInput, context);
+        } catch (err) {
+            const serialized = buildTaskFailureDetailsJson(err);
+            if (!serialized) {
+                throw err;
+            }
+            const original = err instanceof Error ? err : new Error(String(err));
+            let assigned = true;
+            try {
+                original.message = serialized;
```
andystaples, line 104: **Opting in on an older host extension degrades error messages.** Replacing `message` with the JSON payload only helps if the host has the structured parser (`TryExtractSerializedFailureDetailsFromException`), which is in the companion extension PR and not yet released. On an extension without it, `GetFailureDetails` falls through to `TrySplitExceptionTypeFromMessage`, which splits the JSON on the first `": "` — yielding a garbled `errorType`/message that the orchestrator then sees. Since the npm SDK upgrades independently of the deployed extension bundle, this mismatch is plausible and strictly worse than not opting in. Consider gating on a detectable capability / schema version, and/or documenting the minimum extension version and shipping the extension side first.

src/app.ts
andystaples: **Test gap:** the mutation path here — overwriting `message`, and the non-writable-`message` fallback that builds a fresh `Error` preserving the stack — isn't directly covered. `buildTaskFailureDetailsJson` / `extractExceptionProperties` and the executor / `TaskFailedError` path are well tested, but this wrapper isn't. Worth a small test: register a provider, invoke a wrapped handler that throws, and assert the rethrown `message` is the JSON payload; plus a custom `Error` with a frozen `message` to exercise the fallback branch.

src/error/ExceptionPropertiesProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+            errorType,
+            errorMessage,
+            isNonRetriable: false,
+            properties,
+        };
+        if (errorObj?.stack) {
+            payload.stackTrace = errorObj.stack;
+        }
+
+        return JSON.stringify(payload);
```
andystaples, line 81: **Heads-up on sanitization:** the Node worker runs `sanitizeErrorString(error.message)` before sending to the host. If the serialized payload contains substrings like `key=`, `?code=`, `Token=`, or a `user:pass@host:port` URL (in the message, stack, or any property value), the sanitizer rewrites them to `[Hidden Credential]`, which breaks the JSON and causes a silent fallback to legacy (properties dropped). Largely unavoidable, but worth a doc note so users avoid such values in provider properties.

src/error/ExceptionPropertiesProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+export function buildTaskFailureDetailsJson(error: unknown): string | undefined {
+    const properties = extractExceptionProperties(error);
+    if (!properties) {
+        return undefined;
+    }
+
+    try {
+        const errorObj = error instanceof Error ? error : undefined;
+        const errorType = errorObj?.constructor?.name ?? "Error";
```
andystaples: `constructor.name` gets mangled under minification/bundling and ignores a user-set `error.name` that differs from the class name. Consider `errorObj.name ?? errorObj.constructor?.name ?? "Error"` — closer to conventional JS error identification (and to the `extractExceptionProperties` tests, which read `err.name`).

src/error/ExceptionPropertiesProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+        const errorMessage =
+            errorObj?.message ?? (typeof error === "string" ? error : JSON.stringify(error));
+
+        const payload: Record<string, unknown> = {
+            errorType,
+            errorMessage,
+            isNonRetriable: false,
```
andystaples: `isNonRetriable` is hard-coded `false` with no way for a provider to signal otherwise, unlike the .NET provider. Fine for v1 — flagging in case surfacing it was intended.

src/error/ExceptionPropertiesProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/** @hidden */
+export function setRegisteredExceptionPropertiesProvider(
+    provider: types.ExceptionPropertiesProvider | undefined
+): void {
+    registeredProvider = provider;
+}
+
+/** @hidden */
+export function getRegisteredExceptionPropertiesProvider():
```
andystaples: `getRegisteredExceptionPropertiesProvider` is exported but never referenced in `src` or `test`. Dead code — remove it (or wire it up).

src/error/TaskFailedError.ts
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Error subclass thrown into orchestrator generators when a scheduled activity
+ * or sub-orchestration fails and the host has provided structured
+ * `FailureDetails` (including any custom `Properties` attached via an
+ * {@link types.ExceptionPropertiesProvider} on the failing worker).
+ *
+ * Falls back to a plain `Error` for legacy host versions that do not include
+ * `FailureDetails` on the history event.
+ */
+export class TaskFailedError extends Error implements types.TaskFailedError {
```
andystaples, line 13: Conceptual overlap with the existing public `DurableError`, which is *documented* as "thrown when a scheduled activity or suborchestrator has failed" and advertised for `instanceof` guards to catch library errors (though today it's only used for an internal error). Since `TaskFailedError` is what's actually thrown on task failure now, consider having it extend `DurableError` — or at least reconciling the two docs — so existing `instanceof DurableError` guards still catch task failures.

src/error/ExceptionPropertiesProvider.ts
nytian: yeah the provider is just for custom properties of errors, not controlling retries. So we hard code as false, and let existing retry policy owns the retry feature. I think this is more clear and maybe we can add this enhancement in future.

src/error/ExceptionPropertiesProvider.ts
nytian: yes edge case we can't avoid. Add a comment for this case. If this happened it will be fallback to legacy

src/app.ts
nytian: added

types/app.d.ts
nytian: update doc

src/app.ts
nytian: update doc

types/index.d.ts
```diff
@@ -87,9 +89,13 @@ export declare class RetryOptions {
 }
 
 /**
- * A specfic error thrown when a scheduled activity or suborchestrator has failed.
- * This error can be checked for via `instanceof` guards to catch only exceptions thrown
- * by the DurableJS library.
+ * A specific error thrown by the DurableJS library. It is raised when a task's
+ * retry policy has exhausted all attempts. This error can be checked for via
```
andystaples: Small nit: this description is still slightly off. `DurableError` is only ever thrown at `RetryableTask.ts:50` for the internal "no executor passed" guard — on retry exhaustion the task returns the underlying error unwrapped (`setValue(true, this.error)`), not a `DurableError`. So "raised when a task's retry policy has exhausted all attempts" doesn't quite match the code. Not blocking — just consider describing it as an internal framework error (the `TaskFailedError` note below is spot on).

andystaples (approved):

andystaples (approved):

PR #706, opened 2026-06-12
Add getActivityInvocationInfo and getInstanceRetryHistory helpers for activity retry visibility

bachuv:
## Summary

Two new public helpers for the activity retry visibility feature.

### `getActivityInvocationInfo(context)` — activity-side

Reads `durabletask.attempt` / `maxAttempts` / `isMaxAttempt` trigger
metadata exposed on the activity trigger binding for activities scheduled
with a `RetryOptions` policy.

```ts
interface ActivityInvocationInfo {
    readonly attempt: number;       // counter starts at 1; defaults to 1 when unavailable
    readonly maxAttempts: number;   // defaults to 1 when unavailable
    readonly isMaxAttempt: boolean;
    readonly metadataAvailable: boolean;  // ONLY reliable "is this real?" check
}
```

`metadataAvailable: false` is the only safe way to distinguish a genuine
first attempt from missing metadata — `info.attempt === 1` is true in both
cases.

### `getInstanceRetryHistory(client, instanceId)` — client-side

Projects per-activity retry history from `client.getStatus(instanceId,
{showHistory: true})`. Returns `undefined` if the instance does not exist.

```ts
interface InstanceRetryHistory {
    readonly instanceId: string;
    readonly attempts: ActivityRetryRecord[];
    readonly retryAttemptCount: number;          // attempts > 1
    readonly retryMaxAttemptsReached: boolean;
    readonly retryMetadataAvailable: boolean;
}
```

## Implementation notes

- **Two projection paths**:
  - *Path A* — raw DTFx history: walks `TaskScheduledEvent.Tags`, joins to
    `TaskCompleted`/`TaskFailed` via `EventId` ↔ `TaskScheduledId`.
  - *Path B* — the HTTP RPC status response shape: `TaskScheduled` events
    are folded away by the host post-processor, with `Tags` propagated onto
    the aggregated event. Path B walks those aggregated events directly
    when Path A finds nothing.
- Strict-decimal parser shared by both helpers — matches the producer-side
  `int.TryParse(s, NumberStyles.None, CultureInfo.InvariantCulture)`
  contract.
- `DurableOrchestrationStatus` constructor now reads
  `init.history ?? init.historyEvents` — the HTTP RPC status response uses
  `historyEvents`, programmatic status APIs use `history`.

## Files

- `src/retryVisibility.ts` (new) — both helpers + shared parser.
- `src/orchestrations/DurableOrchestrationStatus.ts` — accept either
  history key.
- `src/index.ts` — export the two new helpers.
- `types/retryVisibility.d.ts` (new) + `types/index.d.ts` — public typings.

## Compatibility

- Both helpers are net new — no API breaks.
- Activities without a retry policy see `metadataAvailable: false`.
- `DurableOrchestrationStatus` constructor change is backward-compatible
  (`history` key still wins when present).

### DIFF
--- a/src/retryVisibility.ts
+++ b/src/retryVisibility.ts
@@ -178,22 +178,33 @@ function makeUnavailable(): ActivityInvocationInfo {
 // getInstanceRetryHistory — client-side helper.
 // ---------------------------------------------------------------------------
 
-/** Minimal shape of getStatus payload we depend on. */
+/** Minimal shape of getStatus payload we depend on.
+ *  The extension's HTTP RPC layer returns history under the key `historyEvents`
+ *  (per host-side post-processing), while older or alternate transports may use
+ *  `history`. We accept either. */
 interface DurableClientLike {
     getStatus(
         instanceId: string,
         options?: { showHistory?: boolean; showInput?: boolean; showHistoryOutput?: boolean }
-    ): Promise<{ history?: Array<unknown> } | undefined>;
+    ): Promise<{ history?: Array<unknown>; historyEvents?: Array<unknown> } | undefined>;
 }
 
 /**
  * Project retry history for an orchestration instance.
  *
- * Calls `client.getStatus(instanceId, {showHistory: true})`, filters history
- * events to `TaskScheduled` entries carrying `dt.retry.*` tags, and joins each
- * to its matching `TaskCompleted` / `TaskFailed` (or marks `"running"` if
- * neither has appeared yet). Returns `undefined` when the instance does not
- * exist (matching the missing-instance contract of `getStatus`).
+ * Calls `client.getStatus(instanceId, {showHistory: true})` and walks the
+ * returned history to identify activity invocations carrying `dt.retry.*` tags.
+ *
+ * **Tag-location compatibility.** Tags are read from two places, in priority order:
+ *   1. `TaskScheduledEvent.Tags` — the canonical source of truth (raw DTFx history).
+ *   2. `TaskCompletedEvent.Tags` / `TaskFailedEvent.Tags` — set by the Functions
+ *      extension's `AddScheduledEventDataAndAggregate` post-processor, which
+ *      folds TaskScheduled events away and propagates their Tags onto the
+ *      aggregated Completed/Failed event. This is what `DurableClient.getStatus`
+ *      actually returns over the HTTP RPC endpoint.
+ *
+ * Returns `undefined` when the instance does not exist (matching the
+ * missing-instance contract of `getStatus`).
  *
  * **Complexity:** `O(history length)`. Downloads the full history. For
  * long-running instances with very large histories (>10k events), expect
@@ -216,11 +227,21 @@ export async function getInstanceRetryHistory(
         return undefined;
     }
 
-    const events = Array.isArray(status.history) ? status.history : [];
+    // Accept either casing: `history` (raw DTFx-style) or `historyEvents`
+    // (Functions extension HTTP RPC response).
+    const events: Array<unknown> = Array.isArray(status.history)
+        ? status.history
+        : Array.isArray(status.historyEvents)
+        ? status.historyEvents
+        : [];
 
     interface NormalizedEvent {
         readonly eventType: string;
-        readonly taskScheduledId?: number;
+        /** For TaskScheduled events: the event's own EventId (which is what
+         *  TaskCompleted/TaskFailed point back at via TaskScheduledId).
+         *  For TaskCompleted/TaskFailed events: the original TaskScheduledId.
+         *  Unified into one field so the join below is a single set lookup. */
+        readonly scheduledId?: number;
         readonly name?: string;
         readonly tags?: Record<string, string>;
     }
@@ -237,8 +258,21 @@ export async function getInstanceRetryHistory(
         if (typeof eventType !== "string") {
             return undefined;
         }
-        const taskScheduledIdRaw = o["TaskScheduledId"] ?? o["taskScheduledId"];
-        const name = (o["Name"] ?? o["name"]) as string | undefined;
+        // Read the per-event "join key" depending on event type. The join
+        // direction is: TaskCompleted/TaskFailed.TaskScheduledId === TaskScheduled.EventId.
+        // We collapse both into one normalized `scheduledId` field so the
+        // joiner below does not have to know about event-type specifics.
+        const idCandidate =
+            eventType === "TaskScheduled"
+                ? o["EventId"] ?? o["eventId"]
+                : o["TaskScheduledId"] ?? o["taskScheduledId"];
+        // The activity name lives under different keys depending on whether
+        // the response is raw DTFx history (`Name`/`name`) or post-processed
+        // by the extension's HTTP RPC layer (`FunctionName`/`functionName` —
+        // moved over by AddScheduledEventDataAndAggregate).
+        const name = (o["Name"] ?? o["name"] ?? o["FunctionName"] ?? o["functionName"]) as
+            | string
+            | undefined;
         const tagsRaw = o["Tags"] ?? o["tags"];
         const tags =
             tagsRaw && typeof tagsRaw === "object"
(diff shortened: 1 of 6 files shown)


src/retryVisibility.ts
```diff
@@ -0,0 +1,417 @@
...
+    /** 1-based attempt counter; defaults to 1 when metadata is unavailable. */
+    readonly attempt: number;
+    /** Policy ceiling (the `maxAttempts` value the customer passed to `RetryOptions`);
+     *  defaults to 1 when metadata is unavailable. */
+    readonly maxAttempts: number;
+    /**
+     * True when the retry counter has reached the policy's `maxAttempts`
+     * (i.e. `attempt === maxAttempts`). False when metadata is unavailable.
+     *
+     * ⚠️ This does NOT mean "no further attempts will run." `RetryOptions.handle`
```
nytian: oops maybe need to remove the emoji? Is this allowed in comments?

src/retryVisibility.ts
```diff
@@ -0,0 +1,417 @@
...
+ * `TaskScheduledEvent.Tags` through persistence drop retry metadata.
+ * Use `retryMetadataAvailable` to distinguish "no retries happened" from "we
+ * don't know if retries happened."
+ */
+export async function getInstanceRetryHistory(
+    client: DurableClientLike,
+    instanceId: string
+): Promise<InstanceRetryHistory | undefined> {
+    const status = await client.getStatus(instanceId, { showHistory: true });
+    if (!status) {
```
nytian, line 255: if status/instanceId doesn't exist, the client will throw error. So I think here we should wrap it instead of check if status is null? Or add a catch block?

types/retryVisibility.d.ts
```diff
@@ -0,0 +1,104 @@
...
+    /** 1-based attempt counter; defaults to 1 when metadata is unavailable. */
+    readonly attempt: number;
+    /** Policy ceiling (the `maxAttempts` value passed to `RetryOptions`);
+     *  defaults to 1 when metadata is unavailable. */
+    readonly maxAttempts: number;
+    /**
+     * True when the retry counter has reached the policy's `maxAttempts`
+     * (i.e. `attempt === maxAttempts`). False when metadata is unavailable.
+     *
+     * ⚠️ This does NOT mean "no further attempts will run." `RetryOptions.handle`
```
nytian: same here for the emoji

src/retryVisibility.ts
```diff
@@ -0,0 +1,417 @@
...
+        return undefined;
+    }
+
+    // Accept either casing: `history` (raw DTFx-style) or `historyEvents`
+    // (HTTP RPC status-response shape).
+    const events: Array<unknown> = Array.isArray(status.history)
+        ? status.history
+        : Array.isArray(status.historyEvents)
```
nytian: I think status.historyEvents is always undefined here as the DurableOrchestrationStatus constructor already normalizes history ?? historyEvents into .history and doesn't keep historyEvents as a field.

nytian (commented):
Left some questions. can we add some integration test for this PR?

src/retryVisibility.ts
```diff
@@ -0,0 +1,447 @@
+/**
+ * Native activity retry visibility — helpers for inspecting per-attempt
+ * retry metadata recorded on activity history events.
+ *
+ * Two surfaces are exported from this module:
+ *
+ *  - `getActivityInvocationInfo(context)` — read from inside an activity to learn
```
sophiatev, line 7: what are the "goals" referenced here?

src/retryVisibility.ts
```diff
@@ -0,0 +1,447 @@
...
+    /** 1-based attempt counter; defaults to 1 when metadata is unavailable. */
+    readonly attempt: number;
+    /** Policy ceiling (the `maxAttempts` value the customer passed to `RetryOptions`);
+     *  defaults to 1 when metadata is unavailable. */
+    readonly maxAttempts: number;
+    /**
+     * True when the retry counter has reached the policy's `maxAttempts`
+     * (i.e. `attempt === maxAttempts`). False when metadata is unavailable.
+     *
+     * NOTE: This does NOT mean "no further attempts will run." `RetryOptions.handle`
```
sophiatev: I'm confused here - even if we abort this attempt and do another retry, wouldn't the "attempt" counter stay the same and still equal "maxAttempts" if it did previously?

src/retryVisibility.ts
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    const attempt = parsePositiveInt(meta[TRIGGER_KEY_ATTEMPT]);
+    const maxAttempts = parsePositiveInt(meta[TRIGGER_KEY_MAX_ATTEMPTS]);
+
+    if (
+        attempt === undefined ||
+        maxAttempts === undefined ||
+        attempt < 1 ||
+        maxAttempts < attempt
```
sophiatev, line 156: Can we add a check here for maxAttempts < 1?

src/retryVisibility.ts
```diff
@@ -0,0 +1,447 @@
...
+        /** For TaskScheduled events: the event's own EventId (which is what
+         *  TaskCompleted/TaskFailed point back at via TaskScheduledId).
+         *  For TaskCompleted/TaskFailed events: the original TaskScheduledId.
+         *  Unified into one field so the join below is a single set lookup. */
+        readonly scheduledId?: number;
+        readonly name?: string;
+        readonly tags?: Record<string, string>;
+    }
+
+    // Property casing on getStatus payload — Spike #2 in the design called for
```
sophiatev: What is spike #2?

src/retryVisibility.ts
```diff
@@ -0,0 +1,447 @@
...
+        const tags = e.tags;
+        if (!tags) continue;
+
+        const attempt = parsePositiveInt(tags[HISTORY_TAG_ATTEMPT]);
+        const maxAttempts = parsePositiveInt(tags[HISTORY_TAG_MAX_ATTEMPTS]);
+        if (
+            attempt === undefined ||
+            maxAttempts === undefined ||
+            attempt < 1 ||
+            maxAttempts < attempt
```
sophiatev, line 353: Can we add `maxAttempts < 1` too?

src/retryVisibility.ts
```diff
@@ -0,0 +1,447 @@
...
+            const tags = e.tags;
+            if (!tags) continue;
+
+            const attempt = parsePositiveInt(tags[HISTORY_TAG_ATTEMPT]);
+            const maxAttempts = parsePositiveInt(tags[HISTORY_TAG_MAX_ATTEMPTS]);
+            if (
+                attempt === undefined ||
+                maxAttempts === undefined ||
+                attempt < 1 ||
+                maxAttempts < attempt
```
sophiatev, line 399: Same comment about maxAttempts

types/retryVisibility.d.ts
```diff
@@ -0,0 +1,104 @@
...
+     * `TaskScheduledEvent.Tags` through persistence.
+     */
+    readonly retryMetadataAvailable: boolean;
+}
+
+/**
+ * Read the current retry attempt information for an activity invocation.
+ *
+ * When the activity was scheduled with a `RetryOptions` policy AND the backend
+ * preserves `TaskScheduledEvent.Tags` (currently DTS only), returns the parsed
```
sophiatev: Can we remove the reference to DTS here

types/retryVisibility.d.ts
```diff
@@ -0,0 +1,104 @@
...
+ * attempt counter, policy ceiling, and an `isMaxAttempt` flag.
+ *
+ * When metadata is unavailable (no retry policy, non-DTS backend, parsing
```
sophiatev: and here

src/retryVisibility.ts
bachuv: Removed from comment

src/retryVisibility.ts
bachuv: Updated the note to be clearer. 

`isMaxAttempt` is really just `attempt === maxAttempts`, where `maxAttempts` is the configured max attempts (RetryOptions.maxNumberOfAttempts)

src/retryVisibility.ts
bachuv: That shouldn't be there. I removed it and updated other comments as well.

PR #713, opened 2026-06-29
Switch release pipeline to use ESRP

andystaples:
Switches the release pipeline to use ESRP release pipelines

### DIFF
--- a/azure-pipelines/release.yml
+++ b/azure-pipelines/release.yml
@@ -56,5 +56,8 @@ extends:
             owners: $(Build.RequestedForEmail)
             approvers: $(Build.RequestedForEmail)
             serviceendpointurl: "https://api.esrp.microsoft.com"
+            # durabletask-java is the publisher name we selected when we
+            # first set up ESRP for durable packages (we did Java first),
+            # but is also correct for all other languages.
             mainpublisher: "durabletask-java"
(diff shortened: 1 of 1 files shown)


azure-pipelines/release.yml
```diff
@@ -1,61 +1,60 @@
...
+            folderlocation: "$(System.DefaultWorkingDirectory)/drop"
+            waitforreleasecompletion: true
+            # Auto-populate from the build queuer's identity so we don't
+            # hardcode personal emails in source. ESRP will send the
+            # release notification / approval link to whoever clicked
+            # "Run pipeline".
+            owners: $(Build.RequestedForEmail)
+            approvers: $(Build.RequestedForEmail)
+            serviceendpointurl: "https://api.esrp.microsoft.com"
+            mainpublisher: "durabletask-java"
```
bachuv, line 62: Why does this say `durabletask-java`?

andystaples: `durabletask-java` is just because when we set up the ESRP connection, we did Java first and that was the name we picked. 
It's the same in `durabletask-js` https://github.com/microsoft/durabletask-js/blob/main/eng/ci/release.yml
and `durabletask-python` https://github.com/microsoft/durabletask-python/blob/main/eng/ci/release.yml

bachuv (approved):

PR #721, opened 2026-07-31
Route npm dependency restores through CFS

fabiocav:
## Summary

- route repository-owned npm restores through `azfunc/public/upstream-public`, including explicit mappings for every scoped production dependency
- authenticate both root and detached smoke-test pipeline contexts before npm runs, and quote interpolated paths
- preserve customer-facing samples and lockfile URLs while keeping `.npmrc` out of npm packs and Functions deployment content
- make the documented global Azurite install preserve `NPM_CONFIG_USERCONFIG` through `sudo env`

The repository audit found npm manifests only. There are no repository NuGet or Python manifests/configuration to onboard; customer-created `extensions.csproj`/`func extensions install` guidance and sample npm trees remain unchanged.

## Validation

- restored the root graph from an empty npm cache with ephemeral Azure Artifacts authentication; verbose requests used `pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public` and Azure Artifacts blob hosts, with no direct npmjs.org, nuget.org, or PyPI traffic
- ran `npm run test`, `npm run test:nolint`, and `npm run build` (307 tests passing in each test run; existing non-null-assertion lint warning remains)
- ran production prune, `npm pack --dry-run --json`, `npm publish --dry-run --json`, and a real pack; the 206-entry archive contained no `.npmrc`, credentials, lockfiles, or `node_modules`
- installed the packed SDK and restored/built the detached test app from a Windows path containing spaces using argument-array invocation and an empty cache; no direct public-registry traffic
- installed global `azurite@2.7.1` from an empty cache through CFS with an explicit user config
- exercised `npm audit` through CFS; it reports the existing 8 dependency findings (1 low, 3 moderate, 4 high) without suppression

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ Note: [Azurite V3](https://github.com/Azure/Azurite) does not have support for T
 
 Create a folder say AzureWebJobsStorage
 
-`npm install azurite@2.7.1 -g`
+`sudo env NPM_CONFIG_USERCONFIG="$(git rev-parse --show-toplevel)/.npmrc" npm install azurite@2.7.1 -g`
 
 `azurite -l ./AzureWebJobsStorage`
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ Note: [Azurite V3](https://github.com/Azure/Azurite) does not have support for T
 
 Create a folder say AzureWebJobsStorage
 
-`sudo env NPM_CONFIG_USERCONFIG="$(git rev-parse --show-toplevel)/.npmrc" npm install azurite@2.7.1 -g`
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -40,7 +40,7 @@ Note: [Azurite V3](https://github.com/Azure/Azurite) does not have support for T
 
 Create a folder say AzureWebJobsStorage
 
-`npm install azurite@2.7.1 -g`
+`sudo env NPM_CONFIG_USERCONFIG="$(git rev-parse --show-toplevel)/.npmrc" npm install azurite@2.7.1 -g`
```
jviau: I think this can be removed? I don't think the readme needs to be opinionated about how the user gets `azurite`.

halspang (approved):

PR #727, opened 2026-09-01
Fix Failure Message Handling for MSSQL

nytian:
When structured failure details are empty, the SDK should use the legacy error fields for the MSSQL backend. MSSQL returns FailureDetails: {}, which was incorrectly treated as valid, causing the expected message "InvalidOperationException: This activity failed" to become "Error: ". This fix preserves the expected error message.

### DIFF
--- a/src/orchestrations/TaskOrchestrationExecutor.ts
+++ b/src/orchestrations/TaskOrchestrationExecutor.ts
@@ -37,7 +37,10 @@ function extractFailureDetailsPayload(event: HistoryEvent): FailureDetailsPayloa
     }
     const value = (event as { FailureDetails?: unknown }).FailureDetails;
     if (value && typeof value === "object") {
-        return value as FailureDetailsPayload;
+        const failureDetails = value as FailureDetailsPayload;
+        if (typeof failureDetails.ErrorType === "string" && failureDetails.ErrorType.length > 0) {
+            return failureDetails;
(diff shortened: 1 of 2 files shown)


src/orchestrations/TaskOrchestrationExecutor.ts
```diff
@@ -37,7 +37,10 @@ function extractFailureDetailsPayload(event: HistoryEvent): FailureDetailsPayloa
     }
     const value = (event as { FailureDetails?: unknown }).FailureDetails;
     if (value && typeof value === "object") {
-        return value as FailureDetailsPayload;
+        const failureDetails = value as FailureDetailsPayload;
```
halspang, line 40: Sorry, not super familiar with JS typing rules. We check that `value` is an `object` and then appear to try and cast it into the `FailureDetailsPayload`. Should we check that type instead? If it's a different type, will either of the `value as` throw?

nytian: We can't check it as `FailureDetailsPayload` directly. `FailureDetailsPayload` is a TypeScript interface and only exists at js complier but not at runtime. So we use `as` here since it will only impact the complier and thus won't throw any error. And we validate that the value is an object with a non-empty string `ErrorType`.

halspang (approved):

PR #728, opened 2026-09-01
Add DurableClient.restart with explicit instance ID and version options

bachuv:
## Summary

Adds `DurableClient.restart()` to the classic Durable Functions JavaScript SDK.

The API supports the existing extension Restart behavior:

```typescript
await client.restart("source-instance");
await client.restart("source-instance", false);

It also supports selecting an exact target instance ID and orchestration version:

const instanceId = await client.restart("source-instance", {
    newInstanceId: "target-instance",
    version: "2.0",
});
```

### Related PR
- https://github.com/Azure/azure-functions-durable-extension/pull/3539

### Changes

• Added the  DurableClient.restart(instanceId, restartWithNewInstanceId?)  overload.
• Added the  DurableClient.restart(instanceId, RestartOptions)  overload.
• Added the public  RestartOptions  type with  newInstanceId  and optional  version .
• Uses the existing  /restart  HTTP operation for boolean calls.
• Uses the extension's dedicated  /restartWithOptions  operation for exact instance ID and version requests.
• Supports both the fast local HTTP RPC endpoint and the external management endpoint.
• Preserves task-hub, storage connection, authorization, and other management query parameters.
• URL-encodes source instance IDs as a single path segment.
• Validates target IDs and successful response payloads.
• Added  restartPostUri  to  HttpManagementPayload  as an optional property for compatibility with older extension payloads.
• Added API documentation and unit coverage.

### Compatibility

• Existing JavaScript APIs are unchanged.
• The boolean overload maps to the existing Restart endpoint.
• The options overload requires an extension version containing the paired  restartWithOptions  endpoint.
• Older extensions remain compatible with applications that do not call  restart() .
• No protobuf, dependency, package-lock, or build-system changes are included.

### Testing

• Ran the repository's complete  npm test workflow.
• Build, source lint, test lint, TypeScript compilation, declaration generation, and Mocha tests passed.
• Result: 314 tests passed and 2 tests remained pending.

### AI-assisted code disclosure

• AI tool: GitHub Copilot CLI
• AI-assisted areas: restart API implementation, public declarations, management payload compatibility, documentation, and unit tests
• Review performed for HTTP routing, task-hub and connection propagation, URL encoding, malformed responses, and backward compatibility

### DIFF
--- a/src/durableClient/DurableClient.ts
+++ b/src/durableClient/DurableClient.ts
@@ -10,6 +10,7 @@ import url = require("url");
 import { isURL } from "validator";
 import {
     StartNewOptions,
+    RestartOptions,
     GetStatusOptions,
     OrchestrationFilter,
     TaskHubOptions,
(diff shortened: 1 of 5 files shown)


src/durableClient/DurableClient.ts
```diff
@@ -527,6 +528,96 @@ export class DurableClient implements types.DurableClient {
...
+            const restartedInstanceId = (response.data as HttpManagementPayload).id;
+            if (typeof restartedInstanceId !== "string" || !restartedInstanceId) {
+                throw new Error("The restart operation returned an invalid response.");
+            }
+            if (
+                useOptions &&
+                restartedInstanceId !== restartWithNewInstanceIdOrOptions.newInstanceId
+            ) {
+                throw new Error(
+                    `The restart operation returned instance ID '${restartedInstanceId}' instead of the requested ID '${restartWithNewInstanceIdOrOptions.newInstanceId}'.`
```
YunchuWang: **[P2] Snapshot the requested target ID before awaiting the HTTP response**

The request captures `options.newInstanceId` in the URL before `await`, but this response check reads it again from the caller's mutable options object afterward. If a caller reuses that object for concurrent restarts, changing the target from `"first"` to `"second"` between calls, the first request correctly creates `"first"` but then rejects because it compares the response against `"second"`. This reports a successful orchestration creation as a failure and can trigger unnecessary retries.

Please capture the requested ID in a local constant before sending the request, and use that same value for URL construction, response validation, and the error message. A regression test that mutates the options object while the HTTP response is pending would cover this case.

YunchuWang (approved):
The target instance ID is now captured before the HTTP request and the same snapshot is used for response validation and error reporting, resolving the previous finding. No remaining blocking issues found in the reviewed changes.

PR #735, opened 2026-09-22
Expose source instance ID in cloned orchestration contexts

bachuv:
## Summary

Exposes the immediate source orchestration instance ID to cloned orchestrations through `context.df.sourceInstanceId`.

The Durable Functions extension persists the source reference when restart creates a different logical orchestration instance and includes it in the orchestration-trigger payload. This change reads that payload value and makes it available to JavaScript orchestration code.

## Changes

- Add `sourceInstanceId` to the orchestration-trigger binding model.
- Add the optional `sourceInstanceId` property to `DurableOrchestrationContext`.
- Forward the binding value when constructing `context.df`.
- Update the public TypeScript declarations.
- Update orchestration testing utilities to support source-instance metadata.
- Add integration coverage verifying that the source instance ID is assigned to `context.df`.

## Behavior

```javascript
await client.restart("source-instance", {
    newInstanceId: "clone-instance",
    version: "2.0",
});
```

The cloned orchestration can access its immediate source:

```javascript
df.app.orchestration("workflow", function* (context) {
    context.df.instanceId;       // "clone-instance"
    context.df.sourceInstanceId; // "source-instance"
});
```

`sourceInstanceId` is `undefined` when the extension does not provide source metadata, including for ordinary orchestration starts and older extension versions.

The source reference is not added to orchestration `input`, `parentInstanceId`, or `DurableOrchestrationStatus`.

Related PRs

• Extension support: [Azure/azure-functions-durable-extension](https://github.com/Azure/azure-functions-durable-extension/pull/3555)
• Builds on restart-with-options support from Azure/azure-functions-durable-js#728.

Testing

•  npm test 
• Result: 317 tests passed.


src/orchestrations/DurableOrchestrationContext.ts
```diff
@@ -58,13 +58,15 @@ export class DurableOrchestrationContext implements types.DurableOrchestrationCo
...
-        private taskOrchestratorExecutor: TaskOrchestrationExecutor
+        private taskOrchestratorExecutor: TaskOrchestrationExecutor,
+        sourceInstanceId?: string
     ) {
         this.state = state;
         this.instanceId = instanceId;
         this.isReplaying = isReplaying;
         this.currentUtcDateTime = currentUtcDateTime;
         this.parentInstanceId = parentInstanceId;
+        this.sourceInstanceId = sourceInstanceId;
```
YunchuWang: **[P2] Normalize an explicit null source ID to undefined**

The companion extension [always writes the `sourceInstanceId` JSON property](https://github.com/Azure/azure-functions-durable-extension/blob/91971d5456bf5248f074514acbe416a1780630b1/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs#L253-L264), including `null` for ordinary starts and sub-orchestrations. The binding parser and `Orchestrator` forward it unchanged, so this assignment exposes `null` despite the public `string | undefined` declaration and the PR's promised `undefined` behavior. A consumer using `sourceInstanceId !== undefined` can therefore treat an ordinary orchestration as a clone and dereference null.

Please normalize the wire value to `undefined` at the boundary and represent the nullable wire shape accurately. Cover both an explicitly null field from the new extension and an omitted field from older extensions, in addition to the existing non-empty clone case.

YunchuWang (changes requested):
Requesting changes for the null-versus-undefined mismatch at the extension/JavaScript boundary. Reviewed commit `48f29df333f0a3a6f985108b0b47b4d46eab86c1` together with the companion extension PR #3555.

